Issue #436, opened 2015-01-27
Installing CoolProp 5.0.6 on Linux


JAExposito:
Hello,

I had previously used CoolProp 4.2.0 and I installed it on Linux running the make file in the shared library folder of the source code. Now, in the version 5.06 there is no shared library folder within the wrappers folder so I would like to know how should I install it?

Thank you in advance and kind regards.

ibell:
Have you seen this: http://www.coolprop.org/coolprop/wrappers/SharedLibrary/index.html#shared-library ?

JAExposito:
Hello,

Yes I saw it but I was not sure if this would work like the previous version I used. Will this execute the cmakelist file? I guess that the command "cmake .. -DCOOLPROP_32BIT_STDCALL_SHARED_LIBRARY=ON" will execute the installation as it used to be with the make file y version 4.2.0 and then CoolProp can be used like in the example in C++ code. Is that correct? 
For installing the new versión 5.0.6 should I remove the previous one before?

Kind regards.

jowr:
It will build the library, not install it. We do not include install scripts in the new version, you are very welcome to contribute an updated makefile that can be used to install the shared library on Linux and MacOS.

jowr:
You can also check the cmake docs, something like `cmake -DCMAKE_INSTALL_PREFIX:PATH=/usr/local -DCOOLPROP_32BIT_STDCALL_SHARED_LIBRARY=ON .. && make all install` might work as well, but I am not sure. ... and yes, you should remove the old version before installing a new one.

jowr:
I just tried it myself and something is wrong here: 
```
In file included from /home/jowr/Documents/Fluids/CoolProp.GitHub/src/CoolPropLib.cpp:9:0:
/home/jowr/Documents/Fluids/CoolProp.GitHub/include/CoolPropLib.h:70:35: error: expected initializer before ‘Props1SI’
     EXPORT_CODE double CONVENTION Props1SI(const char *FluidName, const char* Output);
```
It looks like the calling conventions are somehow messed up fir 32bit Linux.

jowr:
For me, `cmake ../CoolProp.GitHub -DCOOLPROP_32BIT_SHARED_LIBRARY_LINUX_MODULE=ON` could compile the code properly. To install, you have to run
```
sudo cp libCoolProp.so /usr/local/lib/libCoolProp.so.32.5.0.7dev
pushd /usr/local/lib
[...]

PR #440, opened 2015-01-29
Implemented Lua C API and LuaJIT Wrappers for CoolProp

bungle:
Okay, here is the first version where I implemented both Lua C API, and LuaJIT FFI wrappers. The `Makefile` is rudimentary, and the `README.rst` is currently just a placeholder. I have tested these on my Mac OSX 10.9.5, but I have not yet checked other platforms. Please give a feedback, and suggestions on these to get merged with the upstream. I had never earlier done any Lua C API bindings (only LuaJIT FFIs), so it would be good if someone more experienced with Lua C API could look at these.


jowr:
I install LuaJIT on my Debian system and test it here. Hang on.

bungle:
Check also that PUC-Lua wrapper works too on your Debian system.

jowr:
I guess you have to help a little bit, here is the first error:
```
cc -std=c99 -pedantic-errors -fpic -I/usr/include/luajit-2.0    -g -O2 -Wall -Wextra -Wswitch-enum -Wwrite-strings -Wshadow -c -o coolprop/capi.o coolprop/capi.c 
coolprop/capi.c: In function ‘luaopen_coolprop_capi’: 
coolprop/capi.c:166:3: error: implicit declaration of function ‘luaL_newlib’ [-Wimplicit-function-declaration] 
    luaL_newlib(L,funcs);
    ^ 
lualib.mk:71: recipe for target 'coolprop/capi.o' failed 
make: *** [coolprop/capi.o] Error 1 
```

jowr:
I tried adding the  `-Wimplicit-function-declaration` to the compiler options, but that did not help.

bungle:
The capi.c needs only be build when using PUC Lua, you can go ahead and just run luajit examples.lua. It will use FFI bindings then. There may be differences in header files, for example there is this kind of check in lua-snappy: https://github.com/forhappy/lua-snappy/blob/master/lua-snappy.cc#L91-L95 (so i think luaL_newlib might be missing from LuaJIT (and you have to use luaL_register instead).

jowr:
Sorry, but #436 has to be solved first. I would like to update my installed library...

craigbarnes:
> Please give a feedback, and suggestions

I'd appreciate it if you updated `lualib.mk` with [this](https://github.com/craigbarnes/lua-gumbo/commit/cfae91494bd68c4ad7de06b9e1190ae588565b00). :wink:

craigbarnes:
You may also find [this](https://github.com/craigbarnes/lua-sass/blob/0bec996d4974d9509f831177d321853eb2ec5aec/compat.h#L9-11) useful for the `luaL_newlib` issue.

bungle:
Thanks @craigbarnes, I added the license information (thanks for that great config), and also added detection of luaL_newlib.

jowr:
Compilation and `luajit example.lua` seems to work. How can I test the FFI version?

jowr:
... or the C version if the example uses the FFI.

bungle:
It will automatically select FFI version of the binding if you run it with LuaJIT (and have not turned JIT off, it's on by default). If you are not using LuaJIT it will automatically select Lua C API version (the capi.c wrapper). LuaJIT doesn't need any compiled wrapper, only that the library is in lib search path. Lua modules are cached, so this check is only done the first time the module is `require`d.

bungle:
I will add more documentation to `README.rst`, hopefully I have some time for it tomorrow.

jowr:
OK. Just let me know if you need someone to do more testing.

bungle:
I will shape this a little bit more (maybe add it to LuaRocks as well). I will let you know when I feel that this is ready to be merged. At its current shape it doesn't look too bad when compared to others in the wrappers folder. So I think we are quite close to integrate it (I had some problems with `get_mixture_binary_pair_data` (symbol not found) so I had to comment it on capi.c version). Of course more testing, and with different platforms is always something that it really valuable. Thanks @jowr for testing this on your Debian system.

ibell:
How about windows?  I guess the makefile might not work, but otherwise
should it be able to be compiled on windows?

bungle:
Not that the bug #441 is fixed I think that I have finished implementing this. I have also added documentation (see: https://github.com/bungle/CoolProp/tree/master/wrappers/Lua). So I consider this as good-to-be-merged with upstream.

I'm not sure if this needs to be integrated with Cmake build process, but I think that you can figure it out?

Regards
Aapo

ibell:
Yes, very nice work!  Out of curiosity, what are you planning on using
CoolProp for in Lua?  I'm aware of Lua, but I've not heard of anyone
wanting to combine Lua and CoolProp

bungle:
@ibell, well it all started from this message to lua-l (a mailing list):
http://lua-users.org/lists/lua-l/2015-01/msg00695.html

So, I'm not the end user. Although sometimes I wish that I could do some scientific work as well, but I'm a programmer, not a scientist. But CoolLib is really nice library. I learned this and that about thermodynamics when implemented it, and now I know where to look if I ever get to the point where I would need to solve problem related to thermodynamics.

I found your high-level C API and thought that it would be simple to do. So I started with LuaJIT and implemented FFI interface to it in couple of hours. I informed Renaud of that but he said that he's using Lua 5.1. I had never done any Lua C API bindings so I thought this could be a good excercise because the high-level API of CoolProp is dead-simple. I also want to learn more about C and its toolset (make, cmake, gcc etc.).

Why I did it? Because I could, and I wanted to leadn. That's all. Great library indeed.

I think Lua is somewhat a hidden gem. More and more people seem to find it though. Google and Facebook are using https://github.com/torch/torch7 and there is also http://www.scilua.org/. Also the Lua was originally developed for the people working on scientific and mathematical topics, more about it here: http://www.lua.org/history.html. So I think that Lua is a perfect match for people working with Thermodynamics. My personal interest in Lua focuses to OpenResty (http://openresty.org/ - Nginx based web development platform / framework).

jowr:
Great, let's hope that someone comes along to test the C implementation.

wrappers/Lua/Makefile
```diff
@@ -0,0 +1,36 @@
+include lualib.mk
+
+COOLPROP_CFLAGS  ?= $(shell $(PKGCONFIG) --cflags CoolProp)
+COOLPROP_LDFLAGS ?= $(shell $(PKGCONFIG) --libs-only-L CoolProp)
+COOLPROP_LDLIBS  ?= $(or $(shell $(PKGCONFIG) --libs-only-l CoolProp), -lCoolProp)
+COOLPROP_INCDIR  ?= $(shell $(PKGCONFIG) --variable=includedir CoolProp)
+COOLPROP_LIBDIR  ?= $(shell $(PKGCONFIG) --variable=libdir CoolProp)
+COOLPROP_HEADER  ?= $(or $(COOLPROP_INCDIR)
```
craigbarnes: This line seems wrong in at least 3 ways.

ibell: Pull request?  Snark helps no one.

craigbarnes: Yes, point taken. I edited my comment shortly after posting it. The line in question isn't actually used, so the fix would just be to remove it.

bungle:
I added a pull request to remove unneeded lines from makefile:
https://github.com/CoolProp/CoolProp/pull/445

PR #474, opened 2015-02-18
Try to fixes #473

### LINKED ISSUE #473: Wrapper for Julia 3
The current Julia wrapper is only compatible for Julia 0.4 (development version), and not compatible with Julia 0.3.
The only difference is that they changed the way to write the unsigned integer.
The old way was Uint and the new way is UInt. So while the new way should be kept for Julia 4, it would be interesting to have a version of the wrapper compatible with the current Julia release.

I tested this on Ubuntu 14.10 64bit and Julia 0.3.5 and the PropsSI function worked fine:

    module CoolProp
    
    export F2K, HAPropsSI, PropsSI, PhaseSI
    
    function F2K(TF::Float64)
      return ccall( (:F2K, "CoolProp"), Cdouble, (Cdouble,), TF)
    end
    
    function HAPropsSI(Output::String, Name1::String, Value1::Float64, Name2::String, Value2::Float64, Name3::String, Value3::Float64)
      return ccall( (:HAPropsSI, "CoolProp"), Cdouble, (Ptr{Uint8},Ptr{Uint8},Float64,Ptr{Uint8},Float64,Ptr{Uint8},Float64), Output,Name1,Value1,Name2,Value2,Name3,Value3)
    end
 
    function PropsSI(Output::String, Name1::String, Value1::Float64, Name2::String, Value2::Float64, Fluid::String)
      return ccall( (:PropsSI, "CoolProp"), Cdouble, (Ptr{Uint8},Ptr{Uint8},Float64,Ptr{Uint8},Float64,Ptr{Uint8}), Output,Name1,Value1,Name2,Value2,Fluid)
    end
 
    function PhaseSI(Name1::String, Value1::Float64, Name2::String, Value2::Float64, Fluid::String)
      outstring = Array(Uint8, 255)
      val = ccall( (:PhaseSI, "CoolProp"), Int32, (Ptr{Uint8},Float64,Ptr{Uint8},Float64,Ptr{Uint8}, Ptr{Uint8}, Int), Name1,Value1,Name2,Value2,Fluid,outstring,length(outstring))
      return bytestring(convert(Ptr{Uint8}, outstring))
    end
 
    end #module

(I only changed the UInt8 by Uint8)

JonWel:
A way to have two different wrappers for Julia Lang depending on the version (3 or 4).
If you prefer an other way (two files in a same Julia folder), just let me know and I'll try to do it.
I updated the webpage doc. I'm not sure it is correctly done nor it is the only doc to be updated.


JonWel:
Sorry, I badly made the pull request (I wanted to add only 4 commit and it send all the diff between master and release with it. I now know I have to carefully check that).

jowr:
Hi, thansk for your help, much appreciated. Would you mind merging to `master` instead of `release`? Should we close this one and you make a new pull request?

JonWel:
Yes, close this one (I would have close it myself if I could).
This is the first time I use GitHub, I realised a bit too late that I should merge to the master.

Issue #483, opened 2015-02-21
Problems noted by VS2013 static analysis


ibell:
Thanks to @mikekaganski for noticing this:

```
C6011    Dereferencing null pointer    Dereferencing NULL pointer 'fractions_ptr'.     CoolProp    coolprop.cpp    234
        'fractions_ptr' is NULL            193
        Skip this branch, (assume 'fluid_names.size()>1' is false)            195
        Skip this branch, (assume 'fluid_names.size()==1' is false)            201
        Skip this branch, (assume '<branch condition>' is false)            225
        Skip this branch, (assume '<branch condition>' is false)            231
        Enter this branch, (assume '<branch condition>')         233
        'fractions_ptr' is an Output from 'CoolProp::AbstractState::set_volu_fractions' (declared at d:\documents\coolprop\coolprop\include\abstractstate.h:363)     234
        'fractions_ptr' is dereferenced, but may still be NULL         234
C6031    Return value ignored    Return value ignored: 'freopen'. CoolProp    coolproplib.cpp    105
C6283    Primitive array-new scalar-delete mismatch    'buffer' is allocated with array new [], but deleted with scalar delete. CoolProp    coolproptools.cpp    33
C6001    Using uninitialized memory    Using uninitialized memory 'y'.    CoolProp    phaseenveloperoutines.cpp    448
        'y' is not initialized            435
        Assume switch ( 'iInput2' ) resolves to the default case         440
        'y' is an Output from 'std::vector<long double,std::allocator<long double> >::[]' (declared at c:\program files (x86)\microsoft visual studio 12.0\vc\include\vector:1196)            448
        'y' is used, but may not have been initialized 448
```

ibell:
I think I have fixed all of these, please confirm.  The freopen was not a real problem, they others were.

mikekaganski:
> please confirm

Excuse me, forgot to confirm.
Yes, no more abovementioned warnings from analyzer.

PR #488, opened 2015-02-22
Some more static analyser warning fixes

mikekaganski:
Here are some fixes:
1. restructuring the fix in CoolProp.cpp for #483;
2. return result of redirect_stdout DLL function (allowing for error handling);
3. fixes for "Local declaration hides local" warnings - there seem to be a number of unnecessary variables, and even one variable declaration that masks return value (iclosest in PhaseEnvelopeRoutines::is_inside())
4. restructuring get_fluid_param_string() exception handling to simplify and unify it; removes two identical code runs.


ibell:
Nice, I'll merge them, but I ran static analyzer in VS2013 with defaults
and it didn't complain about any of these.

By the way, static analysis with VS2013 takes a really long time, it was
several minutes for me.

mikekaganski:
> but I ran static analyzer in VS2013 with defaults and it didn't complain about any of these
Yes, I changes analyzer settings to all rules (ANALYZE->Configure Code Analysis for Solution)

ibell:
Thanks again for your work with this!

PR #489, opened 2015-02-22
Reduce copying of parameters

mikekaganski:
While I started to trace the execution of CoolProp library (to better understand its structure), I see excessive copying of parameters, as the result of (1) using non-ref parameters, and (2) using incompatible vectors (double vs. long double).

I started to reduce this. But I am worried that I had to change the PropsSImulti API to use vector<long double> instead of vector<double>. This is incompatible change, and possibly I need to undo this one, and make it copy parameter trading performance for compatibility?

Please review if possible. Thanks.

### DIFF
--- a/include/AbstractState.h
+++ b/include/AbstractState.h
@@ -239,14 +239,14 @@ protected:
     virtual long double calc_T_reducing(void){throw NotImplementedError("calc_T_reducing is not implemented for this backend");};
     /// Using this backend, get the critical point pressure in Pa
     virtual long double calc_p_critical(void){throw NotImplementedError("calc_p_critical is not implemented for this backend");};
+    /// Using this backend, get the reducing point pressure in Pa
+    virtual long double calc_p_reducing(void){throw NotImplementedError("calc_p_reducing is not implemented for this backend");};
     /// Using this backend, get the critical point molar density in mol/m^3
     virtual long double calc_rhomolar_critical(void){throw NotImplementedError("calc_rhomolar_critical is not implemented for this backend");};
 	/// Using this backend, get the reducing point molar density in mol/m^3
     virtual long double calc_rhomolar_reducing(void){throw NotImplementedError("calc_rhomolar_reducing is not implemented for this backend");};
-
     /// Using this backend, construct the phase envelope, the variable type describes the type of phase envelope to be built.
     virtual void calc_phase_envelope(const std::string &type){throw NotImplementedError("calc_phase_envelope is not implemented for this backend");};
-    
     /// 
     virtual long double calc_rhomass(void){return _rhomolar*molar_mass();}
     virtual long double calc_hmass(void){return hmolar()/molar_mass();}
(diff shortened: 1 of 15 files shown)


mikekaganski:
Or maybe I just need to globally change **long double** to **double**? I see no reason to use that **long double** thing. Why is it used in the first place?

jowr:
I think @ibell should answer that one in detail. All I remember is that we had some trouble with the solvers for the mixtures, which were related to truncation errors that kept the software from converging.

mikekaganski:
> we had some trouble with the solvers for the mixtures, which were related to truncation errors that kept the software from converging

Oh, that must have been fixed somehow differently. Because long double is equal to double in MS VC++, and it is the supported compiler for CoolProp.
Besides, this would imply that the results of floating-point math (just about 100% of CoolProp) would be inconsistent in different compilers (all of them treat double as 64-bit, while **some** treat long double as 80-bit, some as 128-bit, and some as 64-bit). So in some of them, the results would be obtained, while with others, they would not. Then the portability would be questionable...

ibell:
@mikekaganski - the reason we use long doubles in places is to get additional precision on platforms that support it.  Microsoft Visual C++ is, as usual, the only compiler that does not give you additional precision when using long doubles (long double maps to a double).  I haven't comprehensively tested the use of long doubles versus just doubles, but my belief is that the additional precision should help in some of the solvers, especially at low pressure and near the critical point.  If you can prove me wrong, I would be happy to move everything back to doubles.

mikekaganski:
:) Well, then maybe use it universally? At least internally. It would help avoid excessive copying.

And then, I would vote for a typedef to use instead of "long double" inside, like that:
**typedef longdouble long double**
(or an equivalent macro)

ibell:
Yes I have thought about something similar(*typedef longdouble long double)
.  *I think it is a good idea, and you are welcome to go ahead and
implement it.

The reason to not universally use long double is that it can't be used in
shared library interface functions, as well as some wrappers (SWIG)
struggle with it sometimes.  The SWIG limitation might not be true anymore,
but at the very least it would be good to see if we implement the macro you
recommend and switch everything to just double, if there are any
performance differences, both from accuracy as well as speed.

So a +1 from me on *typedef longdouble long double*

ibell:
Can you please undo the non-backwards-compatible changes to PropsSImulti and I will go ahead and merge?

ibell:
Let's do the typedef as well

mikekaganski:
I'm sorry for late reply, my timezone is UTC+10. :-) I'll do it as soon as possible.


Regards, Mike Kaganski

-------- Исходное сообщение --------
От: Ian Bell <notifications@github.com>
Дата:23.02.2015  14:42  (GMT+10:00)
Кому: CoolProp/CoolProp <CoolProp@noreply.github.com>
Копия: mikekaganski <mikekaganski@hotmail.com>
Тема: Re: [CoolProp] Reduce copying of parameters (#489)

Let's do the typedef as well

---

ibell:
No worries, there is no pressing rush.  I'm UTC-7 btw

mikekaganski:
> Let's do the typedef as well

Could you suggest a header where to put this typedef? Or should I create new?

jowr:
What about `datastructures.h`?

jowr:
https://github.com/CoolProp/CoolProp/blob/master/include/DataStructures.h

jowr:
actually, https://github.com/CoolProp/CoolProp/blob/master/include/CoolPropTools.h might be even better. That is where we determine the platform and I think this might be related...

PR #491, opened 2015-02-23
Avoid copying parameters

mikekaganski:
Here are the changes that should be safe.
I don't know how to completely remove some commits from my fork, so I have to create this request with those bad commit and its revert. Sorry.

### DIFF
--- a/dev/TTSE/TimeComp.py
+++ b/dev/TTSE/TimeComp.py
@@ -1,0 +1,1623 @@
+#!/usr/bin/python
+# -*- coding: ascii -*-
+#
+from __future__ import print_function, division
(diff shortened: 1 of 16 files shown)


mikekaganski:
I'm sorry that I make so much noise. I'm still learning.
Let me close this request, too, and try to start over, to make good request without artifact commits etc.

ibell:
Its quite fine, it took me a while to get the hang of git, github, pull
requests, etc..  It's fine with me if you leave it as is, its up to you.
Just let me know when it is at a stage to be merged.  In the end, all the
commits get squashed into one commit in the master, so its not a big
problem.

PR #495, opened 2015-02-24
CoolPropDbl

mikekaganski:
Introduce CoolPropDbl to replace long doubles. This may help to overcome
possible portability problems, and test for efficiency and precision.


ibell:
Ok, let's see how the testing looks.  I'm quite curious.  And then we can add a preprocessor macro that allows you to disable the mapping

mikekaganski:
Unfortunately, I cannot test it (I mean both performance and precision) - I only have MS VC++ under Win7x64, and there long double == double.

ibell:
Right, thats where our buildbot network is useful:
http://www.coolprop.dreamhosters.com:8010/

ibell:
I just fired off all the builders on that branch, let's see what the
results are - you can see it in the grid

PR #499, opened 2015-02-25
Longdouble remap

ibell:


ibell:
This is a great idea - thanks

PR #500, opened 2015-02-25
Fix directory size calculations

mikekaganski:


src/CoolPropTools.cpp
```diff
@@ -17,7 +17,7 @@
 #include <stdint.h>
 #include <iostream>
 
-static double ftw_summer; // An evil global variable for the ftw function
+static thread_local unsigned long long ftw_summer; // An evil global variable for the ftw function
```
ibell: +1 If you notice other things that are not thread-safe, PLEASE fix them.  Thread safeness is something I am only aware of in a limited fashion

ibell: I had to revert this thread_local change since a lot of our compilers (VS2008 especially) don't support C++11 unfortunately.

PR #504, opened 2015-02-26
More ref args

mikekaganski:


include/CoolPropTools.h
```diff
@@ -341,54 +341,62 @@
...
+                throw CoolProp::ValueError(format("%s could not be matched in get_number",s.c_str()));
             }
         };
-        double get_number(std::string s)
+        double get_number(const std::string &s) const
         {
             return get_double(s);
         };
-        const std::vector<double>& get_double_vector(std::string s)
+        const std::vector<double>& get_double_vector(const std::string &s) const
```
ibell: Isn't the const on the right not necessary?

include/CoolPropTools.h
```diff
@@ -616,50 +624,39 @@ template<class T> void normalize_vector(std::vector<T> &x)
         std::replace( file_path.begin(), file_path.end(), '\\', '/'); // replace all '\' with '/'
         
         #if defined(__ISWINDOWS__)
-        char sep = '\\';
+        const char sep = '\\'; // well, Windows (and DOS) allows forward slash "/", too :)
```
ibell: But some functions in WINAPI do not accept forward slashes as the path delimiter.  I'm not sure if that is the case within this function.  I'd be happy to change this if we can show it works with /

ibell:
Thanks for your work on CoolProp tidying up

include/CoolPropTools.h
mikekaganski: the rightmost const makes this function "constant function": it means that
1. we declare that the function does not change the state of the object.
2. thus, compiler does not allow us to change any of object's members, and only allows to call their "const" functions;

and only this allows to call this function, when the object itself is const.

ibell: Another C++ lesson for me!

ibell: So there is an impact on the optimization possible since it is very const?

mikekaganski: yes, and it was three-fold here.
1. I noticed the param-by-value Dictionary (ReducingFunctions.h line 327). The dictionary has four maps, that contain numbers, strings, and vectors of numbers and strings. Each has to be recursively copied, and released after use, and each object has its constructors and destructors. So, naturally, I wanted to pass it by const ref. But when I did that, the error appeared that the object cannot be **const**, as the methods called are not const.
2. I see that methods should not make changes to Dictionary - they are simple selectors. So I go and change their declarations as const, too. But compiler complains that as the method is const, it's illegal to call non-const methods of members there. I see, that the methods are implemented this way: first, the key is searched using **find()**, and only if found, the **operator []** is called, that makes search again. This is not optimal per se, but additionally, in maps, **operator []** is not const: it looks for the key, and if not found, it automatically adds new entry for this key, thus changing the object. This helped me to notice this inefficient implementation. So, I simplified things by making only one lookup using const method **find()**.
3. Further, it **may** help compiler generate more efficient code internally, even in places where the code is unchanged, as now it has a hint that the object's state stays unchanged.

include/CoolPropTools.h
mikekaganski: Honestly, I don't know which WinAPIs don't accept "/". It's better to know of them, so please give an example.
But either way, I don't feel like this should be changed, as it makes no difference efficiency-wise. This looks more robust, so I have no objections, only the notice.

mikekaganski:
Actually, it's honour for me to help people that give their time, knowledge and work to others. Ant who do it for free! Thank you.

PR #510, opened 2015-03-01
const, ref and iterator optimization

mikekaganski:
1. Another set of arguments optimized:
* args-by-val converted to args-by-ref
* in some cases, opposite has been done, if that is better: if the
object's copy is modified in the func, it's better to avoid extra copy
from arg to local var
2. some const functions marked as such (this is just a beginning)
3. iterators were replaced with const_iterators where applicable
4. catches that catch std::exception& changed to catch (...) - that is
safer; if exception handling is restructured, this will have to be
reconsidered anyway
5. removed some basic structures' constructors; changed them to hold
const char*s to avoid unnecessary string constructions;
6. in some places, moved variable declarations to their definitions, to
avoid calling default constructors and then assigning
7. removed some unnecessary shared_pointers in favor of local objects;
8. in FORTRAN-style functions, added const specifiers to input doubles;
9. fixed a place where values were inserted into a map while iterating
through it
10. fixed is_valid_*_derivative: they could accept incorrect values and
throw when upper index is less than lower
11. Simplified strsplit

### DIFF
--- a/src/Backends/Helmholtz/VLERoutines.cpp
+++ b/src/Backends/Helmholtz/VLERoutines.cpp
@@ -114,7 +114,7 @@ void SaturationSolvers::saturation_T_pure_1D_P(HelmholtzEOSMixtureBackend &HEOS,
     try{
         Secant(resid, options.p, options.p*1.1, 1e-10, 100, errstr);
     }
-    catch(std::exception &){
+    catch(...){
         CoolPropDbl pmax = std::min(options.p*1.03, static_cast<CoolPropDbl>(HEOS.p_critical()+1e-6));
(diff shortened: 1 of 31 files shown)


src/Backends/Helmholtz/VLERoutines.cpp
```diff
@@ -613,9 +613,9 @@ void SaturationSolvers::saturation_D_pure(HelmholtzEOSMixtureBackend &HEOS, Cool
         deltaV = rhoV/reduce.rhomolar;
         tau = reduce.T/T;
     }
-    catch(NotImplementedError &e)
+    catch(NotImplementedError &)
     {
-        throw e;
+        throw; // ??? What is this try...catch for?
```
ibell: I think this was for debugging purposes so I could put a breakpoint here but I forgot to remove it

ibell:
Thanks!

PR #513, opened 2015-03-01
Primitive structures simplification attempt 2

mikekaganski:
This time I tested everything under Ubuntu.

-> I reworked primitive structures simplification. This time I don't use c++11 syntax
```c++
var = struct{init_list}
```
Instead, good old structure initializer
```c++
var = {init_list}
```
is used. I'm doing this to allow compiler to compile these primitive structures into library as blobs, ready to use without initialization. They are anyway only needed to initialize smart classes. Now it looks like this:
- strings and numbers are stored in the library binary (data section) and loaded into the memory on library load
- this data is used to initialize array of structs (using a number of calls to struct contructor, and to std::string constructors inside)
- this array is used to initialize static info class object (through its constructor).

I hope to do this:
- array with strings and numbers is stored in library binary and loaded into memory
- this array is used to initialize static info class object

-> Removed <memory> include that I added to use std::shared_ptr (reverted by b28583f4b6e2e861b75a0293c07d9bf9bc0e4005).

-> Fixed compiler warning that deleting void* is undefined

Further commits will be tested under Linux, too. Hope that I will not create much troubles anymore. Sorry for being ignorant.

### DIFF
--- a/src/CoolPropTools.cpp
+++ b/src/CoolPropTools.cpp
@@ -4,7 +4,6 @@
 #include <cstdio>
 #include <cstdarg>
 #include <stdlib.h>
-#include <memory>
 #include "math.h"
 #include "stdio.h"
 #include "float.h"
--- a/src/CoolPropTools.cpp
+++ b/src/CoolPropTools.cpp
@@ -45,7 +45,7 @@ double root_sum_square(const std::vector<double> &x)
 std::string format(const char* fmt, ...)
 {
     const int size = 512;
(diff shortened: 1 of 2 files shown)


ibell:
And this should compile with C++03?

mikekaganski:
yes, I believe so

ibell:
Right, well lets see how it goes when I merge.  Are you getting error messages when builds fail on buildbot?  You should be I think...

mikekaganski:
> You should be I think
No I'm not. Should I?

Issue #512, opened 2015-03-01
Kill off the CRT deprecate warning


ibell:

mikekaganski:
I'll do it

PR #515, opened 2015-03-01
Kill off the CRT deprecate warning (#512)

mikekaganski:


ibell:
Two comments:

This should only be applied if it is windows and the compiler is MSVC (``#if defined _MSC_VER``)

For some reason that I don't understand, setting ``#define _CRT_SECURE_NO_WARNINGS`` didn't silence this warning

mikekaganski:
I suppose that you defined it (_CRT_SECURE_NO_WARNINGS) in place. It should be defined before header that declares the function, maybe at project level, and will silence out all such warnings, in all other possible cases, which may be undesirable

PR #534, opened 2015-03-10
Add error handling to Julia's wrapper

JonWel:
Add error handling to Julia's wrapper (works for Julia 0.4, still not working with Julia's 0.3 Ubuntu ppa).

Changed numeric input requirement to be ´Number´ instead of restricting to ´Float64´, that allows both integer and floats to be used as input with few overhead (the Float64 version is still internally created by Julia and the overhead for the conversion from Int64 to Float64 seems negligible compare to the execution time from the functions, I could only time a difference in jit compilation time).
Added ´K2F´,  ´get_global_param_string´ and ´get_param_index´ functions
Also add usage of Props1SI when two string arguments are given to PropsSI.

(I see that my update from the CoolProp GitHub to my cloned repository is put in the list of commits, if someone has advice for better practices to send only the needed commits, It's welcome).

PR #536, opened 2015-03-12
Add low level interface to Julia wrapper as discussed in #534

JonWel:
Low level access throw Julia wrapper is now possible.
See the source and updated web documentation for usage.


Web/coolprop/LowLevelAPI.rst
```diff
@@ -25,6 +25,46 @@ which yields the output:
 
 .. literalinclude:: snippets/AbstractState1.cxx.output
 
+Here is an example of the shared library usage with Julia wrapper::
+
+    julia> import CoolProp
+
+    julia> PT_INPUTS = CoolProp.get_input_pair_index("PT_INPUTS")
```
ibell: Does this function capture errors properly or just crash?

ibell: nevermind...

JonWel: On Julia 0.4, all the error I tested had adapted error messages.
On Julia 0.3, I still need the ppa to be corrected to have correct error handling.

This is possible that I did not test all the possible errors, but I tried to have the main possible ones (up to my knowledge).

It may be useful to have a list somewhere with the possible strings for the input pair.

jowr: Are you thinking of something like http://www.coolprop.org/_static/doxygen/html/namespace_cool_prop.html#a58e7d98861406dedb48e07f551a61efb ?

JonWel: Exactly, I think it would be great to link it to the low level documentation.
(I do not know yet how internal links in the documentation works)

jowr: Here is an example: https://github.com/CoolProp/CoolProp/commit/9c5a708f221a2ae154d68cc4b14656bd5741558b#diff-adfc138b4cc89b7e3ab2346946539625R70

Issue #547, opened 2015-03-20
Transport property status


jowr:
Hi, is there a list anywhere about the missing transport properties? I just noticed that pentane is not yet supported and I was wondering if there is a road map.

paarfi:
Hi jowr. Ian and I discussed some of this a couple months ago in issue #415. Hope that helps.

jowr:
Thanks, I was just wondering how many fluids have the proprietary correction polynomials for the ECS method. The main point is that we should document a feature regression (like no more transport properties available) for certain fluids.

ibell:
Were they there before in v4? I thought I had reimplemented everything from
v4

jowr:
Iam afraid so: http://www.coolprop.org/v4/Fluids/n-Pentane.html vs. http://coolprop.sourceforge.net/fluid_properties/fluids/n-Pentane.html#fluid-n-pentane

jowr:
But I do not think that the Lennard-Jones parameters are proprietary material...

ibell:
Yes that should be fine.  I'll fix it this weekend

jowr:
Do you have a good idea about how to find which fluids are afffected? It might be that there are quite a few, at least I remember that quite a few fluids used parameters from *Poling, Bruce E.; Prausnitz, John M.; O'Connell, John P., 2001, The Properties of Gases and Liquids, 5th edition, McGraw Hill*.

ibell:
Yes I could go back in and take a look.  I think I just didn't implement
the L-J parameters for those fluids.  The predictions are going to be not
good, but better than nothing I guess.

ibell:
There's a viscosity correlation for pentane here: http://pubs.acs.org/doi/abs/10.1021/jp0618577

ibell:
Its of a similar form to H2S and SF6, but it turns out that the docs of the transport routines/properties seem to not be very good. :(

jowr:
Should we move this one to 5.2 and release the next version? There have been plenty of fixes that justify a new release and I think it is not too long until 5.1.1 comes around with Modelica integration.

ibell:
Yes, let's do that.  I'm not sure what there is to do on this one anyway.

ibell:
So what else is missing that we could in principle implement?

jowr:
I am not sure. This issue was supposed to encourage to improve the documentation for the transport properties. I see many people using them due to the lack of alternatives, but hardly anyone realises how poor many of the approximations are.

[...]

PR #589, opened 2015-04-11
Implemented friction theory for methane and pentane viscosity

### LINKED ISSUE #574: Methane (and pentane) transport properties
Running module CoolProp.CoolProp.PropsSI for viscosity of methane results in an error stating "viscosity model is not available for this fluid."

Properties can be found in this document:
http://www.nist.gov/data/PDFfiles/jpcrd97.pdf

Thanks

ibell:
Closes #574


jl230:
This ticket may also affect #547

PR #612, opened 2015-04-21
Added missing cell "Input/Output" for enthalpy row.

postmath:
There appears to be a missing cell in this table. Added it.


ibell:
Thanks for the contribution!

PR #614, opened 2015-04-22
Typos in parameter descriptions.

postmath:
There were a few typos in the description field of `parameter_info_list`. If these strings are currently literally used somewhere, they should probably be changed there as well -- but I couldn't find anything, at least not in the CoolProp code base itself (except some comments in C sharp files, which I guessed would not be as important).

I think "presssure" and "gobal" are pretty uncontroversial typos; I would guess that the underscores in the description "fundamental_derivative_of_gas_dynamics" are also not meant to be there, but I could see that being subject to discussion.


ibell:
Typos all around. Doh!

ibell:
Thanks!

PR #667, opened 2015-05-12
If -DCMAKE_BUILD_TYPE is set on the cmake command line, do not overwr…

postmath:
…ite it.

Currently, there is some logic that says if COOLPROP_VERSION_REVISION is "dev",
then the build type is set to Debug by default, otherwise to Release. With
this change, that logic will only trigger if CMAKE_BUILD_TYPE was not set
already (e.g. on the command line with a -D option). The behaviour is
overridden by the COOLPROP_RELEASE and COOLPROP_DEBUG variables.

This makes it easier to integrate the CoolProp build into build systems that build
multiple projects and expect to be able to set CMAKE_BUILD_TYPE. (Maplesoft's
is an example of such build systems.)


jowr:
No problem, if that makes things easier for you.

postmath:
Thanks Jorrit.

Erik.

Issue #882, opened 2015-11-26
Tabular backends and phase specification


jowr:
I know that the tables can yields "wrong" results for inputs close to saturation. I tried to circumvent this by specifying the phase because I know the phase for all calls beforehand. Unfortunately, this does not seem to make a difference. This could be related to #739 and #656 

Is that a correct observation? Do we have plans about implementing such functionality?

ibell:
What inputs?  That's not a bad idea, to make sure you get the right phase
if phase is imposed

jowr:
Typically, I would use T,p inputs that are very close to saturation and then I impose the phase to tell CoolProp whether I want the results for gas or liquid.

ibell:
That makes sense to me.  I can see that being useful.

mapipolo:
So do I understand correctly that phase specification using AbstractState.specify_phase() is currently ignored for tabular calculations? It doesn't seem to make a difference for me either with QT or TP inputs. This feature would be very useful for me too, and could eliminate the need to fudge the pressure and temperature away from the saturation line to ensure that CoolProp calculates for a gas rather than a mixture (the "fudging" method is not especially reliable).

ibell:
Phase specification will never be needed or used for QT as that is always
two-phase.  PT, on the other hand, could be made much more reliable by
using the specified phase, which would fix the "bumping" issue.

ibell:
This is a little bit complex to implement - I'll have to think about how to resolve the problem.  Should be doable though.  Should greatly improve the reliability near saturation.

ibell:
I've made progress in tabular_refactor branch.  I merged update functions from TTSE and bicubic into the common backend.  Now only have one location to update code.  I'm about ready to merge it back in I think.

PT in liquid phase at low temperatures is very difficult - we quickly run out of space to bump to lower temperature, might have to consider using TTSE down there.

jowr:
I have some time to look at this today and can do some testing. What do you mean by *might have to consider using TTSE down there*? instead of bicubic?

ibell:
It seems that in the wedge near triple point pressure and between minimum
temperature and saturation temperature, it is quite often (actually most of
the time) that the algorithm yields an entirely wrong density as function
of t and p.  This is because all four nodes must be to left of saturation
curve to yield a truly liquid  solution.  In vapor phase, this is very
simple to resolve.  I think the only way to solve this for liquid near
triple point pressure is to use TTSE rather than bicubic.  In this way, you
[...]

Issue #873, opened 2015-11-18
Implement DS, DH, DT for tabular backends


paarfi:
Could you please add support for these iteration types on the tabular backends? DS and DH are useful for calculations partway through the compression stroke.

DT is more direct with the EOS, but I'm hoping it would still be faster iterating off of the tables for blends. If you don't think that would be the case, then maybe that's not as important to do.

ibell:
DT already works and is implemented.  And it is much faster for blends.

DS and DH are a bit more complicated, but can be done.  DS is problematic
since neither variable overlaps the variables for the two tables we build.

paarfi:
@ibell: You're right; DT is already implemented. My bad.

ibell:
See also #827

ibell:
Closing as part of an issue-tracker cleanup — this issue has had no activity in over two years.

*To be transparent: this comment and the close action are being performed by Claude (Anthropic's AI coding assistant), running on behalf of @ibell. Ian reviewed and approved the bulk-close criteria; I am the one executing the action.*

If you would like this issue reconsidered, please feel free to **re-open it** — a fresh comment with current behaviour against the latest CoolProp release is the most useful way to revive the discussion.

— Claude (acting for @ibell)

PR #896, opened 2015-12-12
Tabular refactor

ibell:
Refactor the tables, working towards #882


jowr:
Nice! Thanks for all the help with this!

ibell:
My pleasure, my hope is this will make it easier to add other inputs more easily. See for instance #827 and #873

Issue #907, opened 2015-12-24
Replace string formatting with C++ format library


ibell:
See https://github.com/cppformat/cppformat

ibell:
Looks like MSVC2008 is not supported...

ibell:
Wrong.  See here: https://msdn.microsoft.com/en-us/library/ms177415(v=vs.90).aspx

ibell:
Closed by #952

PR #952, opened 2016-02-01
Switch string formatting to use the cppformat library; see #907

ibell:


jowr:
Is there any particular reason for adding the sources to CoolProp? Would you mind if I moved the cppformat files to an external submodule in git?

ibell:
Yes of course that's fine.  Initially I thought that they had combined
several files into the release that's on github, but it seems that I was
wrong about that.  Go for it.  Make sure to update the CMakeLists.txt and
the python setup.py.

PR #960, opened 2016-02-08
Update index.rst

saha84:
For the path of the shared library in Linux, it is enough to put the 'libCoolProp.so' in '/usr/lib/scilab' and use the sample.sce file as it is now. I don't know the equivalent folder in windows.

note: As I say in my first comment, "English is not my mother tongue; please excuse any errors on my part."


ibell:
Can you also add that comment about the shared library?  Didn't know about that.  I'm merging this PR.

saha84:
hello,

I just updated the comment about the shared library, but I don't see it anywhere.

Sorry this is the first time I do this thing in github

ibell:
I think you need to make a new pull request.  The .rst files will get
compiled into the website automatically, so your changes you made in the
last pull are here:
http://www.coolprop.dreamhosters.com/binaries/sphinx/coolprop/wrappers/Scilab/index.html#scilab

PR #961, opened 2016-02-09
Update index.rst

saha84:
For windows maybe there are abetter way to do it.


ibell:
Please remove your comment about ``c:\windows`` for windows platform.  Usually you should not be putting DLL in ``c:\windows`` directory.  Better to modify the PATH environmental variable.

Issue #977, opened 2016-02-18
Access to coexisting 2-phase densities for REFPROP backend


jowr:
For a `T,D`-flash calculation, I need access to the densities of the coexisting vapour and liquid phase. The `AbstractState` has the cached variables `_rhoLmolar` and `_rhoVmolar`, but they only appear in the `REFPROPMixtureBackend` and there they are commented out. There also seem to be no functions to access them. 

Am I missing something or is there no simple way to get these properties?

ibell:
How about this:
http://www.coolprop.org/dev/coolprop/LowLevelAPI.html#things-only-in-the-low-level-interface
(18 & 19)

jowr:
Thanks for the hint, I found those as well. My problem is that I need something that works for HEOS *and* REFPROP. Guess I'll have to implement it myself :smile:

ibell:
Ah right, I had misunderstood your question.  You want it for the REFPROP
backend...

jowr:
Actually for both HEOS and REFPROP - I did not mention it in the first place and simply assumed that you would read my mind.

ibell:
Now I'm confused....  For HEOS it works just like in the example I linked.
Isn't that exactly what you want?  Or do you need it from the high-level
interface?

ibell:
So, in summary, you are fine for HEOS, but need to implement the necessary
accessor function for REFPROP.  Correct?

jowr:
Correct, the only thing that confused me, was that there is a function for HEOS and that there are `_rhoLmolar` and `_rhoVmolar`, which are only used in the REFPROP backend. 

**Solution:** Implement `saturated_liquid_keyed_output` and `saturated_vapor_keyed_output` for the REFPROP backend and remove `_rhoLmolar` and `_rhoVmolar` from the base class as they are only used in one backend.

ibell:
I'm going to need a similar formulation for the cubic backend as well, but
I guess you can drop those variables from the base class.

ibell:
Closed by #981

PR #981, opened 2016-02-22
Saturation states

jowr:
This is a very quick fix for #977 , but it serves my purpose. @ibell , do you see any problems merging this?


src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
```diff
@@ -203,8 +199,8 @@ void HelmholtzEOSMixtureBackend::set_binary_interaction_double(const std::string
 
     // Update the values in this instance and the saturation states too
     this->set_mixture_parameters();
-    if (this->SatL){ this->SatL->set_mixture_parameters(); }
```
ibell: I actually prefer having the braces - it's entirely personal style, but that way if you add to the  actions taken in the true branch, you can be sure you have the correct scope.

ibell:
Nope, I think that seems fine to me.

PR #1041, opened 2016-04-28
Fixed Minor MSVC Compiler Warnings

henningjp:
int vs unsigned in several comparisons and one initialization.  Wrapped the unsigned values with int().


ibell:
While int() does work, it is generally recommended to do ``static_cast<int>(...)`` rather than ``int(...)``.  See for instance google's style guide: https://google.github.io/styleguide/cppguide.html#Casting

Issue #946, opened 2016-01-26
Coolprop for ACADO


adexcoca:
Hi,
I am using Coolprop developed for Matlab. Recently, I am working on model predictive control where I will generate the real time optimizer using ACADO toolkit developed for Matlab. Is it possible to use in the Coolprop inside my control code and autogenerate the real time optimizer? 

Thank you,

ibell:
Sadly, I have no idea.  Also, I don't have access to MATLAB.  My guess is
probably no since it would probably have to compile your MATLAB code and I
doubt it can link in the CoolProp SWIG interface.

It is possible that you could load the CoolProp DLL using the loadlibrary()
function.  Somewhat analogously to this MATLAB example:
http://www.coolprop.org/coolprop/wrappers/MATLAB/index.html#calling-low-level-interface-through-dll

If you get this working, I would definitely appreciate it if you would add
a small description to the CoolProp documentation.

Astrinus:
@adexcoca You could create a MEX C++ function and then call CoolProp from it. ACADO documentation says it is supported. I think it is not harder than writing a C++ S-Function for Simulink.

adexcoca:
I am using Matlab both for ACADO and Coolprop. The problem is I am not
familiar with C++.
ACADO for Matlab has capability to create MEX function directly from Matlab
script.

Let me try, thank you for the suggestion. I will share you the outcomes.

On Tue, Jan 26, 2016 at 3:00 AM, Astrinus <notifications@github.com> wrote:

> @adexcoca <https://github.com/adexcoca> You could create a MEX C++
> function and then call CoolProp from it. ACADO documentation says it is
> supported. I think it is not harder than writing a C++ S-Function for
> Simulink.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/CoolProp/CoolProp/issues/946#issuecomment-174882293>.
>



-- 
Adamu Yebi, Ph.D.
Research Associate
International Center for Automotive Research
Clemson University
Greenville, SC 29607, USA

ibell:
Any update?

adexcoca:
Hi,

I am working on another issue at the moment, but will dive soon toward
Coolprop and ACADO integration.
So far what I learn is, ACADO may not accept any table reference except
something in terms of empirical relationship.
I will continue researching on limitation and capability of ACADO plus
Coolprop.

Thank you,

[...]

PR #1060, opened 2016-05-11
Export first_partial_deriv, see #946

JonWel:
The file `CoolPropLib.def` has not been updated in in this patch.

### DIFF
--- a/include/CoolPropLib.h
+++ b/include/CoolPropLib.h
@@ -295,7 +295,20 @@
     * @return
     */
     EXPORT_CODE double CONVENTION AbstractState_first_saturation_deriv(const long handle, const long Of, const long Wrt, long *errcode, char *message_buffer, const long buffer_length);
-            
+    
+    /**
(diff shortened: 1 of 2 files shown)


ibell:
Can you update the .def too please?

JonWel:
I've never did that (the .def), how should I find the correct number after the @ ?

```c++
AbstractState_first_partial_deriv = _AbstractState_first_partial_deriv@??
```

ibell:
When you build the dll with msvc I thought we wrote the exported symbols to
a exports.txt with dumpbin

JonWel:
Ok, so I hove to compile on Windows.
(I did only checked on Linux)

ibell:
Don't worry then, I'll fix that, thanks for the code...

JonWel:
This is also good for me to know:
```C++
AbstractState_first_partial_deriv = _AbstractState_first_partial_deriv@28
```
Should be the correct one, I add it to the patch and make a new pull request.

Issue #985, opened 2016-02-24
Updates to numerical solvers 


ianhbell:
- [x] ND numerical solver should quit if x stops changing
- [x] Relax some convergence criteria
- [x] Store the number of calls in the function wrapper base class
- [x] Store the return code and error string in the base class

PR #1080, opened 2016-05-20
Enable infinity in MakeDoubleResult

dcprojects:
Looking around I think this change should be related to the issue #985
Since double.NegativeInfinity and double.PositiveInfinity are valid results from the solvers and SMath.Math.Numeric.TDouble(result) can't handle these values directly, this patch allows the users to see  ∞ / -∞ as results on the canvas

### DIFF
--- a/wrappers/SMath/coolprop_wrapper/coolpropPlugin.cs
+++ b/wrappers/SMath/coolprop_wrapper/coolpropPlugin.cs
@@ -59,6 +59,25 @@ namespace coolprop_wrapper
 
     public static Term[] MakeDoubleResult(double result, SMath.Math.Symbolic.MItem unit)
     {
+        // override ResultIsAboveMaxAllowedPositiveNumber exception for +/- infinity results
+        if (double.IsInfinity(result))
+        {
+          List<Term> xTerms = new List<Term>();
+
+          xTerms.Add(new Term(Symbols.Infinity, TermType.Operand, 0));
+
+          if (result == double.NegativeInfinity)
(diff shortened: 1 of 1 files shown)


ibell:
I'm not a SMath user, so I'm going to just trust you on this one...

ibell:
This broke the SMath builds: http://www.coolprop.dreamhosters.com:8010/builders/SMath/builds/846/steps/shell_1/logs/stdio  Please figure out how to unbreak this, or I will have to revert this PR.

ibell:
Error is: ``coolpropPlugin.cs(70,45): error CS0117: 'SMath.Manager.Operators' does not contain a definition for 'Subtraction' ``

dcprojects:
It depends on what is the target build. For SMath Studio <= 0.97.5346 there is the Operators.Sub**s**traction; this typo was fixed in later versions, removing it and replacing with Operators.Subtraction (without s); in later APIs (0.98) there are both the constants, but the old one is marked as outdated. Therefore there are 3 possible cases:
- only substraction -> SMath <= 0.97.5346)
- only subtraction -> 0.97.346 < SMath < 0.98)
- both (with the older deprecated) -> SMath < 0.98

In case of removing, an error will be shown in the canvas. A trap function is available - try(1), but only for very recent versions (0.98.5952 Nightly Build).

ibell:
So if I updated SMath on the build machine to 0.98, that should fix this
issue, right?

dcprojects:
Correct :)

ibell:
Ok good, I'll do that ASAP then.

ibell:
Yup, that does it.  Thanks for the info!

Issue #1077, opened 2016-05-20
SMath wrapper: CoolProp_get_parameter_information_string always returns long description


mikekaganski:
I have upgraded to 6.0.0, and now my SMath wrapper function CoolProp_get_parameter_information_string(Param,Output) gives always only long description.
Before, it used to give the result dependent on Output string: "full", "short", "IO" and "unit".

I do have a SMath file with saved calculations showing that it was different at one point.
[CoolProp_demo.sm.zip](https://github.com/CoolProp/CoolProp/files/274540/CoolProp_demo.sm.zip)

The history for src/CoolPropLib.cpp tells that commit 763d4ceb11da081df6653dec29d54e4ba1ff2cc2 changed this, so that Output isn't input/output anymore, but only output. Before, the value passed through it, was used as column key.

Coudn't that be reverted, and #974 be corrected in a different way?

ibell:
@jowr Please make a PR reverting this change so that this function is able to return any of the desired parameters

PR #1093, opened 2016-06-02
Revert part of 763d4ce to solve #1077

JonWel:
And adapt the julia wrapper so it still works.

### DIFF
--- a/wrappers/Julia/CoolProp.jl
+++ b/wrappers/Julia/CoolProp.jl
@@ -54,15 +54,13 @@ function get_global_param_string(Key::AbstractString)
 end
 
 # CoolProp::get_parameter_information_string
-function get_parameter_information_string(Key::AbstractString)
-message_buffer[1:5] = b"long\0" # "long" in Byte Array Literals
+function get_parameter_information_string(Key::AbstractString,OutType::AbstractString)
+  message_buffer[1:length(OutType)+1] = [OutType.data; 0x00]
   val = ccall( (:get_parameter_information_string, "CoolProp"), Clong, (Ptr{UInt8},Ptr{UInt8},Int), Key,message_buffer::Array{UInt8,1},buffer_length)
   return bytestring(convert(Ptr{UInt8}, pointer(message_buffer::Array{UInt8,1})))
 end
-function get_parameter_information_string(Key::AbstractString,OutType::AbstractString)
-message_buffer[1:length(OutType)+1] = [OutType.data; 0x00]
-  val = ccall( (:get_parameter_information_string, "CoolProp"), Clong, (Ptr{UInt8},Ptr{UInt8},Int), Key,message_buffer::Array{UInt8,1},buffer_length)
-  return bytestring(convert(Ptr{UInt8}, pointer(message_buffer::Array{UInt8,1})))
+function get_parameter_information_string(Key::AbstractString)
+  return get_parameter_information_string(Key,"long")
 end
 
 # CoolProp::get_mixture_binary_pair_data
--- a/wrappers/Julia/CoolProp.jl
+++ b/wrappers/Julia/CoolProp.jl
@@ -55,6 +55,7 @@ end
 
 # CoolProp::get_parameter_information_string
 function get_parameter_information_string(Key::AbstractString)
+message_buffer[1:4] = UInt8[108 111 110 103] # "long" written in UInt8
   val = ccall( (:get_parameter_information_string, "CoolProp"), Clong, (Ptr{UInt8},Ptr{UInt8},Int), Key,message_buffer::Array{UInt8,1},buffer_length)
   return bytestring(convert(Ptr{UInt8}, pointer(message_buffer::Array{UInt8,1})))
 end
--- a/wrappers/Julia/CoolProp.jl
+++ b/wrappers/Julia/CoolProp.jl
@@ -55,12 +55,12 @@ end
 
 # CoolProp::get_parameter_information_string
 function get_parameter_information_string(Key::AbstractString)
-message_buffer[1:4] = b"long" # "long" in Byte Array Literals
+message_buffer[1:5] = b"long\0" # "long" in Byte Array Literals
   val = ccall( (:get_parameter_information_string, "CoolProp"), Clong, (Ptr{UInt8},Ptr{UInt8},Int), Key,message_buffer::Array{UInt8,1},buffer_length)
   return bytestring(convert(Ptr{UInt8}, pointer(message_buffer::Array{UInt8,1})))
 end
 function get_parameter_information_string(Key::AbstractString,OutType::AbstractString)
-message_buffer[1:length(OutType)] = OutType.data
+message_buffer[1:length(OutType)+1] = [OutType.data; 0x00]
   val = ccall( (:get_parameter_information_string, "CoolProp"), Clong, (Ptr{UInt8},Ptr{UInt8},Int), Key,message_buffer::Array{UInt8,1},buffer_length)
   return bytestring(convert(Ptr{UInt8}, pointer(message_buffer::Array{UInt8,1})))
 end
--- a/wrappers/Julia/CoolProp.jl
+++ b/wrappers/Julia/CoolProp.jl
@@ -55,7 +55,7 @@ end
 
 # CoolProp::get_parameter_information_string
 function get_parameter_information_string(Key::AbstractString)
-message_buffer[1:4] = UInt8[108 111 110 103] # "long" written in UInt8
+message_buffer[1:4] = b"long" # "long" written in UInt8
   val = ccall( (:get_parameter_information_string, "CoolProp"), Clong, (Ptr{UInt8},Ptr{UInt8},Int), Key,message_buffer::Array{UInt8,1},buffer_length)
   return bytestring(convert(Ptr{UInt8}, pointer(message_buffer::Array{UInt8,1})))
 end
--- a/wrappers/Julia/CoolProp.jl
+++ b/wrappers/Julia/CoolProp.jl
@@ -55,7 +55,12 @@ end
 
 # CoolProp::get_parameter_information_string
 function get_parameter_information_string(Key::AbstractString)
-message_buffer[1:4] = b"long" # "long" written in UInt8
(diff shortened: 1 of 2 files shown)


wrappers/Julia/CoolProp.jl
```diff
@@ -55,6 +55,7 @@ end
 
 # CoolProp::get_parameter_information_string
 function get_parameter_information_string(Key::AbstractString)
+message_buffer[1:4] = UInt8[108 111 110 103] # "long" written in UInt8
```
ibell: Is this really necessary?

mikekaganski: Well, this is because get_parameter_information_string() wrapper in Julia was implemented to take only one argument. So, if it is to be kept (for backward compatibility?), it should return some meaningful information, which is hardcoded here.
It's another question if this wrapper should stay or if it should be replaced with another one with two arguments, following the underlying API.

JonWel: A better way would be to have two arguments in this wrapper, than to convert that argument to a UInt8 UTF8 string, than to place it inside the buffer.
It is necessary to place it in the buffer as the buffer is used to take the result string back.

I did not did it as I've just explained because I didn't find yet the function to make that conversion (I need to have a UInt8 array, and the functions I found until now keep being of type string, even if internally, they are a compatible UInt8 array).
So I put this workaround that allows the function to be still working.

JonWel: I can do a conversion char by char in a loop, this would work:
```Julia
function get_parameter_information_string(Key::AbstractString,OutType::AbstractString)
  for k = 1:length(OutType)
     message_buffer[k] = UInt8(OutType[k])
  end
  val = ccall( (:get_parameter_information_string, "CoolProp"), Clong, (Ptr{UInt8},Ptr{UInt8},Int), Key,message_buffer::Array{UInt8,1},buffer_length)
  return bytestring(convert(Ptr{UInt8}, pointer(message_buffer::Array{UInt8,1})))
end
```
Would you like that better?

mikekaganski: http://docs.julialang.org/en/release-0.4/stdlib/strings/
isn't bytestring(s) OK?

JonWel: @mikekaganski I first thought too, but the result is a `ASCIIString`. I also tried `utf8()`, but then the result is `UTF8String`.
So I can't insert them in the buffer as it requires `Array{UInt8,1}`.

Theses strings could be send to the CoolProp library, but the problem is that I don't have the buffer anymore to get the result back.

mikekaganski: Coudn't you then use reshape() on string's .data? To get its copy as a UInt8 buffer of required size?
I'm sorry that I don't know nothing about Julia, just guessing.

JonWel: @mikekaganski not exactly, but I could define the buffer to be a string full of "\0":

```julia
message_buffer = repeat("\0",buffer_length)
function get_parameter_information_string(Key::AbstractString,OutType::AbstractString)
  message_buffer = OutType*message_buffer[length(OutType):end]
  val = ccall( (:get_parameter_information_string, "CoolProp"), Clong, (Ptr{UInt8},Ptr{UInt8},Int), Key,message_buffer::Array{UInt8,1},buffer_length)
  return bytestring(convert(Ptr{UInt8}, pointer(message_buffer::Array{UInt8,1})))
end
```

With the restriction that strings can't be modified, so I have to replace it with the two peaces of string (the `*` operator allows to attach two strings together. Julia is quite type sensitive)

mikekaganski: Great! :)

JonWel: @mikekaganski This `string.data` is exactly what I needed, this makes things way simpler :) :
```julia
function get_parameter_information_string(Key::AbstractString,OutType::AbstractString)
  message_buffer[1:length(OutType)] = OutType.data
  val = ccall( (:get_parameter_information_string, "CoolProp"), Clong, (Ptr{UInt8},Ptr{UInt8},Int), Key,message_buffer::Array{UInt8,1},buffer_length)
  return bytestring(convert(Ptr{UInt8}, pointer(message_buffer::Array{UInt8,1})))
end
```

With the buffer remaining the same as before.

mikekaganski: Is the buffer null-terminated then?

mikekaganski: @JonWel , personally I would make a local variable like that:

local_buffer=(s*repeat("\0",buffer_length)).data

That way, you have buffer of bigger size, zero-terminated, local. I hope that Julia has scope concept?

JonWel: The Array{UInt8,1} is initialized with `0x00`, so yes, null terminated if `OutType` is smaller than the size of the buffer.

I made the patch (I kept the function with only one argument in case someone would prefer it, overloading is working well in Julia).
However, during the testing, I tend to have the first call going well, but not the next, telling that the parameter is invalid (whereas it is valid as it also works sometimes).

JonWel:
```julia
julia> using CoolProp

julia> get_parameter_information_string("Dmolar","long")
"Molar density"

julia> get_parameter_information_string("Dmolar","long")
"parameter is invalid: Dmolar"
```

mikekaganski: Then you could try to dump the buffer for debugging.
Actually, I still suspect not-null-terminating case: first, your buffer is created zero-filled, then you copy "long" there, after that, result is put there that is longer, and then you copy "long" (4 bytes) to buffer already filled with previous grabage.

mikekaganski: And you yourself made the c++ function to only report first argument "param" on error:

catch(...){
 		str2buf(format("parameter is invalid: %s", param), Output, n);
 	}

JonWel: Oh, the message buffer needs to be reset!
That was not needed before as the buffer was always cleaned by CoolProp when writing error message, but here is needed to force this 0 after the last character.
I correct this.

mikekaganski: Why do you use global buffer at the first place???

JonWel: @mikekaganski So the last character in string is now forced to 0 in the last commit I added.

The idea of this global buffer is to allocate the resources only once in the Julia side.
I don't want to allocate 255 bites at each call of a function that may return an error, that would have a really bad impact on functions like `AbstractState_update`.

JonWel: @mikekaganski You are right, the error caching is a bit obsolete now, as there is two possible wrong parameters.
```C++
catch(...){
str2buf(format("parameter is invalid: %s", param), Output, n);
}
```
I think it would be good that I also update that ^^

JonWel: @mikekaganski I think it is better this way.
It does not tell anymore the first parameter is wrong when this is only the second that do not work correctly, which help debugging ^^.

ibell:
Thanks @JonWel and @mikekaganski!

wrappers/Julia/CoolProp.jl
JonWel: @mikekaganski So, yes, for the specific function in use, a local buffer would also work fine as the function is not really intended to be run tenth of thousands of times.
But as there is already the global buffer present, I found that replacing only the needed part in the buffer was a more elegant solution (once this `0x00` has been taken care of).

mikekaganski: Heh.

strlen(get_global_param_string("parameter_list"))=1606
strlen(get_global_param_string("fluids_list"))=1071
strlen(get_global_param_string("predefined_mixtures"))=1091

JonWel: So some of the `get_global_param_string` can require up to 1606 characters, so quite way more than 254 (255 minus the last one being the terminating 0). I did not knew that, and it has some importance.
This may in deed be in favour to have long enough buffer to handle such long outputs.
In an other way, as only the pointer is passed to the library, keeping the single buffer but larger will just slow by a little the loading of the library, have no impact on the functions execution, and allowing any output message (up to this size of course).
So for me the best compromise for Julia’s wrapper would be to set the buffer length to 1607. What's your opinion about that?

mikekaganski: @JonWel , in SMath wrapper, I use buffer of 10 000 chars long.
I don't think limiting it too much is a good idea, because e.g. fluids list is constantly growing, and it may be required to constantly monitor this.

JonWel: Yes, it's not a heavy operation:
```Julia
julia> @time A = Array(UInt8, 10000)
  0.000009 seconds (157 allocations: 52.042 KB)
```
And the computer in use is not the best ever (a tablet suffering from the heat).

That's not that that will makes the wrapper becoming heavy.

And about the idea to keep a single buffer of this size? That makes also the error handling benefit from this larger buffer.

mikekaganski: @JonWel : about the buffer - it's up to you, but then you are in charge for guaranteeing that there will be no possibility that two functions will try to use it at the same time (like writing into this buffer a string that is based on this buffer's contents from other function). AFAICT, it doesn't seem to be a problem unless it's not thread-safe. Are Julia's global variables thread-local?

JonWel: @mikekaganski When parallel processing is in use, the module is loaded independently on each thread. So each thread will have it's own buffer.
```Julia
$ julia -p 2
julia> using CoolProp
julia> fetch(@spawnat 1 CoolProp.message_buffer)
255-element Array{UInt8,1}:
 0x11
 0x27
 0xcf
 0xfc
 0x05
    ⋮
 0x00
 0x00
 0x00
 0x00

julia> fetch(@spawnat 2 CoolProp.message_buffer)
255-element Array{UInt8,1}:
 0x00
 0x00
 0x00
 0x00
 0x00
    ⋮
 0x00
 0x00
 0x00
 0x00
```
The two buffer have different content depending on the thread. If I modify the one of the 1rst thread, only that one is changed.

Thanks for your advices

PR #1097, opened 2016-06-04
Set error string in get_parameter_information_string() and fix SMath wrapper : fixes #1096

### LINKED ISSUE #1096: a858417 follow-up: use CoolProp::set_error_string() instead of returning error string in get_parameter_information_string()
Previously, get_parameter_information_string() used to return its error string in its result buffer. This was the case long ago, at least since 9299cb8e073bef6254815ed2e49db096be555ca8.

This is a problem in wrappers:
some of them (like SMath) detect error return code (0) and ignore the returned buffer contents;
others (like Julia by @JonWel ) seem to ignore the returned code, but instead always return the buffer contents, thus disallowing programmatic detection of the error.

The proposal is to follow the practice of using CoolProp::set_error_string() for that. If a wrapper still needs the previous behaviour, then it may use get_global_param_string("errstring").

mikekaganski:


src/CoolPropLib.cpp
```diff
@@ -101,99 +101,101 @@ double convert_from_SI_to_kSI(long iInput, double value)
 }
 
 EXPORT_CODE long CONVENTION redirect_stdout(const char* file){
+    fpu_reset_guard guard;
     FILE *fp = freopen(file, "a+", stdout);
-    reset_fpu();
```
mikekaganski: Actually, I don't know why fpu resetting is required here. Does freopen() do floating-point math?

ibell: No, that should be fine to remove the reset_fpu() call

src/CoolPropLib.cpp
```diff
@@ -18,27 +18,27 @@
...
+    throw CoolProp::ValueError("Buffer size is too small");
 }
 
 // In Microsoft Excel, they seem to check the FPU exception bits and error out because of it.  
 // By calling the _clearfp(), we can reset these bits, and not get the error
 // See also http://stackoverflow.com/questions/11685441/floating-point-error-when-calling-dll-function-from-vba/27336496#27336496
 // See also http://stackoverflow.com/questions/16849009/in-linux-do-there-exist-functions-similar-to-clearfp-and-statusfp for linux and OSX
-void reset_fpu()
-{
+struct fpu_reset_guard{
```
ibell: I like what you did here with turning the call into a class where the destructor is fired at the end.

wrappers/SMath/coolprop_wrapper/install.bat.template
```diff
@@ -1,4 +1,4 @@
 
 set VERSION=${COOLPROP_VERSION}
-xcopy %VERSION% %AppData%\SMath\extensions\plugins\2be8d14b-ba62-4002-a959-9e5dad905310\%VERSION%\ /S
-xcopy config.ini %AppData%\SMath\extensions\plugins\2be8d14b-ba62-4002-a959-9e5dad905310\config.ini
\ No newline at end of file
+xcopy %VERSION% %AppData%\SMath\extensions\plugins\ca92ef03-c7da-4888-98ad-528482733e2f\%VERSION%\ /S
```
mikekaganski: I definitely did...

ibell: Ah yes, sorry my apologies.  BTW, why did you change the GUID?

mikekaganski: Oh, that's because it's the GUID that is used in downstream SMath repo, and it's now much more widely used, so I think it's good to standardise on one. Otherwise SMath complains that they are incompatible

PR #1103, opened 2016-06-06
One small tweak to Props1

mikekaganski:
Do not try to call possibly throwing get_parameter_index if Props1SI
already returned an error

Also, add include/cubic_fluids_schema_JSON.h to gitignore

### DIFF
--- a/src/CoolPropLib.cpp
+++ b/src/CoolPropLib.cpp
@@ -131,6 +131,10 @@ EXPORT_CODE int CONVENTION set_reference_stateD(const char *Ref, double T, doubl
 EXPORT_CODE double CONVENTION Props1(const char *FluidName, const char *Output){
     fpu_reset_guard guard;
     double val = Props1SI(Output, FluidName);
+    if (!ValidNumber(val))
+        // Error code was already set in Props1SI
+        return val;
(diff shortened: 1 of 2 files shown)


src/CoolPropLib.cpp
```diff
@@ -131,6 +131,10 @@ EXPORT_CODE int CONVENTION set_reference_stateD(const char *Ref, double T, doubl
 EXPORT_CODE double CONVENTION Props1(const char *FluidName, const char *Output){
     fpu_reset_guard guard;
```
mikekaganski: Ian, could you comment on this line:
I'm unsure if fpu reset is required here.
In this function, Props1SI already does that; but after that, two float-point operations happen:

1. ValidNumber() compares (potentially _HUGE) to max/min float (I don't know if this may set some fpu state flag);
2. Valid result is divided by 1000.0. I suppose that underflow is possible here.

So, is the fpu reset actually needed here?

ibell: When in doubt, reset the guard. I'd say keep the fpu reset

mikekaganski: Got it; thanks!

PR #1117, opened 2016-06-09
Correct AbstractState_set_binary_interaction_double in shared library

JonWel:
See comments in c16a4b6


ibell:
Thanks, looks good!

Issue #1127, opened 2016-06-17
PQ_INPUTS and QT_INPUTS crash for cubic mixtures.


JonWel:
It must be caused by a very recent change, this input combination was still working recently with low level interface.
```Julia
julia> using CoolProp

julia>  PropsSI("Dmolar", "P", 1e5, "T", 400, "PR::R125[0.17956]&R134a[0.43933]&R32[0.38111]")
30.247210923254897

julia> PropsSI("Dmolar", "P", 1e5, "Q", 0, "HEOS::R125[0.17956]&R134a[0.43933]&R32[0.38111]")
15977.479239072072

julia> PropsSI("Dmolar", "P", 1e5, "Q", 0, "PR::R125[0.17956]&R134a[0.43933]&R32[0.38111]")

Please submit a bug report with steps to reproduce this fault, and any error mes
sages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION at 0x7fedcc32494 -- unknown function (ip:
000007FEDCC32494)
unknown function (ip: 000007FEDCC32494)
unknown function (ip: 000007FEDCC1D1CC)
unknown function (ip: 000007FEDCC1DDBC)
unknown function (ip: 000007FEDCCAD16D)
unknown function (ip: 000007FEDCC3FA26)
unknown function (ip: 000007FEDCC7848D)
unknown function (ip: 000007FEDCC1D516)
unknown function (ip: 000007FEDCC1E5A3)
unknown function (ip: 000007FEDCCEFD41)
unknown function (ip: 000007FEDCCF037F)
unknown function (ip: 000007FEDCCEEA3A)
PropsSI at C:\Users\s161576\AppData\Local\Julia-0.4.5\bin\CoolProp.DLL (unknown line)
PropsSI at C:\Users\s161576\.julia\v0.4\CoolProp\src\CoolProp.jl:34 
jlcall_PropsSI_1402 at  (unknown line)
jl_apply_generic at C:\Users\s161576\AppData\Local\Julia-0.4.5\bin\libjulia.dll (unknown line)
jl_interpret_toplevel_expr at C:\Users\s161576\AppData\Local\Julia-0.4.5\bin\lib julia.dll (unknown line)
jl_interpret_toplevel_thunk_with at C:\Users\s161576\AppData\Local\Julia-0.4.5\bin\libjulia.dll (unknown line)
[...]

PR #1129, opened 2016-06-21
Correct vector size in the cubic backend, solves #1127

JonWel:


ibell:
Looks good, thank you!

PR #1132, opened 2016-06-22
Cubic calc_molar_mass, closes #1124

### LINKED ISSUE #1124: Mass-based properties don't work for cubic backends
```python
import CoolProp.CoolProp as CP, numpy as np, matplotlib.pyplot as plt

for fluid in ['PROPANE']:
    Ts = np.linspace(100, 290)
    rhoL = CP.PropsSI('D','T',Ts,'Q',0,'REFPROP::'+fluid)
    rhoL_PR = CP.PropsSI('D','T',Ts,'Q',0,'PR::'+fluid)
    rhoL_SRK = CP.PropsSI('D','T',Ts,'Q',0,'SRK::'+fluid)
    plt.plot(Ts, rhoL, label = 'REFPROP')
    plt.plot(Ts, rhoL_PR, label = 'PR')
    plt.plot(Ts, rhoL_SRK, label = 'SRK')
plt.legend()
plt.show()
```

JonWel:
Implement `calc_molar_mass()` for the cubic backend.

### DIFF
--- a/src/Backends/Cubics/CubicBackend.h
+++ b/src/Backends/Cubics/CubicBackend.h
@@ -31,6 +31,7 @@ class CubicResidualHelmholtz;
 class AbstractCubicBackend : public HelmholtzEOSMixtureBackend  {
 protected:
     shared_ptr<AbstractCubic> cubic;
+    std::vector<CubicLibrary::CubicsValues> components; ///< The components that are in use
 public:
 	
 	/// Set the pointer to the residual helmholtz class, etc.
@@ -157,6 +158,8 @@ public:
     
     /// Cubic backend flashes for PQ, and QT
     void saturation(CoolProp::input_pairs inputs);
+
+    CoolPropDbl calc_molar_mass(void);
     
     void set_binary_interaction_double(const std::size_t i1, const std::size_t i2, const std::string &parameter, const double value);
     double get_binary_interaction_double(const std::size_t i1, const std::size_t i2, const std::string &parameter);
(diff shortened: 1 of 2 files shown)


src/Backends/Cubics/CubicBackend.h
```diff
@@ -31,6 +31,7 @@ class CubicResidualHelmholtz;
 class AbstractCubicBackend : public HelmholtzEOSMixtureBackend  {
 protected:
     shared_ptr<AbstractCubic> cubic;
+    std::vector<CubicLibrary::CubicsValues> components; ///< The components that are in use
```
JonWel: In `HelmholtzEOSMixtureBackend` components is of type `std::vector<CoolPropFluid>`.
The two types are not compatible, for example, in `CoolPropFluid` molar mass is called by `components[i].molar_mass()`, whereas in the cubic it is `components[i].molemass`.
This also explain the need to a new cubic `calc_molar_mass()` definition.

src/Backends/Cubics/CubicBackend.h
```diff
@@ -200,11 +203,12 @@ class SRKBackend : public AbstractCubicBackend  {
                bool generate_SatL_and_SatV = true){
         std::vector<double> Tc, pc, acentric;
         N = fluid_identifiers.size();
+        components.resize(N);
         for (std::size_t i = 0; i < fluid_identifiers.size(); ++i){
-            CubicLibrary::CubicsValues val = CubicLibrary::get_cubic_values(fluid_identifiers[i]);
-            Tc.push_back(val.Tc);
-            pc.push_back(val.pc);
-            acentric.push_back(val.acentric);
+            components[i] = CubicLibrary::get_cubic_values(fluid_identifiers[i]);
```
JonWel: Storing the values here to use them in `calc_molar_mass()`.

ibell: Looks good.  It would be perhaps a good idea to write a common setter function for Tc, pc, acentric for both of the cubic backends in the AbstractCubicBackend.  For now this is fine...

PR #1134, opened 2016-06-22
Export specify_phase() to the library, closes #1133

### LINKED ISSUE #1133: Export specify_phase in the shared library
I would like to export the function `specify_phase()` and `unspecify_phase()` in the shared library.

The `unspecify_phase()` is quite straightforward.

For the `specify_phase()`, I see different possibilities.

1. Two functions `specify_phase_liquid()` and `specify_phase_gas()`.
2. One function, `specify_phase("string")`, with an internal switch case to call the `specify_phase(CoolProp::iphase_liquid)` or `specify_phase(CoolProp::iphase_gas)`.
3. One function, `specify_phase(double param)`, which would call directly `specify_phase(static_cast<CoolProp::parameters>(param))`.
4. ?

Which solution would be preferred?
For me the first solution would be easier to use, the second has some overhead with the switch case, and the third needs to get the `param` value first.

Rem: the arguments are simplified here, of course the library functions would have at least the complete `(const long handle, long *errcode, char *message_buffer, const long buffer_length)` inputs.

JonWel:
Add `specify_phase()` to the library
Add `AbstractState_update_and_1_out` to the library
Add the functions in the Julia wrapper
Some enhancement in the Julia wrapper

### DIFF
--- a/wrappers/Julia/CoolProp.jl
+++ b/wrappers/Julia/CoolProp.jl
@@ -1,7 +1,7 @@
 VERSION < v"0.4.0" && __precompile__()
 module CoolProp
 
-export PropsSI, PhaseSI, get_global_param_string, get_parameter_information_string,get_fluid_param_string,set_reference_stateS, get_param_index, get_input_pair_index, F2K, K2F, HAPropsSI, AbstractState_factory, AbstractState_free, AbstractState_set_fractions, AbstractState_update, AbstractState_keyed_output, AbstractState_output, AbstractState_update_and_common_out, AbstractState_update_and_5_out, AbstractState_set_binary_interaction_double
+export PropsSI, PhaseSI, get_global_param_string, get_parameter_information_string,get_fluid_param_string,set_reference_stateS, get_param_index, get_input_pair_index, F2K, K2F, HAPropsSI, AbstractState_factory, AbstractState_free, AbstractState_set_fractions, AbstractState_update, AbstractState_keyed_output, AbstractState_output, AbstractState_update_and_common_out, AbstractState_update_and_1_out, AbstractState_update_and_5_out, AbstractState_set_binary_interaction_double
 
 # Check the current Julia version to make this Julia 0.4 code compatible with older version
 if VERSION <= VersionNumber(0,4)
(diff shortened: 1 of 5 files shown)


JonWel:
What's happening with 2201119 ?
It has already be merged as 2315c27 , so why is it here again?

wrappers/Julia/CoolProp.jl
```diff
@@ -30,7 +30,7 @@ function PropsSI(FluidName::AbstractString, Output::AbstractString)
 end
 
 # CoolProp::PropsSI(const std::string &, const std::string &, double, const std::string &, double, const std::string&)
-function PropsSI(Output::AbstractString, Name1::AbstractString, Value1::Number, Name2::AbstractString, Value2::Number, Fluid::AbstractString)
+function PropsSI(Output::AbstractString, Name1::AbstractString, Value1::Real, Name2::AbstractString, Value2::Real, Fluid::AbstractString)
```
JonWel: Number include the possibility to have complex input.
Real don't, but allows to send float and integers (with all being converted to the required type for the c library by ccall).
Thus sending floats will generally be faster (no conversion needed), but this is still convenient to have the possibility to send integers.

wrappers/Julia/CoolProp.jl
```diff
@@ -220,6 +220,49 @@ function AbstractState_update(handle::Clong,input_pair::Clong,value1::Number,val
   end
   return nothing
 end
+function AbstractState_update(handle::Clong,input_pair::AbstractString,value1::Real,value2::Real)
+  AbstractState_update(handle::Clong,get_input_pair_index(input_pair),value1::Real,value2::Real)
+  return nothing
+end
```
JonWel: Overloading the function to be able to give the input pair in the form of a string (calling internally get_input_pair_index).
This allows to write code faster when only one update with this pair is made.

wrappers/Julia/CoolProp.jl
```diff
@@ -239,6 +282,9 @@ function AbstractState_keyed_output(handle::Clong, param::Clong)
   end
   return output
 end
+function AbstractState_output(handle::Clong, param::AbstractString)
+  return AbstractState_keyed_output(handle, get_param_index(param))
+end
```
JonWel: Overloading the function to be able to give the output name in the form of a string (calling internally get_param_index).
This allows to write code faster when only one output with this name is required.

wrappers/Julia/CoolProp.jl
```diff
@@ -264,6 +310,36 @@ function AbstractState_update_and_common_out(handle::Clong, input_pair::Clong, v
   end
   return nothing
 end
+function AbstractState_update_and_common_out(handle::Clong, input_pair::AbstractString, value1::Array{Real}, value2::Array{Real}, length::Real, T::Array{Real}, p::Array{Real}, rhomolar::Array{Real}, hmolar::Array{Real}, smolar::Array{Real})
+  AbstractState_update_and_common_out(handle, get_input_pair_index(input_pair), value1, value2, length, T, p, rhomolar, hmolar, smolar)
+  return nothing
+end
```
JonWel: Also overloading the functions with internal loop for easier usage.

PR #1135, opened 2016-06-23
Update CoolPropLib.def

JonWel:
Update CoolPropLib.def

And correct type specification for Array inputs in Julia wrapper.

### DIFF
--- a/wrappers/Julia/CoolProp.jl
+++ b/wrappers/Julia/CoolProp.jl
@@ -297,7 +297,7 @@ end
 # rhomolar The pointer to the array of molar density
 # hmolar The pointer to the array of molar enthalpy
 # smolar The pointer to the array of molar entropy
-function AbstractState_update_and_common_out(handle::Clong, input_pair::Clong, value1::Array{Real}, value2::Array{Real}, length::Real, T::Array{Real}, p::Array{Real}, rhomolar::Array{Real}, hmolar::Array{Real}, smolar::Array{Real})
+function AbstractState_update_and_common_out{R<:Real}(handle::Clong, input_pair::Clong, value1::Array{R}, value2::Array{R}, length::Real, T::Array{R}, p::Array{R}, rhomolar::Array{R}, hmolar::Array{R}, smolar::Array{R})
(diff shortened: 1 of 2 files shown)


wrappers/Julia/CoolProp.jl
```diff
@@ -297,7 +297,7 @@ end
 # rhomolar The pointer to the array of molar density
 # hmolar The pointer to the array of molar enthalpy
 # smolar The pointer to the array of molar entropy
-function AbstractState_update_and_common_out(handle::Clong, input_pair::Clong, value1::Array{Real}, value2::Array{Real}, length::Real, T::Array{Real}, p::Array{Real}, rhomolar::Array{Real}, hmolar::Array{Real}, smolar::Array{Real})
+function AbstractState_update_and_common_out{R<:Real}(handle::Clong, input_pair::Clong, value1::Array{R}, value2::Array{R}, length::Real, T::Array{R}, p::Array{R}, rhomolar::Array{R}, hmolar::Array{R}, smolar::Array{R})
```
JonWel: http://docs.julialang.org/en/release-0.4/manual/methods/#parametric-methods

Float64 <: Real => true
Array{Float64 } <: Array{Real} => false
Because of Julia's invariance

wrappers/Julia/CoolProp.jl
```diff
@@ -297,7 +297,7 @@ end
 # rhomolar The pointer to the array of molar density
 # hmolar The pointer to the array of molar enthalpy
 # smolar The pointer to the array of molar entropy
-function AbstractState_update_and_common_out{R<:Real}(handle::Clong, input_pair::Clong, value1::Array{R}, value2::Array{R}, length::Real, T::Array{R}, p::Array{R}, rhomolar::Array{R}, hmolar::Array{R}, smolar::Array{R})
+function AbstractState_update_and_common_out{R1<:Real,R2<:Real,F<:AbstractFloat}(handle::Clong, input_pair::Clong, value1::Array{R1}, value2::Array{R2}, length::Real, T::Array{F}, p::Array{F}, rhomolar::Array{F}, hmolar::Array{F}, smolar::Array{F})
```
JonWel: Outputs of type AbstractFloat should gives 32/64 bit compatibility out of the box.
The inputs can be arrays of Integers or Floats (32 or 64), to the choice. Best in term of performance is to use `Cdouble` as no conversion will be needed.

PR #1177, opened 2016-08-04
Volume translation to cubics, closes #1168

### LINKED ISSUE #1168: Volume translation for cubic
I would like to add a simple volume translation to the cubic backend.

To first start simple, I plan to make it constant (we could have temperature dependence etc later, but temperature dependence would not be fun for the derivatives), with a value set by imposing it from outside `set_volume_translation_parameter(c)` or `set_VT_parameter(c)`.
So the values will need to be computed from outside.
I'll add a `cm_term` in `GeneralisedCubic` (and change the cm to cm_term in the VPTR), this will just return the value from a private variable cm, but will allows to have more complex return later like temperature dependency.

I'm modifying the alphar (and derivatives) of the cubic to include this volume translation based on the analytic developments from [here](https://www.researchgate.net/publication/304186500_Helmholtz_Energy_Transformations_of_Common_Cubic_Equations_of_State_for_Use_with_Pure_Fluids_and_Mixtures).
Once done, the alphar from the VTPRCubic.h will not be needed any-more (this way, the translation will be possible with all cubic models, and the VTPR will have the analytic development for alphar).

JonWel:
Add volume translation to the cubics, VTPR included.

The volume translation is thus possible for all cubics.

All the residual non-dimensional Helmholtz energy derivatives analytic equations have been adapted to include the volume translation.


include/AbstractState.h
```diff
@@ -502,6 +502,8 @@ class AbstractState {
     virtual std::string get_binary_interaction_string(const std::string &CAS1, const std::string &CAS2, const std::string &parameter){ throw NotImplementedError("get_binary_interaction_string is not implemented for this backend"); };
     /// Apply a simple mixing rule (EXPERT USE ONLY!!!)
     virtual void apply_simple_mixing_rule(std::size_t i, std::size_t j, const std::string &model) { throw NotImplementedError("apply_simple_mixing_rule is not implemented for this backend"); };
+    // Set the volume translation parameter (for cubic)
+    virtual void set_volume_translation(const double value) { throw ValueError("set_volume_translation only defined for cubic backends"); };
```
ibell: I think it would be better to create a more generalized function called set_fluid_parameter_double(size_t i, std::string parameter, double value).  This function could be used for lots of different things...

include/CoolPropLib.h
```diff
@@ -406,6 +412,17 @@
...
+    /**
+    * @brief Set the volume translation parameter (for cubic)
+    * @param handle The integer handle for the state class stored in memory
+    * @param value the value of the volume translation parameter
+    * @param errcode The errorcode that is returned (0 = no error, !0 = error)
+    * @param message_buffer A buffer for the error code
+    * @param buffer_length The length of the buffer for the error code
+    * @return
+    */
+    EXPORT_CODE void CONVENTION  AbstractState_set_volume_translation(const long handle, const double value, long *errcode, char *message_buffer, const long buffer_length);
```
ibell: Same comment as in AbstractState.h

include/AbstractState.h
JonWel: Ok, I'll do something similar to [set_binary_interaction_double](https://github.com/CoolProp/CoolProp/blob/master/src/Backends/Cubics/CubicBackend.cpp#L455).

JonWel: @ibell I only changed it in the VTPR branch (as it's the one I'm doing the PR from).
So the master branch also has the volume translation, but still the old function to set it.

ibell:
This PR is currently dead because of my rebase.  Sorry!

PR #1210, opened 2016-08-26
Use aii_term and b0_ii from generalised cubic

JonWel:
Making use of their feature (aii_term has tau derivative), and only do numeric derivatives for the gE_R term.


JonWel:
Mh, seems I made the change just some seconds to late. I'll do a cleaner PR.

PR #1214, opened 2016-08-27
Refactor exceptions in CoolPropLib.cpp close #1200

### LINKED ISSUE #1200: Refactor CoolPropLib.cpp
I wonder if we could apply this logic (http://codereview.stackexchange.com/a/2517) to make the CoolPropLib.cpp file easier to maintain:

``` c++
int rc = 0;
try{
// statements that can throw
}catch(...){
    HandleException(rc, "Extra info to append to logged exception message");
}



void HandleException(int rc, std::string msg)
{
    try{
        throw;
    }catch(Poco::Exception &e){
        // ...
    }catch( const std::exception &e){
        // ...
    }
}
```

JonWel:
There are also some exceptions that do not use this new `HandleException` function yet as they don't use the errcode and message buffer (they just set the error string):
```C++
catch (std::exception &e){ CoolProp::set_error_string(e.what()); }
    catch (...){ CoolProp::set_error_string("Undefined error"); }
```
Should we do a second function to handle that case?

Some functions like `AbstractState_update_and_common_out` do not have the `CoolPropBaseError` in the previous code. This will be also tested now, but should not be a problem (the `HandleError` is triggered first).

### DIFF
--- a/src/CoolPropLib.cpp
+++ b/src/CoolPropLib.cpp
@@ -26,6 +26,35 @@ void str2buf(const std::string& str, char * buf, int n)
   else
     throw CoolProp::ValueError("Buffer size is too small");
 }
+void HandleException(long *errcode, char *message_buffer, const long buffer_length)
+{
+    try{
+        throw;
+    }
+    catch (CoolProp::HandleError &e) {
+        std::string errmsg = std::string("HandleError: ") + e.what();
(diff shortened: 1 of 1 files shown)


ibell:
Excellent!  Thank you very much.  Nice work on this one.  It makes adding functionality to CoolPropLib.cpp much less painful than before.

PR #1257, opened 2016-10-10
Update iPhone compilation docs

babaksamareh:
As we discussed before


ibell:
You did just fine!  Thanks!

PR #1258, opened 2016-10-11
Solves a bug with cubic and mixtures

JonWel:
The transient_pure_state has no mole fractions when created, this was triggering a CoolProp: `Mole fractions must be set` error.

I guess it's because that transient state is supposed to be pure, but this fix allows to get the answer.
Is there something more elegant to do?


ibell:
Sounds elegant enough to me!

PR #1266, opened 2016-10-14
Small fixes for VTPR

JonWel:


src/Backends/Cubics/VTPRBackend.h
```diff
@@ -45,6 +45,8 @@ class VTPRBackend : public PengRobinsonBackend  {
         const UNIFAQLibrary::UNIFAQParameterLibrary & lib = LoadLibrary();
         for (std::size_t i = 0; i < fluid_identifiers.size(); ++i){
             UNIFAQLibrary::Component comp = lib.get_component("name", fluid_identifiers[i]);
+            // Taking the mole fraction from the cubic library. TODO: Need a better solution
```
ibell: This won't work for fluids that are not already in the the cubics library.

JonWel: Of course, the idea was just to have a temporary fix to still have it for the fluids who are in the cubics too.
When the molar mass would be available directly in the VTPR library, this line would have to be replaced with something to take the molar mass from the VTPR library.

PR #1269, opened 2016-10-15
SatL and SatV of type VTPR too

JonWel:
The copy was previously creating PengRobinson instances for SatV and SatL, thus, for mixtures, calls like "QT_INPUTS" where giving identical results as PR.


ibell:
Aha, this is the reason behind the bug you reported on the paper, right?

JonWel:
Unfortunately not.
On my last commit in that repository, I already had that patch.
After applying this, I got back the results of the 13 october (whereas just before this, the became identical to PR).

PR #1271, opened 2016-10-17
Remove possible division by 0, closes #1270

### LINKED ISSUE #1270: Possible NaN in successive_substitution
[This line](https://github.com/CoolProp/CoolProp/blob/master/src/Backends/Helmholtz/VLERoutines.cpp#L1176) has a division by `df`.
I felt on a specific case where this `df` is `0`, thus `change` becomes `NaN`, and a bit later is added to `p` (or `T`), which is of course not an ideal output value (in my case, `f` is also null, thus it goes out of the loop as converged).

The specific case where this happened to me is in fact a pure fluid case (molar fraction = [0,1]), thus maybe this issue can only happen in that case.

If `df` can only happen to be null when it's a pure fluid, maybe this could be interesting to set a switch to use the pure fluid routine in that case (should be faster than the mixture routine and avoid the issue).

JonWel:


src/Backends/Helmholtz/VLERoutines.cpp
```diff
@@ -1173,7 +1173,16 @@ void SaturationSolvers::successive_substitution(HelmholtzEOSMixtureBackend &HEOS
             df += dfdK*(deriv_liq-deriv_vap);
         }
         
-        change = -f/df;
+        if (std::abs(df) <= 1e-12) { // To avoid dividing by 0
+            change = -f;
+            if (std::abs(f) > 1e-12)
+            {
+                set_warning_string("df verry small in successive_substitution but f do not seems converged");
```
ibell: verry -> very

src/Backends/Helmholtz/VLERoutines.cpp
```diff
@@ -1173,7 +1173,16 @@ void SaturationSolvers::successive_substitution(HelmholtzEOSMixtureBackend &HEOS
             df += dfdK*(deriv_liq-deriv_vap);
         }
         
-        change = -f/df;
+        if (std::abs(df) <= 1e-12) { // To avoid dividing by 0
+            change = -f;
```
JonWel: I was not fully sure about what to do with this case.
I don't have the understanding on the physics involved in these variables, so I don't really know which value they are expected to have.

So if f is in tolerance, we finish the loop (the the K that have just been computed is used), and if not, we error out.

I don't think I ever experienced the case where df would be small without f being converged.
So I'll change to error out when it happens with an nicer message (`very` with only one `r`and value of  `f` and `df` displayed).
That way, if it happens, we will have the informations to know what to do, and if it doesn't happen, the problem is solved. (case where df where too small where already quite rare).

src/Backends/Helmholtz/VLERoutines.cpp
```diff
@@ -1173,7 +1173,18 @@ void SaturationSolvers::successive_substitution(HelmholtzEOSMixtureBackend &HEOS
             df += dfdK*(deriv_liq-deriv_vap);
         }
         
-        change = -f/df;
+        if (std::abs(df) <= 1e-12) { // To avoid dividing by 0
+            if (std::abs(f) <= 1e-12)
++            {
++                change = -f; // Should be converged. f <= e-12, so change will have nearly no impact.
```
ibell: I would make change zero in this case.  It will fall through and exit the loop.

src/Backends/Helmholtz/VLERoutines.cpp
```diff
@@ -1173,7 +1173,18 @@ void SaturationSolvers::successive_substitution(HelmholtzEOSMixtureBackend &HEOS
             df += dfdK*(deriv_liq-deriv_vap);
         }
         
-        change = -f/df;
+        if (std::abs(df) <= 1e-12) { // To avoid dividing by 0
+            if (std::abs(f) <= 1e-12)
```
JonWel: [The loop tolerance](https://github.com/JonWel/CoolProp-f/blob/patch-4/src/Backends/Helmholtz/VLERoutines.cpp#L1212) is hard coded, I just took the same value.
I did not see any [function](https://github.com/JonWel/CoolProp-f/blob/patch-4/src/Backends/Helmholtz/VLERoutines.cpp#L1098) tolerance.

src/Backends/Helmholtz/VLERoutines.cpp
JonWel: The loop tolerance being hard coded as [std::abs(f) > 1e-12](https://github.com/JonWel/CoolProp-f/blob/patch-4/src/Backends/Helmholtz/VLERoutines.cpp#L1212) this will just finish the loop and exit.
I thought about putting a break, but in that case, the updated K would not be used, so the update just after the loop be just useless (even if K should not be extremely different than the previous one).

PR #1272, opened 2016-10-17
Solves a bug in VTPR

JonWel:
The value of lnGammag was not cleared, this was creating a sort of caching issue.
(@ibell this solves the issue I was had in Paper1, I finally got it)


ibell:
Ok, glad you were able to track this down finally!

PR #1302, opened 2016-10-29
Export cubic's alpha functions

JonWel:
Allows to set coefficients for Mathias-Copemann and TWU from the shared library.
Functionality also added to the Julia Wrapper

### DIFF
--- a/include/AbstractState.h
+++ b/include/AbstractState.h
@@ -515,9 +515,9 @@ public:
     /// Apply a simple mixing rule (EXPERT USE ONLY!!!)
     virtual void apply_simple_mixing_rule(std::size_t i, std::size_t j, const std::string &model) { throw NotImplementedError("apply_simple_mixing_rule is not implemented for this backend"); };
     /// Set the cubic alpha function's constants:
-    virtual void set_cubic_alpha_C(const size_t i, const std::string parameter, const double c1, const double c2, const double c3) { throw ValueError("set_cubic_alpha_C only defined for cubic backends"); };
+    virtual void set_cubic_alpha_C(const size_t i, const std::string &parameter, const double c1, const double c2, const double c3) { throw ValueError("set_cubic_alpha_C only defined for cubic backends"); };
     /// Set fluid parameter (currently the volume translation parameter for cubic)
-	virtual void set_fluid_parameter_double(const size_t i, const std::string parameter, const double value) { throw ValueError("set_fluid_parameter_double only defined for cubic backends"); };
+	virtual void set_fluid_parameter_double(const size_t i, const std::string &parameter, const double value) { throw ValueError("set_fluid_parameter_double only defined for cubic backends"); };
 
     /// Clear all the cached values
     virtual bool clear();
--- a/include/AbstractState.h
+++ b/include/AbstractState.h
@@ -514,7 +514,9 @@ public:
     virtual std::string get_binary_interaction_string(const std::string &CAS1, const std::string &CAS2, const std::string &parameter){ throw NotImplementedError("get_binary_interaction_string is not implemented for this backend"); };
     /// Apply a simple mixing rule (EXPERT USE ONLY!!!)
     virtual void apply_simple_mixing_rule(std::size_t i, std::size_t j, const std::string &model) { throw NotImplementedError("apply_simple_mixing_rule is not implemented for this backend"); };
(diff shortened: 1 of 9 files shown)


src/Backends/Cubics/GeneralizedCubic.h
```diff
@@ -123,10 +123,18 @@ class AbstractCubic
     void set_C_MC(double c1, double c2, double c3){
         alpha[0].reset(new MathiasCopemanAlphaFunction(a0_ii(0),c1,c2,c3,T_r/Tc[0]));
     }
+    /// Set the three Mathias-Copeman constants in one shot for the component i of a mixturee
```
ibell: Maybe just replace the other set_C_MC function?  Seems a bit redundant having two functions basically doing exactly the same thing.  Or you make the old one call the new one with index of 0

src/Backends/Cubics/GeneralizedCubic.h
```diff
@@ -123,10 +123,18 @@ class AbstractCubic
...
+    void set_C_MC(std::size_t i,double c1, double c2, double c3){
+        alpha[i].reset(new MathiasCopemanAlphaFunction(a0_ii(i),c1,c2,c3,T_r/Tc[i]));
+    }
     /// Set the three Twu constants in one shot for a pure fluid
     void set_C_Twu(double L, double M, double N){
         alpha[0].reset(new TwuAlphaFunction(a0_ii(0),L,M,N,T_r/Tc[0]));
     }
+    /// Set the three Twu constants in one shot for the component i of a mixture
+    void set_C_Twu(std::size_t i,double L, double M, double N){
```
ibell: Same thing here

include/AbstractState.h
```diff
@@ -514,7 +514,9 @@ class AbstractState {
     virtual std::string get_binary_interaction_string(const std::string &CAS1, const std::string &CAS2, const std::string &parameter){ throw NotImplementedError("get_binary_interaction_string is not implemented for this backend"); };
     /// Apply a simple mixing rule (EXPERT USE ONLY!!!)
     virtual void apply_simple_mixing_rule(std::size_t i, std::size_t j, const std::string &model) { throw NotImplementedError("apply_simple_mixing_rule is not implemented for this backend"); };
-    // Set fluid parameter (currently the volume translation parameter for cubic)
+    /// Set the cubic alpha function's constants:
+    virtual void set_cubic_alpha_C(const size_t i, const std::string parameter, const double c1, const double c2, const double c3) { throw ValueError("set_cubic_alpha_C only defined for cubic backends"); };
```
ibell: This should pass the parameter by reference

src/CoolPropLib.cpp
```diff
@@ -591,7 +602,7 @@ EXPORT_CODE void CONVENTION  AbstractState_set_fluid_parameter_double(const long
     *errcode = 0;
     try {
         shared_ptr<CoolProp::AbstractState> &AS = handle_manager.get(handle);
-        AS->set_fluid_parameter_double(static_cast<std::size_t>(i), parameter, value);
+        AS->set_fluid_parameter_double(static_cast<std::size_t>(i), std::string(parameter), value);
```
ibell: Since the parameter will be passed by reference, this cast is not needed

JonWel:
Should be better now.

Is the name "set_cubic_alpha_C" good enough?
Maybe "set_cubic_alpha_const" would be  more explicit?

src/Backends/Cubics/GeneralizedCubic.h
```diff
@@ -119,13 +119,13 @@ class AbstractCubic
     double get_Delta_2(){ return Delta_2; }
     /// Read-only accessor for value of R_u (universal gas constant)
     double get_R_u(){ return R_u; }
-    /// Set the three Mathias-Copeman constants in one shot for a pure fluid
-    void set_C_MC(double c1, double c2, double c3){
-        alpha[0].reset(new MathiasCopemanAlphaFunction(a0_ii(0),c1,c2,c3,T_r/Tc[0]));
+    /// Set the three Mathias-Copeman constants in one shot for the component i of a mixturee
```
ibell: mixturee -> mixture

This will break my code, but I am fine with that since I doubt anyone else is working so deep inside the internals of CoolProp.

ibell:
Looks good, thanks!

PR #1308, opened 2016-11-02
Add Trivial Parameter calls to IF97 Backend

henningjp:
Only implemented the trivial parameters that are defined in IAPWS Guides or the IAPWS-IF97 release.  The only exception is the acentric factor which is calculated from P @ Tr=0.7 and Pcrit.  Had to add an extra virtual function to AbstractState.h (calc_rhomass_critical) to allow for mass based IF97 calculations.  This function will be ignored by the other backends.

Function verification (through Mathcad) is shown in attached.
trivial

Slight adjustment to .ignore also to allow for other build directories starting with "build".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,7 +37,7 @@
 /include/cubic_fluids_schema_JSON.h
 /include/version.h
 /include/catch.hpp
-/build/
+/build*/
 /dev/hashes.json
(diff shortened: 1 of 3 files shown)


ibell:
Looks good! Merging.

Issue #1310, opened 2016-11-02
Full analytical derivatives in VTPR


JonWel:
This will not be easy and it will take times.

Right now, as time is limited, I'm thinking about doing one level at a time.

JonWel:
First level from numerical derivatives of gE_R to a numerical derivative of ln_gamma_R:
![capture du 2016-11-02 17-01-09](https://cloud.githubusercontent.com/assets/11057909/19936546/121ec634-a11e-11e6-8d29-2131d86fee6d.png)

ibell:
This is awesome.  I know it is going to be painful.  I'd be interested to
help; I haven't done crazy derivatives in a while...

JonWel:
The most painful part will be the one for lnGamma, it has a fraction (and a ln which becomes a fraction after a derivative). Fractions gives big amount of terms: https://www.wolframalpha.com/input/?i=d%5E4(f(x)%2Fg(x))%2Fdx%5E4.
The remaining is still "nice", as you can see.

JonWel:
After #1312, I think it will be required to change the structure of the code in order to introduce the other derivatives (we don't want to compute all derivatives when needing only the first one).

Basically, the `set_mole_fractions` remains, but the `set_temperature`, will change as the derivatives will need to be applied through it and we need to compute only the derivative of interest for a given call.
And if possible to cache as much as we can between calls (for instance, psi will be required several times for all the derivatives, and the derivatives of `ln_gamma_R` needs derivatives of different orders, so that may be a good idea to store a matrix of psi_mn in order to compute them only once per given temperature. This makes one psi_mn matrix for each fluids, so this will take a bit more memory).

I noticed `ln_Gamma_C` is computed at each `set_mole_fractions`, but it's only necessary in computing the `activity_coefficient`.
A first easy step may be to put it in a stand-alone function or in the `activity_coefficient` code, so this will slightly speed the other calls.
Be aware, if you are using `activity_coefficient` in your codes, that I'm changing the input arguments, in order to have it compatible with the new way of calling `ln_gamma_R`.

JonWel:
I noticed `ln_Gamma_C` is computed at each `set_mole_fractions`, but it's only necessary in computing the `activity_coefficient`.

Shouldn't we put it in a stand-alone function or in the `activity_coefficient` code, so this will slightly speed the other calls?

The `activity_coefficient` function depends on the molar fraction and the temperature but currently relies on them to be set before.
[...]

PR #1312, opened 2016-11-02
First step toward #1310

JonWel:
Now `gE_R` has analytical derivatives.
The numerical derivatives are pushed into `ln_gamma_R`


ibell:
Have you tested these derivatives?

JonWel:
Now yes,
worth it as the catch routine for the UNIFAQ was not using the updated functions.
Derivatives did not trigger any messages in the test file.

PR #1316, opened 2016-11-04
Added configuration options for MSVCRT linking, changed the output di…

jowr:
…rectory of the static libraries, added a "d" to the debugging libraries on Windows

Do these changes cause any problems? I needed more flexibility for Windows builds...


ibell:
Nothing jumps out at me.  Let's merge and see :)

ibell:
Well, there's your answer: http://www.coolprop.dreamhosters.com:8010/builders/Python-binaries-windows-32bit-py27/builds/1659 :(

ibell:
I'm still frustrated with this change, because now any time that I want to integrate a CoolProp static library into another EXE or dynamic library, I need to set a flag that turns the MSVC runtime linkage to dynamic linkage. I guess its not the end of the world, but perhaps for static library at least, the default should be to leave the runtime linkage as is? For the other ones, that are terminal objects (shared libraries and so on), the logic you integrated makes total sense.

jowr:
I hope 3905f38 fixes this once and for all. By the way, sorry for the whitespace noise - I used a new editor that did not like the old format...

ibell:
Yes, I think we are good.

PR #1323, opened 2016-11-07
Added PQ and QT Input Pairs to provide Saturation Values to IF97 Backend

henningjp:
- [x] Added PQ and QT Input Pair Facility for all Thermo props.  
- [x] Implemented Temperature/Pressure saturation curve (both ways).  
- [x] Handle mixed Vapor/Liquid inside saturation dome for Dmass, Hmass, Smass, Umass by linear mixing when 0 < Q < 1.  However, Cpmass, Cvmass, and speed of sound throw and exception as these values are not defined in the two phase region.  
- [x] Also implemented Molar values by multiplying (or dividing) by molar_mass().
- [x] Touched up some other minor nits.

Partial verification (through Mathcad wrapper) shown below.  You may want to verify that Cpmass, Cvmass, and speed of sound are properly throwing the exception messages (e.g. in Python) as I can't see them in Mathcad yet.
saturation verif1
saturation verif2
saturation verif3
saturation verif4
saturation verif5

### DIFF
--- a/src/Backends/IF97/IF97Backend.h
+++ b/src/Backends/IF97/IF97Backend.h
@@ -14,9 +14,9 @@ class IF97Backend : public AbstractState  {
 
 protected:
 
-	/// Additional cached elements used only in this backend since the "normal"
+    /// Additional cached elements used only in this backend since the "normal"
     /// backends use only molar units while IF97 uses mass-based units
-	CachedElement  _cpmass, _cvmass, _hmass, _rhomass, _smass, _umass;
+    CachedElement  _cpmass, _cvmass, _hmass, _rhomass, _smass, _umass;
 
 public:
     /// The name of the backend being used
@@ -35,12 +35,13 @@ public:
     bool clear() {
     // Reset all instances of CachedElement and overwrite
     // the internal double values with -_HUGE
+    // Default phase condition is no phase imposed
+    // IF97 will make phase/region determination
         this->_rhomolar = -_HUGE;
         this->_T = -_HUGE;
         this->_p = -_HUGE;
         this->_Q = -_HUGE;
-        this->_phase = iphase_not_imposed; // Default condition is no phse imposed
-
+        this->_phase = iphase_not_imposed; 
         return true;
     };
 
@@ -62,28 +63,14 @@ public:
             case PQ_INPUTS: 
                 _p = value1; 
                 _Q = value2; 
-                _T = IF97::Tsat97(_p); 
-                if (std::abs(_Q) < 1e-10){
-                    _phase = iphase_liquid; // bubble point
-                }
-                else if (std::abs(value2-1) < 1e-10){
-                    _phase = iphase_gas; // dew point
-                }
-                else
-                    _phase = iphase_twophase;
+                _T = IF97::Tsat97(_p);  // ...will throw exception if _P not on saturation curve
+                _phase = iphase_twophase;
                 break;
             case QT_INPUTS: 
                 _Q = value1; 
                 _T = value2; 
-                _p = IF97::psat97(_T); 
-                if (std::abs(_Q) < 1e-10){
-                    _phase = iphase_liquid; // bubble point
-                }
-                else if (std::abs(value2-1) < 1e-10){
-                    _phase = iphase_gas; // dew point
-                }
-                else
-                    _phase = iphase_twophase;
+                _p = IF97::psat97(_T);  // ...will throw exception if _P not on saturation curve
+                _phase = iphase_twophase;
                 break;
             default:
                 throw ValueError("Bad input_pair");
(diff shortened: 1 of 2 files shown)


src/Backends/IF97/IF97Backend.h
```diff
@@ -59,31 +97,115 @@ class IF97Backend : public AbstractState  {
...
+                break;
+            case iphase_gas:    
+                return calc_Vapor(iCalc); 
+                break;
+            case iphase_twophase:
+                switch(iCalc){ 
+                case iDmass:
+                    return 1.0/(_Q/calc_Vapor(iCalc) + (1.0-_Q)/calc_Liquid(iCalc)); break;
+                case iCpmass:
+                    throw NotImplementedError(format("Isobaric Specific Heat not implemented in two phase region")); break;
```
ibell: This *is* defined if Q == 0 or if Q == 1

src/Backends/IF97/IF97Backend.h
```diff
@@ -59,31 +97,115 @@ class IF97Backend : public AbstractState  {
...
+            case iphase_liquid: 
+                return calc_Liquid(iCalc); 
+                break;
+            case iphase_gas:    
+                return calc_Vapor(iCalc); 
+                break;
+            case iphase_twophase:
+                switch(iCalc){ 
+                case iDmass:
+                    return 1.0/(_Q/calc_Vapor(iCalc) + (1.0-_Q)/calc_Liquid(iCalc)); break;
```
ibell: For clarity, might want to put iDmass in place of iCalc here.

src/Backends/IF97/IF97Backend.h
```diff
@@ -31,24 +31,62 @@ class IF97Backend : public AbstractState  {
...
+                }
+                else
+                    _phase = iphase_twophase;
+                break;
+            case QT_INPUTS: 
+                _Q = value1; 
+                _T = value2; 
+                _p = IF97::psat97(_T); 
+                if (std::abs(_Q) < 1e-10){
+                    _phase = iphase_liquid; // bubble point
```
ibell: Technically bubble-point and dew-point are two-phase, not liquid.

ibell:
I really like how you show your work for the unit conversions in the comments.

src/Backends/IF97/IF97Backend.h
henningjp: Technically yes, but I'm trying to make only one saturated function call and not two in the calc_Flash function and need the flag.  If at the bubble or dew point, I only want to call satL() or satV(), not both to minimize the real computations.  Not a big deal normally unless you're calculating a million points; then it saves 50% of the computational time.  This contributes to your third comment.  I could set, the phase to iphase_two_phase and do the checks for Q ==0 or Q == 1 in the calc_Flash routine.  It just looks cleaner this way AND its for a PQ/QT input pair so we should only be in the vapor dome anyway.

I can move the logic if you think the semantics will create confusion.  IF97 doesn't use these flags internally so it won't care.  LMK

src/Backends/IF97/IF97Backend.h
henningjp: Yes, those two cases are handled just above that logic for (_phase == iphase_liquid) and (_phase == iphase_vapor), where I only want to call one function instead of two to save time in the spirit of IF97.  This instance is only called when Q is _between_ 0 and 1 (_phase == iphase_two_phase); in which case these quantities (Cp, Cv, Speed of Sound) are undefined.  (Ignore my external e-mail on this, by the way.)

src/Backends/IF97/IF97Backend.h
henningjp: An additional thought from my morning commute...

IF97 needs to be able to stand-alone outside of CoolProp.   This means that IF97 needs to be able to make the phase determination as it determines the IF97Region based on the state.  We shouldn't be doing this within the IF97 Backend, So there are only three states that need to be considered in the IF97 Backend:
- Non-Saturation (iphase_not_imposed): Let IF97 make it's phase determination  
- Q = 0 or 1 (iphase_liquid or iphase_vapor): Call only one IF97 Saturation function  
- 0 < Q < 1 (iphase_two_phase): Call two IF97 Saturation functions and calculate the two phase property (or not).

The 'phase' output variable, if implemented would use the IF97 Region Determination function directly, not just parrot back the '_phase' variable.  Outside the phase envelope, the '_phase' is always iphase_not_imposed to let IF97 do its job. However, if this will cause a conflict in other parts of CoolProp, we can do it another way.

I can always add copious comments as well so that future developers won't be confused by the use of iphase_liquid and iphase_vapor only on the saturation boundary.

henningjp: Done.  Moved ( Q == 1)/(Q == 0) logic.

src/Backends/IF97/IF97Backend.h
henningjp: Done.

PR #1351, opened 2016-11-27
replace bytestring (deprecated) with unsafe_string

DANA-Laboratory:


ibell:
Is this change backwards compatible?

vimalaad:
I added a second commit to keep it compatible with older Julia, but I think it is not necessary to hard code this type of version  checking, normally interoperability between older and newer versions of the Julia language done by [Compat.jl](https://github.com/JuliaLang/Compat.jl), so +1 to remove version checking and start using Compat.jl

ibell:
+1 on using Compat.jl from my quick perusal

ibell:
Okie dokie, looks good.  Hopefully we won't get complaints from other julia users.

Issue #1355, opened 2016-11-29
SMath Wrapper refactoring


dcprojects:
- [x] upgrade to SMath.Math.IPluginLowLevelEvaluationFast interface;

- [x] add Test folder to store test files;

- [x] code indent;

- [x] review of descriptions in Dynamic Assistant;

- [x] README.md listing available functions;

ibell:
Are you going to do these things?  A PR would of course be appreciated.

dcprojects:
Yes, I'll make a pull request in days.

ibell:
Hmm...  Seems your changes are not backwards compatible: http://www.coolprop.dreamhosters.com:8010/builders/SMath/builds/1058/steps/shell_1/logs/stdio

ibell:
Can you perhaps figure out how to make them backwards compatible?  I can also update my SMath Studio on the build server...

dcprojects:
Currently this code should compile in the latest 2 major relases of SMath Studio and more than 20 minor builds before them (precisely I can compile it from SS 0.98.5973); my suggestion is to update the program on the Build Server.

ibell:
I updated SMath on the build server, works now.

PR #1365, opened 2016-12-01
SMath Wrapper refactoring

dcprojects:
PR for #1355
@mikekaganski


ibell:
Hey @mikekaganski can you do a code review on this?

wrappers/SMath/coolprop_wrapper/Functions/CoolProp_HAProps.cs
```diff
@@ -3,136 +3,136 @@
...
-    }
+            return new TermInfo(inf.Text,
+                                inf.Type,
+                                funcInfo,
+                                FunctionSections.Unknown,
+                                true,
+                                argsInfos);
+        }
 
-    bool IFunction.ExpressionEvaluation(Term root, Term[][] args, ref SMath.Math.Store context, ref Term[] result)
```
mikekaganski: Is it OK that we pass context by value? Won't be here substantial overhead?

dcprojects: not sure if this is the only reason behind this change, but context is a reference type, thus not using ref should be an issue only if one wants to initialize a new instance (that shouldn't happens plugin-side);

mikekaganski: Thank you! I'm fine with this.

ibell:
Thanks guys

Issue #1368, opened 2016-12-05
segmentation fault when calling get_global_param_string("REFPROP_version") from Python


bremme:
I wanted to get the available REFPROP version by calling `get_global_param_string("REFPROP_version")`. When REFPROP is installed this return the installed version (e.g. 9.1) as expected. But when REFPROP is not installed I get a `segmentation fault`.  I'm using the CoolProp Python wrapper (version 6.1.0) which I installed using pip.

The python code which gives me the `segmentation fault` is:

```python
import CoolProp.CoolProp as cp

cp.get_global_param_string("REFPROP_version")
```

The full output I get in my shell is:

```shell
Good news: It is possible to use REFPROP on your system! However, the library 
could not be loaded. Please make sure that REFPROP is available on your system.

Neither found in current location nor found in system PATH.
If you already obtained a copy of REFPROP from http://www.nist.gov/srd/, 
add location of REFPROP to the PATH environment variable or your library path.

In case you do not use Windows, have a look at https://github.com/jowr/librefprop.so 
to find instructions on how to compile your own version of the REFPROP library.

ALTERNATIVE_REFPROP_PATH: 
[1]    6491 segmentation fault  python test.py
```

I quickly tested if it would make any difference if I do set the `ALTERNATIVE_REFPROP_PATH`, but it didn't, I still got the `segmentation fault`.

Let me know if there is anything more I can do to help solve this issue.

ibell:
That will be a one-liner to fix, basically it should check if the SETUPdll
function is a null pointer, and if it is, not evaluate it.

bremme:
I have not really looked at the `C++` CoolProp sources before but I guess you mean to check for a nullptr just before line 244 of REFPROPMixtureBackend.cpp:

```c++
SETUPdll(&N, fluids, hmx, default_reference_state,
            &ierr, herr,
            10000, // Length of component_string (see PASS_FTN.for from REFPROP)
            refpropcharlength, // Length of path_HMX_BNC
            lengthofreference, // Length of reference
            errormessagelength // Length of error message
            );
```

Do you want me to make a pull request for this, or perhaps its faster if you or one of the other main developers fixes this small bug?

ibell:
I can fix it, but a PR would be awesome.  Just throw a ValueError with the
right comment if SETUPdll == NULL

PR #1369, opened 2016-12-05
Return 'n/a'  REFPROP version when not loaded or supported.

bremme:
Create a pull request to solve issue #1368. When the `REFPROPMixtureBackend::version()` is called, the method will return `n\a` when `REFPROPMixtureBackend::REFPROP_supported()` returns `false`. This should prevent the segmentation fault when getting the REFPROP version if REFPROP is not installed. What do you think? Is returning 'n/a' a good solution or do you prefer to raise a ValueError when `SETUPdll == NULL`? My reasoning behind this implementation is that a complete check if REFPROP is supported and safe to load was already implemented in `REFPROP_supported()`, so it would make sense to just trust the value it returns.

P.S. I noticed that my editor removed some trailing white spaces I hope this is not a problem?

### DIFF
--- a/src/Backends/REFPROP/REFPROPMixtureBackend.cpp
+++ b/src/Backends/REFPROP/REFPROPMixtureBackend.cpp
@@ -236,7 +236,9 @@ std::string REFPROPMixtureBackend::version(){
     long N = -1;
     long ierr = 0;
     char fluids[10000] = "", hmx[] = "HMX.BNC", default_reference_state[] = "DEF", herr[255] = "";
-    REFPROPMixtureBackend::REFPROP_supported();
+    if (!REFPROP_supported()) {
+      return "n/a";
+    };
     // Pad the version string with NULL characters
     for (int i = 0; i < 255; ++i){
         herr[i] = '\0';


ibell:
Can you perhaps use a different editor and rollback the whitespace changes?  It is much nicer to see clean diffs, I'm lost in all the whitespace changes...

bremme:
No problem, I will do that! I guess I'm just so used to not having trailing white spaces that I didn't even knew my editor removed them by default. I did found the setting that caused is though: 
image

Will turn it off and make a new commit, sorry for the hassle.

ibell:
Nice!  Is that atom you are using by the way?  Also, if you wanted to provide a PR removing all the trailing whitespace, I'd be good with that!

bremme:
You're right that's the Atom text editor. Been using it for some months now and overal its a nice experience. Do you mean to create a PR for just the REFPROPMixtureBackend.cpp file or a larger part of the code base? Are there some guidelines on contributions to the project? I guess this would be a good place to define if the project favors/requires white spaces at the end of a line or after a line break for example.

thorade:
Here is a github pro tip you might already know: Append `?w=1` to a commit URL to hide whitespace changes. Compare these two:
- https://github.com/CoolProp/CoolProp/pull/1369/commits/abd916ac598a164ee9ffb1625ca031f488b91fc3
- https://github.com/CoolProp/CoolProp/pull/1369/commits/abd916ac598a164ee9ffb1625ca031f488b91fc3?w=1

As seen on https://github.com/tiimgreen/github-cheat-sheet/blob/master/README.md

bremme:
@thorade Thanks for the tip!

ibell:
I didn't know about that either!  Thanks!

henningjp:
This might be a useful tip on the Wiki.

ibell:
@henningjp - you are welcome to got for it, sorry I haven't been looking into the wiki, been busy in my non-CoolProp life

Issue #1293, opened 2016-10-26
CoolProp.jl as a registered Julia package


simulkade:
I added a few lines to your coolprop.jl code to make the installation easier. The code is here: https://github.com/simulkade/CoolProp.jl
Right now, the installation downloads all the libraries in the 'deps' folder irrespective of the platform. I can improve it by adding direct links to the latest libcoolprop library for different platforms. Is there any direct link to those libraries? I could not use the sourcefoge links, since they are not direct.

JonWel:
This would in deed be convenient to have it as a registered package.

Note that the julia wrapper file you are using is not the latest in date (I added some functions) and that I still did not make the required changes for Julia 0.5 (it works with 0.5, but uses the deprecated `bytestring` command. I'll have to check how to keep compatibility with 0.4 while removing this warnings).

I like the idea to place the library in a `deps` folder next to the `src` one. I think we can make this change in the wrapper here.

I don't know about the possibility for direct links but it may be possible to compile CoolProp when installing the package (a bit longer thou).

For Windows, it may also be possible to add the julia wrapper to CoolProp's windows installer.

simulkade:
Thanks for your reply. I only tested the package for some very simple calculations in Julia 0.5 and it worked fine. Perhaps I can add more to the test function.
I was thinking of adding a `build.jl` file to the `deps` folder to download/build the latest coolprop library. I give it a try this weekend on Linux.
I think for windows, it is better to install the `libcoolprop` library through the windows installer, and install the Julia wrapper via the usual `Pkg.add("CoolProp")`

DANA-Laboratory:
I run this test for `get_global_param_string` https://github.com/CoolProp/CoolProp/blob/master/src/CoolProp.cpp#L850
```
for param in ["version", "gitrevision", "errstring", "warnstring", "FluidsList", "incompressible_list_pure", "incompressible_list_solution", "mixture_binary_pairs_list", "parameter_list", "predefined_mixtures", "HOME", "cubic_fluids_schema", "invalid"]
  println(param * " = " * get_global_param_string(param))
end
```
the result is 

> version = 6.1.0
> gitrevision = 7864c2614ce5a0ef972386ee7f39859f0d3f8038
> errstring =
> warnstring =
> FluidsList =
> incompressible_list_pure =
> incompressible_list_solution =
> mixture_binary_pairs_list =
> parameter_list =
> predefined_mixtures =
> HOME = C:\Users\AliReza
[...]

PR #1384, opened 2017-01-06
Update Julia wrapper, ref #1293

DANA-Laboratory:
Hi, CoolProp guys.
Before actually merge this pr, I want to share my works on Julia wrapper, and wait to get your advises.
As you can [see](https://github.com/DANA-Laboratory/CoolProp/commit/6141dcf223bea3142a0efdcbf54b0432ef4a0c69), up to now many things have been changed. Most major updates are adding documents and examples, beside that:

1. To fulfill [ref](http://docs.julialang.org/en/stable/manual/style-guide/#append-to-names-of-functions-that-modify-their-arguments) and [ref](http://docs.julialang.org/en/stable/manual/variables/#stylistic-conventions) Julia style guide, but keep names as close as possible to originals, I've changed all function names to lower case, and add '!' for ones which write to their arguments.

Also from Julia doc:

> Word separation can be indicated by underscores ('_'), but use of underscores is discouraged unless the name would be hard to read otherwise. 

...So, IMO, it is better to simplify names by removing `abstractstate` phrase and `_` from them. Do you think `updateandcommonout!` is harder to read than `update_and_common_out!` is to type?

2. To keep backward compatibility of 
   **A)** Older Julia, I've decided to `use Compat`  
   **B)** Older `CoolProp.jl` I added older symbols at the end of the wrapper [here](https://github.com/DANA-Laboratory/CoolProp/blob/master/wrappers/Julia/CoolProp.jl)

3. To make calls generic, less error prone and clear it is important to apply guidance from [documentation ](http://docs.julialang.org/en/stable/manual/calling-c-and-fortran-code/) carefully. e.g. :

> Julia automatically inserts calls to the `cconvert()` function to convert each argument to the specified type....

So we could use most generic `Integer`, `Real` types (I think in places that a number normally is return value of another function e.g. `handle::Clong` or `input_pair::Clong` keeping the type solid is more clear so I did). Also refer the same part form Julia doc:

> When an array is passed to C as a `Ptr{T}` argument, it is not reinterpret-cast: Julia requires that the element type of the array matches T, and the address of the first element is passed...

So avoid using `Array{Real}`, because as the above mentioned the elements of an array do not convert to C automatically.

Also to be more specific and clear I use `Cstring` in place of a `Ptr{UInt8}` in most cases.

4. To help users get a simple list of all `parameters` and `fluids`, I build 2 tables and read them into Doc strings. Yet the process of building table files, is a part of testing but it may move some where else future.     

[More is on the way....](https://github.com/vimalaad/CoolProp.jl)


ibell:
It would be cool if you could document how you generated the tables, etc. by providing them as scripts.  Also, you should try to find a way to auto-update the configuration variables in the docs, that is important to keep up-to-date.

DANA-Laboratory:
> With regards to the julia interface, my preference is to keep the function names, etc., as close as possible to being exactly the same as in C++/python. Things that are not already in C++, you are welcome to name what you like.

The current position is __we keep the exact `C++` names plus a Julia compatible one__. Between differences is minimal:
  * lower the case
  * add `!` when necessary

I think it comply with the above quote.

DANA-Laboratory:
One thing really hurt my heart:
Why do not return multiple variables when need e.g. `AbstractState_update_and_5_out()`
any side effect?

JonWel:
@DANA-Laboratory I think Ian's requirements on the names are linked to a will to ease maintainability for anyone working on any of the wrappers without the need to learn a specific syntax.
Thus having first function having Julia specific name, and later, on different places, the conversion to CoolProp names do not necessarily ease the maintenance.

I don't know the exact thoughts of Ian regarding having Julia style adapted syntaxes as a plus, but if we go for this, I would rather first see the function names as the CooolProp way, and just after the function definition, the specification of the Julia's style of naming variables, and keep the `export` around too. For instance:
```Julia
export PropsSI, propssi
function PropsSI(output::AbstractString, name1::AbstractString, value1::Real, name2::AbstractString, value2::Real, fluid::AbstractString)
  val = ccall( (:PropsSI, "CoolProp"), Cdouble, (Cstring, Cstring, Cdouble, Cstring, Cdouble, Cstring), output, name1, value1, name2, value2, fluid)
  if val == Inf
    error("CoolProp: ", get_global_param_string("errstring"))
  end
  return val
end
const propssi = PropsSI
```

The basic Idea would be to have all things concerning a given function at the same place to ease maintenance by people not using this wrapper on a daily basis.


I guess Ian wants a way to automate the update of the two tables when things change on the CoolProp source code.
For instance, if a new fluid is added, not having to go add manually the new line in the wrapper to have it ending up up to date.
So if these tables could be generated in `deps\build.jl`, at each release by reading the relevant files of the source code, I guess that would be really nice.

Keep in mind that I only give suggestions, I have no decision power here, and I still have a lot to learn :).

JonWel:
> Why do not return multiple variables when need e.g. `AbstractState_update_and_5_out()`
Not sure to understand.
This function returns multiple variables defined in the outputs vector.

But we may make a more generic function who could allows more flexibility when using Julia:
```C++
EXPORT_CODE void CONVENTION AbstractState_update_and_multiple_out(const long handle, const long input_pair, const double* value1, const double* value2, const long length, const long nParam, long* outputs, double* out, long *errcode, char *message_buffer, const long buffer_length)
{
    *errcode = 0;
    try{
        shared_ptr<CoolProp::AbstractState> &AS = handle_manager.get(handle);

        for (int i = 0; i<length; i++){
            try{
                AS->update(static_cast<CoolProp::input_pairs>(input_pair), *(value1+i), *(value2+i));
                for (int p = 0; p<nParam; p++){
                    *(out+i+p*length) = AS->keyed_output(static_cast<CoolProp::parameters>(outputs[p]));
                }
            }
            catch (...){

            }
        };
    }
    catch (...) {
		HandleException(errcode, message_buffer, buffer_length);
	}
}
```

```Julia
function AbstractState_update_and_multiple_out!{F<:Float64}(handle::Clong, input_pair::Clong, value1::Array{F}, value2::Array{F}, outputs::Array{Clong}, out::Array{F})
  n, s2 = size(out)
  nParam = length(outputs)
  s2 == nParam ? nothing : error("Incompatible dimensions between outputs vector (lenth ", nParam, ") and out array size ("n,",",s2")")
  ccall( (:AbstractState_update_and_multiple_out, "CoolProp"), Void, (Clong, Clong, Ref{Cdouble}, Ref{Cdouble}, Clong, Clong, Ref{Clong}, Ref{Cdouble}, Ref{Clong}, Ptr{UInt8}, Clong), handle, input_pair, value1, value2, n, nParam , outputs, out, errcode, message_buffer::Array{UInt8, 1}, buffer_length)
  raise(errcode, message_buffer)
  return nothing
end
```
So basically using an `out` 2D array instead of 1D, and use linear indexing to fill it.
Of course, this could be tuned, I did this quickly to get an idea and there may be some typos.

DANA-Laboratory:
@JonWel , the `AbstractState_update_and_multiple_out` concept could happen but my idea is simply:
```julia
function abstractstate_update_and_common_out!{F<:Float64}(handle::Clong, input_pair::Clong, value1::Array{F}, value2::Array{F}, length::Integer, T::Array{F}, p::Array{F}, rhomolar::Array{F}, hmolar::Array{F}, smolar::Array{F})
  ccall( (:AbstractState_update_and_common_out, "CoolProp"), Void, (Clong, Clong, Ref{Cdouble}, Ref{Cdouble}, Clong, Ref{Cdouble}, Ref{Cdouble}, Ref{Cdouble}, Ref{Cdouble}, Ref{Cdouble}, Ref{Clong}, Ptr{UInt8}, Clong), handle, input_pair, value1, value2, length, T, p, rhomolar, hmolar, smolar, errcode, message_buffer::Array{UInt8, 1}, buffer_length)
  raise(errcode, message_buffer)
  return T, p, rhomolar, hmolar, smolar
end

function abstractstate_update_and_common_out!{F<:Float64}(handle::Clong, input_pair::Clong, value1::Array{F}, value2::Array{F}, length::Integer)
  T, p, rhomolar, hmolar, smolar = [zeros(length) for i=1:5]
  return abstractstate_update_and_common_out!(handle, input_pair, value1, value2, length, T, p, rhomolar, hmolar, smolar)
end
```  

**Also I hate `length`**

DANA-Laboratory:
> `const propssi = PropsSI` :+1:

DANA-Laboratory:
https://github.com/vimalaad/CoolProp.jl/commit/81ca6868ff95f47677b00b0c04e986e9f3ff6275

JonWel:
@DANA-Laboratory Ok, I get it now, this is the fact that the functions return nothing and just modify the variables internally.
I personally see no needs to output anything as the initial vectors are already modified.
I may be wrong, but to me, outputting means making a copy, thus wasting some time.

However, I think your second function could be useful.
I however think this is better to fill with `NaN` as this can help know the values who would remain `NaN` are failed, whereas 0 could be misleading (The library keep untouched the values of failed updates).
Thus I suggest having these two functions:
```Julia
function abstractstate_update_and_common_out!{F<:Float64}(handle::Clong, input_pair::Clong, value1::Array{F}, value2::Array{F}, n::Integer, T::Array{F}, p::Array{F}, rhomolar::Array{F}, hmolar::Array{F}, smolar::Array{F})
  ccall( (:AbstractState_update_and_common_out, "CoolProp"), Void, (Clong, Clong, Ref{Cdouble}, Ref{Cdouble}, Clong, Ref{Cdouble}, Ref{Cdouble}, Ref{Cdouble}, Ref{Cdouble}, Ref{Cdouble}, Ref{Clong}, Ptr{UInt8}, Clong), handle, input_pair, value1, value2, n, T, p, rhomolar, hmolar, smolar, errcode, message_buffer::Array{UInt8, 1}, buffer_length)
  raise(errcode, message_buffer)
  return nothing
end

function abstractstate_update_and_common_out{F<:Float64}(handle::Clong, input_pair::Clong, value1::Array{F}, value2::Array{F}, n::Integer)
  T, p, rhomolar, hmolar, smolar = [fill(NaN,n) for i=1:5]
  abstractstate_update_and_common_out!(handle, input_pair, value1, value2, n, T, p, rhomolar, hmolar, smolar)
  return T, p, rhomolar, hmolar, smolar
end
```

I think that it in deed may not be good to have the name `length` as a variable (as it is already a function). It doesn't seem to make problems here as we usually do not use the function `length` inside of our functions, but I may be a better practice to use a better name in the Julia wrapper.
We may also want to overload some functions with the length computed inside, in order to simplify the calls, but I like the idea to keep the one with the length as parameter as I usually already have this variable, so can use it directly for more straight calls, particularly the possibility to call only on a part of the input vector.
What about using `n` instead of `length` ? Or would there be a more common way to express this? (it needs to be clear, even if the help explain what the variables means).

JonWel:
@DANA-Laboratory I see now the table building.
Shouldn't there be two functions for each table?
One `parametersbuild()` building the table, and the second `parameters()` witch, if the file exist, load the table, and if no file, build it? (and force the table to rebuild in the /dev/build.jl , as it's done now)?

(Of course, the same for the fluids table)

DANA-Laboratory:
I just update `fluids()` and `parameters()` https://github.com/CoolProp/CoolProp/pull/1384/commits/b7e592728446bb801810bfbc05824bf714ff3f60 so thay do the building task for related tables on-demand. also building process have been moved to `Pkg.build()` with https://github.com/vimalaad/CoolProp.jl/commit/81ca6868ff95f47677b00b0c04e986e9f3ff6275
____
length->n :+1:  But as @JonWel truly mention, I think keeping it in place, without any change would be better. (more clear, minimal change, ...)
____
About multiple returning in update functions I would like to draw your attention to many build-in update functions that return the updated argument e.g. `At_ldiv_B!`, `axpy!`, ...
really its a better practice to return updated argument.

DANA-Laboratory:
@JonWel , I don`t want to define new functionality for CoolProp, really the only usability of these two are documenting, as user could get a list by `?parameters` or `?fluids`. or update Doc by `fluids()` and `parameters()`.

JonWel:
@DANA-Laboratory I don't know yet how the documentation work, I'll have to experience it a bit.
I was thinking of distinct functions in case the documentation would have needed these functions to load the files, in order to have a quick access to the documentation.

> really its a better practice to return updated argument.

So, I checked and the returned values are just references, thus no copy overhead.
And we don't have to specify return, we can directly write the list:
```Julia
function abstractstate_update_and_common_out!{F<:Float64}(handle::Clong, input_pair::Clong, value1::Array{F}, value2::Array{F}, n::Integer, T::Array{F}, p::Array{F}, rhomolar::Array{F}, hmolar::Array{F}, smolar::Array{F})
  ccall( (:AbstractState_update_and_common_out, "CoolProp"), Void, (Clong, Clong, Ref{Cdouble}, Ref{Cdouble}, Clong, Ref{Cdouble}, Ref{Cdouble}, Ref{Cdouble}, Ref{Cdouble}, Ref{Cdouble}, Ref{Clong}, Ptr{UInt8}, Clong), handle, input_pair, value1, value2, n, T, p, rhomolar, hmolar, smolar, errcode, message_buffer::Array{UInt8, 1}, buffer_length)
  raise(errcode, message_buffer)
  T, p, rhomolar, hmolar, smolar
end

function abstractstate_update_and_common_out{F<:Float64}(handle::Clong, input_pair::Clong, value1::Array{F}, value2::Array{F}, n::Integer)
  T, p, rhomolar, hmolar, smolar = [fill(NaN,n) for i=1:5]
  abstractstate_update_and_common_out!(handle, input_pair, value1, value2, n, T, p, rhomolar, hmolar, smolar)
end
```

DANA-Laboratory:
Here `retrun` is optional (because it is the last line), but IMO helps the code being more clear. 
I just withdrawal of `;`s :)

DANA-Laboratory:
latest changes:

- almost all names are exactly original then `const propssi = PropsSI` as suggested by @JonWel 
- `fluids()` and `parameters()` help functions are replaces by `const CoolProp_fluid` and `const CoolProp_parameters`
- building help tables is done by `CoolProp.buildfluids()` and `CoolProp.buildparameters()`

JonWel:
Hi @DANA-Laboratory ,
Shouldn't we have, as you suggested, the `!` version returning the modified output (as these are just references), and reuse that function for the one who do not need to specify the output vectors?

JonWel:
this way:
```Julia
function AbstractState_update_and_common_out!{F<:Float64}(handle::Clong, input_pair::Clong, value1::Array{F}, value2::Array{F}, length::Integer, T::Array{F}, p::Array{F}, rhomolar::Array{F}, hmolar::Array{F}, smolar::Array{F})
  ccall( (:AbstractState_update_and_common_out, "CoolProp"), Void, (Clong, Clong, Ref{Cdouble}, Ref{Cdouble}, Clong, Ref{Cdouble}, Ref{Cdouble}, Ref{Cdouble}, Ref{Cdouble}, Ref{Cdouble}, Ref{Clong}, Ptr{UInt8}, Clong), handle, input_pair, value1, value2, length, T, p, rhomolar, hmolar, smolar, errcode, message_buffer::Array{UInt8, 1}, buffer_length)
  raise(errcode, message_buffer)
  return T, p, rhomolar, hmolar, smolar
end

function AbstractState_update_and_common_out!{F<:Float64}(handle::Clong, input_pair::AbstractString, value1::Array{F}, value2::Array{F}, length::Integer, T::Array{F}, p::Array{F}, rhomolar::Array{F}, hmolar::Array{F}, smolar::Array{F})
  return AbstractState_update_and_common_out!(handle, get_input_pair_index(input_pair), value1, value2, length, T, p, rhomolar, hmolar, smolar)
end
 
function AbstractState_update_and_common_out{F<:Float64}(handle::Clong, input_pair::Clong, value1::Array{F}, value2::Array{F}, length::Integer)
  T, p, rhomolar, hmolar, smolar = [fill(NaN,length) for i=1:5]
  return abstractstate_update_and_common_out!(handle, input_pair, value1, value2, n, T, p, rhomolar, hmolar, smolar)
end

function AbstractState_update_and_common_out{F<:Float64}(handle::Clong, input_pair::AbstractString, value1::Array{F}, value2::Array{F}, length::Integer)
  return AbstractState_update_and_common_out(handle, get_input_pair_index(input_pair), value1, value2, length)
end
```

DANA-Laboratory:
@JonWel you are right, but as @ibell mentioned, I prefer not to change any name. 

> With regards to the julia interface, my preference is to keep the function names, etc., as close as possible to being exactly the same as in C++/python. Things that are not already in C++, you are welcome to name what you like.

JonWel:
@DANA-Laboratory I did not want to focus on the `!`, but on the idea to having only the first function calling the library and outputting the references to the outputs vectors, and all the other ending calling the first function.
I added the `!` to ease the readability, not to say you should use it. Sorry for it is misleading.

DANA-Laboratory:
@JonWel , sorry that was my fault. 
any way.. Yes
also I had that in my mind, but with a little bit difference: (I am not sure to remove `return nothing`s)

JonWel:
@DANA-Laboratory I thought it was upsetting you ^^

> One thing really hurt my heart:
Why do not return multiple variables when need e.g. AbstractState_update_and_5_out()
any side effect?

And I see no negative side effects (it just returns by reference, this means no overhead for copy, and if the function is called as before, the behaviour will be very similar), and now I see the positive side that it allows more flexible use (like reuse it in the other functions).

And I have to thank you to have point this issue to me, this allowed me to improve my knowledge.

DANA-Laboratory:
@ibell , If you find this useful please merge.

ibell:
There is so much going on here - generally I prefer more bite-sized PR, but thanks for spending all the time to update the Julia interface.  Hopefully all of you have taken a look and are in agreement that it is ready to merge.  @JonWel - do you agree that this is ready to merge?

wrappers/Julia/CoolProp.jl
```diff
@@ -1,190 +1,677 @@
-VERSION < v"0.4.0" && __precompile__()
```
JonWel: Why do you remove this?
It should allows to pre-compile at first use than use directly the compiled version on the next calls (until an update where it compiles again)
This makes quicker loading of the library once compiled.

JonWel: By the way, should be:
```Julia
VERSION > v"0.4.0" && __precompile__()
```
Or even
```Julia
VERSION >= v"0.4.0-dev+6521" && __precompile__()
```
The line had a problem.

JonWel:
@DANA-Laboratory I think it would be nice to add a line or two in the online doc to specify help is now available and give an example on how to use it: https://github.com/CoolProp/CoolProp/blob/master/Web/coolprop/wrappers/Julia/index.rst

This could be done in a future PR I think.

@ibell I think you can merge this. It is still backward compatible and the help it introduce is really a nice feature to have. The use of `Compat` will help to maintain backward compatibility in an easy way (we can focus on keeping compatible with the last release only).

DANA-Laboratory:
@JonWel , Agree with **....could be done in a future PR I think.**, and many thanks for your worthy help and support.

JonWel:
@DANA-Laboratory yet, this is probably better in a next PR.
The updating the website instruction is not needed to be able to use the new wrapper, this is just needed to make people aware of the new features.
This will also be required when the package will be registered (to explain how people can install with the registered package). So this may be good to mainly add things that will be useful then, or even only update at that time.

wrappers/Julia/CoolProp.jl
DANA-Laboratory: To find what is the best practice for  `__precompile__()` I had a survey by checking many of the most favorite Julia packages. (also using [github search result](https://github.com/search?utf8=%E2%9C%93&q=__precompile__&type=Code&ref=searchresults) is helpful) , results show that most of package developers do not care about old Julia versions e.g.: 
1. [HDF5](https://github.com/JuliaIO/HDF5.jl/blob/efc4a3ae5cca7e46d246fe0e1e305e4dd3a8b0e8/src/HDF5.jl#L1)
2. [ACME](https://github.com/HSU-ANT/ACME.jl/blob/develop/src/ACME.jl#L4)
3. [ApproxFun](https://github.com/JuliaApproximation/ApproxFun.jl/blob/development/src/ApproxFun.jl#L1)
4. MixedModels.jl
5. NLopt.jl
6. Optim.jl
7. Plots.jl
8. PyPlot.jl
9. QuantEcon.jl
10. RCall.jl
11. Reactive.jl
12. Requests.jl
13. StatsBase.jl
14. Sundials.jl
15. TensorFlow.jl
16. Weave.jl

or even don't add `__precompile__()` on master (some had commented it) (and I didn't look at tags):

1. ArrayFire.jl
2. MachineLearning.jl
3. Mocha.jl
4. Morsel.jl
5. Mux.jl
6. AudioIO.jl
7. OnlineStats.jl
8. OpenCL.jl
9. Orchestra.jl
10. ParallelAccelerator.jl
11. Playground.jl
12. RDatasets.jl
13. Reel.jl
14. TextAnalysis.jl
15. TypeCheck.jl
16. XGBoost.jl

IMO, `__precompile__()` make it fast for users but [buggy for developers](http://stackoverflow.com/questions/40116045/why-is-julia-taking-a-long-time-on-the-first-call-into-my-module/40140349#40140349).  
So I prefer to add a comment on master and let it happen on release. In this case we will need to look for a tagged suitable CoolProp wrapper when he `Pkg.add("CoolProp")`. (maybe by using same tag numbers in both Repos).

JonWel: The version is only relevant because the `__precompile__()` did not exist before version 4.
So if their package requires Julia 4 or 5, they simply do not need to check before as it is already done in the requirements.

I see that in the repository to make the Julia package, we have `julia 0.4` as requirement. Thus, if we keep this requirement, we do not need the other checking. I'm however wondering if we shouldn't lower this to `julia 0.3` as it should still be compatible with thanks to `Compat` and to specify the tag in the same way as [Compat](https://github.com/JuliaLang/Compat.jl/blob/master/src/Compat.jl#L1), so having the version limit.
Before this update, the wrapper was still compatible with `julia 0.3`, but I have no idea if there are any users of CoolProp still using Julia 0.3. So I don't know if we should keep compatibility with (it just look to me it is a no pain move thanks to Compat).

This do not need to be on the development and could be only on the release. However, I think it can be dangerous as we would not see issues with top level statements being executed at compile time and no more at run time. Having it during the development also allows to debug all issues before making a release.

JonWel:
@ibell I think you can merge this.

DANA-Laboratory:
@ibell , What do you prefer to register CoolProp package from here?

ibell:
Here comes the merge!

PR #1402, opened 2017-01-26
set_reference_state for AbstractSate

tobias-loew:
this allows for setting reference state without changing the state of the JSONLibrary


tobias-loew:
I'll do it again without range-based for-loops

Issue #1311, opened 2016-11-02
Set VTPR constants through the set_binary_interaction_parameter function


ianhbell:
This would be aij, aji... needed for fitting these parameters for VTPR models

ibell:
This is a problem because the UNIFAQ library is a const reference. @JonWel do you have any ideas?

JonWel:
The library is in deed a const.
However, `UNIFAQMixture` is not const, and it's where we need these parameters.
`set_interaction_parameters` is where the parameters (aij, aji, bij, bji, cij, cji) are extracted from the library.
So if we overload this `set_interaction_parameters` with a version where these parameters can be given as input, we can have what we want without needing to change the library.

We have to give the parameters with corresponding `mgi`, so it can place the coefficients in the right place in the `std::map<std::pair<int, int>, UNIFAQLibrary::InteractionParameters> interaction;`.

JonWel:
I have made some changes in order to speed up VTPR in my branch [VTPR_speed](https://github.com/JonWel/CoolProp-f/tree/VTPR_speed).

Improvement is not as big as I wanted, but on a binary mixture and QT_INPUTS (I compute 200 points of the saturation curve), the time is reduced by 27%, which is already better (the before and after are both with -march=native), but still 10 times slower than PR for the same points.

One of the changes I made is to [set the interaction parameters in maps depending on the sgi pair](https://github.com/JonWel/CoolProp-f/blob/VTPR_speed/src/Backends/Cubics/UNIFAQ.cpp).
The `set_interaction_parameters`can be simplified as the first step already collect these interaction parameters. I did not touch it much right now as what comes up here may also change the way we want everything to be set up there, and that this part is not recursively executed.

JonWel:
I have made some changes in order to speed up VTPR in my branch [VTPR_speed](https://github.com/JonWel/CoolProp-f/tree/VTPR_speed).

Improvement is not as big as I wanted, but on a binary mixture and QT_INPUTS (I compute 200 points of the saturation curve), the time is reduced by 27%, which is already better (the before and after are both with -march=native), but still 10 times slower than PR for the same computations.

One of the changes I made is to [set the interaction parameters in maps depending on the sgi pair](https://github.com/JonWel/CoolProp-f/blob/VTPR_speed/src/Backends/Cubics/UNIFAQ.cpp).
[...]

PR #1418, opened 2017-02-07
Improves VTPR's speed

JonWel:
Moves part of the pure fluid code that was constant out of the `set_mole_fractions` part.
Saves the computed values for Psi in order to compute it less times.

On my test case (200 QT_inputs to plot a saturation curve for "Ethane&Acetone" mixture), it allows me to shrink time from 5.13s to 3.19s, reducing execution time by around 38%.

I also corrected an error in `cd.lnGamma.insert(std::pair<int, double>(sgik, Q*s));`, where it should not be an insert as insert do not overwrite if the result is already present (or force cleaning before, but cleaning is slower, probably as map will move content when inserting new elements).
That could lead to strange results at some points and came from the fact that at the beginning the full `pure_data` variable was cleared at each temperature update.

### DIFF
--- a/src/Backends/Cubics/UNIFAQ.cpp
+++ b/src/Backends/Cubics/UNIFAQ.cpp
@@ -14,20 +14,9 @@ void UNIFAQ::UNIFAQMixture::set_interaction_parameters() {
             }
         }
     }
-}
-
-/// Set the mole fractions of the components in the mixtures (not the groups)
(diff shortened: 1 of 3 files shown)


JonWel:
@ibell Can you merge this?
I would prefer to have separate small merges for this and #1311 instead of one bigger merge.

JonWel:
These changes are include in the new #1420 PR.
I would have preferred to merge them in two separate commit, but it may creates conflicts now.

PR #1427, opened 2017-02-10
UNIFAC compile errors, closes #1426

### LINKED ISSUE #1426: UNIFAQ compile errors
@JonWel your fix to the VTPR code broke compilation on a bunch of builders with errors like:
```
c:\users\ian\slave\csharp-windows\build\src\backends\cubics\UNIFAQ.h(23): error C2864: 'UNIFAQ::UNIFAQMixture::N' : only static const integral data members can be initialized within a class [C:\Users\ian\slave\Csharp-windows\build\build\CoolProp.vcxproj]
  c:\users\ian\slave\csharp-windows\build\src\backends\cubics\UNIFAQ.h(23): error C2864: 'UNIFAQ::UNIFAQMixture::N' : only static const integral data members can be initialized within a class [C:\Users\ian\slave\Csharp-windows\build\build\CoolProp.vcxproj]
  c:\users\ian\slave\csharp-windows\build\src\backends\cubics\UNIFAQ.h(23): error C2864: 'UNIFAQ::UNIFAQMixture::N' : only static const integral data members can be initialized within a class [C:\Users\ian\slave\Csharp-windows\build\build\CoolProp.vcxproj]
```
See also http://www.coolprop.dreamhosters.com:8010/grid.

Can you please fix this?

JonWel:
This should solve it.

### DIFF
--- a/src/Backends/Cubics/UNIFAC.cpp
+++ b/src/Backends/Cubics/UNIFAC.cpp
@@ -195,7 +195,9 @@ double UNIFAC::UNIFACMixture::ln_gamma_R(const double tau, std::size_t i, std::s
     }
 }
 void UNIFAC::UNIFACMixture::activity_coefficients(double tau, const std::vector<double> &z, std::vector<double> &gamma){
-    std::size_t N = z.size();
+    if (this->N != z.size()) {
+        throw CoolProp::ValueError("Size of molar fraction do not match number of components.");
+    }
     std::vector<double> r(N), q(N), l(N), phi(N), theta(N), ln_Gamma_C(N);
     double summerzr = 0, summerzq = 0, summerzl = 0;
     for (std::size_t i = 0; i < N; ++i) {
--- a/src/Backends/Cubics/UNIFAC.h
+++ b/src/Backends/Cubics/UNIFAC.h
@@ -20,7 +20,7 @@ namespace UNIFAC
     private:
         CoolProp::CachedElement _T; ///< The cached temperature
 
-        std::size_t N = 0; ///< Number of components
+        std::size_t N; ///< Number of components
 
         double m_T; ///< The temperature in K
         double T_r; ///< Reduce temperature


src/Backends/Cubics/UNIFAC.cpp
```diff
@@ -195,7 +195,9 @@ double UNIFAC::UNIFACMixture::ln_gamma_R(const double tau, std::size_t i, std::s
     }
 }
 void UNIFAC::UNIFACMixture::activity_coefficients(double tau, const std::vector<double> &z, std::vector<double> &gamma){
-    std::size_t N = z.size();
+    if (this->N != z.size()) {
```
ibell: Does N still get set somewhere?

src/Backends/Cubics/UNIFAC.h
```diff
@@ -20,7 +20,7 @@ namespace UNIFAC
     private:
         CoolProp::CachedElement _T; ///< The cached temperature
 
-        std::size_t N = 0; ///< Number of components
+        std::size_t N; ///< Number of components
```
ibell: If you did want to initialize this variable, the right place for that is the initializer for the class.

JonWel:
Yes, [here](https://github.com/CoolProp/CoolProp/blob/master/src/Backends/Cubics/UNIFAC.cpp#L247).
The one I removed is because I do not want it to interfere with `set_components`, which is the place it should be set (ie. I corrected something I forgot to change earlier).

src/Backends/Cubics/UNIFAC.cpp
JonWel: Yes, [here](https://github.com/CoolProp/CoolProp/blob/master/src/Backends/Cubics/UNIFAC.cpp#L247).
The one I removed is because I do not want it to interfere with `set_components`, which is the place it should be set (ie. I corrected something I forgot to change earlier).

src/Backends/Cubics/UNIFAC.h
JonWel: `set_components` should always be called, so there is no need to initialise before.
I don't know why I did initialised it to 0.

However, If we plan to use `add_component` alone somewhere to add components after having set them (add_component is public), this may be required to add something in `add_component` that would adapt `N`.

PR #1415, opened 2017-02-06
Bump IF97 revision

henningjp:
fixes #1412

PR #1428, opened 2017-02-10
Implemented Limited Error Handling in the Mathcad Wrapper

### LINKED ISSUE #1326: Mathcad Wrapper Error Messages
### Mathcad Wrapper Enhancements Needed
- [ ] Need to catch thrown exceptions, scan the error, and generate a _proper_ Mathcad error message that will pop-up in the Mathcad (15 and Prime) interfaces; instead of just returning no result.  Possible error messages:
    - Temperature Out of Range
    - Pressure Out of Range
    - Inputs Out of Range
    - Input Pair Unimplemented
    - Output Variable Unimplemented
    - Convergence Failure

### LINKED ISSUE #1424: Nonsensical Vapor Quality allowed as Input Value
The backends are allowing inputs of Q < 0 and Q > 1, which results in obscure error messages like "_rhomolar is less than zero..._" or returns a nonsensical value that is extrapolated outside the vapor dome.   I can fix it, either by:

- [ ] Setting _Q = min( max(value, 0), 1),  _- might be confusing to the user that this "clipping" is occurring_
- [x] Just set an error condition if this occurs
- [ ] Leave it alone, because there is some purpose to this that I am unaware of....

Anyone have a preference?  Obviously, users should know better, but....

henningjp:
- [x] Most common CoolProp function errors are handled by Mathcad interface; closes #1326 
- [x] CoolProp errors that aren't identified and handled as Mathcad errors flag a generic Error with instructions to check get_global_param_string("errstring") for more information.
- [x] Throw error in IF97 and HEOS Backends if user enters Q < 0 or Q > 1; closes #1424 
- [x] Updated last page of sample worksheets (both Mathad 15 and Prime and pdf's) as shown below.  
- [x] IF97 revision bump to 58b11f9 (The bump in #1415 still didn't get us to the latest patch on master).

mcad_error1
mcad_error2


ibell:
Can we not have the PDF get updated?  Or keep that somewhere else? It causes the repository to bloat in size, for a project the size of CoolProp, that's an issue.

PR #1446, opened 2017-03-04
Add function to change Q in VTPR

JonWel:
@ibell It compiles under Linux, but I did not checked yet if it works correctly.


JonWel:
VTPR still works, but the call to `set_Q_k` does not work correctly yet (seems I did not export the function in a place it's needed).

JonWel:
@ianhbell The code seems to work now (same result as before if `Q_k` is not changed or if the same value as before is imposed, and the curve moves if the value of `Q_k` changes).

I don't know if expected, but this mostly changes the value in the gaseous phase (the dew line seems to get translated). This means it may make sense to fit this value according to dew points.

PR #1463, opened 2017-03-23
Update Mathcad wrapper to use MathcadAllocate() for returned strings.

### LINKED ISSUE #1462: Mathcad Prime crashes when calling CoolProp functions that return strings
While there are no issues when compiled with VS2010, Mathcad Prime compiled with VS2015 crashes when calling functions that return string values.

henningjp:
Closes #1462 .  Tracked to legacy code in Mathcad wrapper that used standard memory allocation methods for strings (`new`) instead of using the provided `MathcadAllocate()` function, which allows Mathcad 15 and Prime to track and release memory allocations when they are no longer needed and provides consistent memory allocation across DLL boundaries.  This didn't cause a crash in either Mathcad version when compiled with VS2010, but does with VS2015 and may have been causing an undetected memory leak.  No sign of a memory leak now.  Two functions were affected, `get_global_param_string()` and `get_fluid_param_string()`.

Interestingly, I found S. Polak's inquiry about this very issue on the [Mathcad support forum](https://www.ptcusercommunity.com/thread/51416) in 2012.  It was answered correctly, but apparently never implemented in the CoolProp repository (it was on SourceForge then and version 4.0.0). 

Also cleaned up some commented code that was previously used for debugging and no longer needed.

### DIFF
--- a/wrappers/MathCAD/CoolPropMathcad.cpp
+++ b/wrappers/MathCAD/CoolPropMathcad.cpp
@@ -17,14 +17,6 @@ enum { MC_STRING = STRING };  // substitute enumeration variable MC_STRING for S
 #include "DataStructures.h"
 #include "HumidAirProp.h"
 
-/*
-#define MSGBOX(x) \
-{ \
-   std::ostringstream oss; \
-   oss << x; \
-   MessageBox(NULL,oss.str().c_str(), "CoolProp Error Message", MB_OK | MB_ICONINFORMATION); \
-}
-*/
 
 enum EC { MUST_BE_REAL = 1, INSUFFICIENT_MEMORY, INTERRUPTED,       // Mathcad Error Codes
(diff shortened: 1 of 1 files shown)


ibell:
Seems fine to me - does it work?  I'm not familiar with the guts of MathCAD

henningjp:
@ibell Yes, it works.  Tested worksheet with 100's of get param functions, recalculated over and over in both 15 and Prime, VS2010 and VS2015.  Saw a momentary rise in memory as Mathcad dealt with the memory allocations, but it leveled off and didn't appear to be leaking.  What I've read says this is the proper way to do it, although [PTC's documentation of this function](https://help.ptc.com/mathcad/en/index.html#page/PTC_Mathcad_Help/mathcad_allocate.html) leaves much to be desired.  Oh, and it crashes Prime with 100% consistency without it.  It's still a mystery to me why VS2010 appears to be more forgiving on these kinds of issues and actually creates code that runs.

PR #1481, opened 2017-04-24
Bump IF97 revision to include recent patches

henningjp:
Bump IF97 Revision to include recent patches to
- rho_PX function
- Saturation functions
- warnings and optimizations


henningjp:
@jowr & @ibell , Travis is failing here on use of ``emplace_back()`` in IF97.h (replacing ``push_back()`` in commit 5d2d294).  This is a C++11 construct that I'm assuming isn't supported in gcc, clang, Xcode,  or g++ and is apparently only emulated in MSVC10 (see [StackOverflow](http://stackoverflow.com/questions/4303513/push-back-vs-emplace-back) article).

Does this need to be reverted or is Travis CI using old versions of these compilers that don't yet support the new std::vector implementation?

ibell:
Please revert it - we need to support MSVC 2008.

henningjp:
OK.  Leave this P.R. hang for a bit.  I'm out of town this week but can shore it up next weekend.  There are other optimizations in the commit in question that are useful.  Only the emplace_back functions need to be reverted.  I'll have to hand pick.

henningjp:
@ibell  Working on this, but I have to ask...

1. Where is it documented that MSVC 2008 is a required support target?  I've added a comment to the Wiki to this regard but can't find any other documentation.
2. Why?  Will this always be the case?
3. Are there other specifically outlined coding requirements for CoolProp?  Are they written down?

PR #1562, opened 2017-10-07
Fixes out-of-sync fluids map when replacing a JSON fluid.

### LINKED ISSUE #1548: Dev Docs for Mixtures Failing
The /dev/ documentation for mixtures is failing (both for my doc builds and the official http://www.coolprop.org/dev/fluid_properties/Mixtures.html page).
ipython error

It looks to me like the ipython instance of CoolProp already thinks that the Helium/Xenon mixture has been added, so when the script shows an example of creating a new mixture, it fails.  This does NOT happen on the official 6.1.0 release documentation page.  It doesn't happen when issuing these commands from a fresh python or Jupyter iPython window either.  I've poked around at it, but am now at a loss as to what is going on.  It seems to me that this page should be creating a fresh instance of CoolProp, but apparently it's not.

If you have any ideas, I can work on it, but right now I'm stumped.  I don't know if it's an environment issue or something that has crept in with 6.1.1dev.

henningjp:
As discussed in issue #1548, when a JSON fluid is added to the FluidLibrary, replacing a fluid that already exists, as in the example in the High-Level Interface docs, the fluid_map gets corrupted.  The fluid to be replaced is being erased from the fluid_map, and then added back in at the max index + 1.  Unfortunately, the fluids with higher indices are not re-indexed, and so the last fluid in the list (Xenon in this case) gets overwritten.  Future references to Xenon then pull all fluid info and properties for the fluid that was re-inserted; the Xenon data is lost.  Logic is not flawed when new fluids are added, as they are when the initial JSON library is loaded.

The fix proposed here to the `JSONFluidLibrary:add_one` routine does not erase the fluid to be replaced, but sets the insertion index appropriately to overwrite the old fluid data.  None of the other fluids are then affected and the size of the fluid_map is unchanged.  IMHO, the resulting code is a little cleaner without all the erasure code.  The fixed code has been tested under python for the sequence of actions that was causing errors to crop up in the dev/docs.  A before and after sequence is attached as PDF files to show that the bad behavior has been corrected and the materials involved are left unchanged.

[Before Fix.pdf](https://github.com/CoolProp/CoolProp/files/1364293/Before.Fix.pdf)
[After Fix.pdf](https://github.com/CoolProp/CoolProp/files/1364294/After.Fix.pdf)

This patch closes #1548 .

### DIFF
--- a/src/Backends/Helmholtz/Fluids/FluidLibrary.cpp
+++ b/src/Backends/Helmholtz/Fluids/FluidLibrary.cpp
@@ -102,12 +102,21 @@ void JSONFluidLibrary::add_one(rapidjson::Value &fluid_json)
 {
     _is_empty = false;
     
+    // The variable index is initialized to the size of the fluid_map.
+    // Since the first fluid_map key equals zero (0), index is initialized to the key
+    // value for the next fluid to be added. (e.g. fluid_map[0..140]; index = 141 )
     std::size_t index = fluid_map.size();
     
-    CoolPropFluid fluid;
+    CoolPropFluid fluid;     // create a new CoolPropFluid object
     
-    // Fluid name
-    fluid.name = fluid_json["INFO"]["NAME"].GetString(); name_vector.push_back(fluid.name);
+    // Assign the fluid properties based on the passed in fluid_json
+    // =============================================================
+    // Parse out Fluid name   
+    fluid.name = fluid_json["INFO"]["NAME"].GetString(); 
+    
+    // Push the fluid name onto the name_vector used for returning the full list of library fluids
+    // If it is found that this fluid already exists in the library, it will be popped back off below.
+    name_vector.push_back(fluid.name);
     
     try{
         // CAS number
@@ -118,11 +127,13 @@ void JSONFluidLibrary::add_one(rapidjson::Value &fluid_json)
         if (!fluid_json["INFO"].HasMember("REFPROP_NAME")){ throw ValueError(format("fluid [%s] does not have \"REFPROP_NAME\" member",fluid.name.c_str())); }
         fluid.REFPROPname = fluid_json["INFO"]["REFPROP_NAME"].GetString();
         
+        // FORMULA
         if (fluid_json["INFO"].HasMember("FORMULA")){
             fluid.formula = cpjson::get_string(fluid_json["INFO"], "FORMULA");
         }
         else{ fluid.formula = "N/A"; }
         
+        // Abstract references
         if (fluid_json["INFO"].HasMember("INCHI_STRING")){
             fluid.InChI = cpjson::get_string(fluid_json["INFO"], "INCHI_STRING");
         }
@@ -199,7 +210,7 @@ void JSONFluidLibrary::add_one(rapidjson::Value &fluid_json)
             parse_melting_line(fluid_json["ANCILLARIES"]["melting_line"], fluid);
         }
         
-        // Parse the environmental parameters
+        // Parse the transport property (viscosity and/or thermal conductivity) parameters
         if (!(fluid_json.HasMember("TRANSPORT"))){
             default_transport(fluid);
         }
@@ -210,7 +221,10 @@ void JSONFluidLibrary::add_one(rapidjson::Value &fluid_json)
         // If the fluid is ok...
         
         // First check that none of the identifiers are already present
-        // Remember that index is already initialized to fluid_map.size() = max index + 1
+        // ===============================================================
+        // Remember that index is already initialized to fluid_map.size() = max index + 1.
+        // If the new fluid name, CAS, or aliases are found in the string_to_index_map, then
+        // the fluid is already in the fluid_map, so reset index to it's key.
         
         if (string_to_index_map.find(fluid.CAS) != string_to_index_map.end()){
             index = string_to_index_map.find(fluid.CAS)->second;                 //if CAS found, grab index
@@ -236,8 +250,8 @@ void JSONFluidLibrary::add_one(rapidjson::Value &fluid_json)
             }
         }
         
-        if (index != fluid_map.size()){        // Fluid already in list if index was reset to < fluid_map.size()
-            name_vector.pop_back();            // Pop duplicate name off the back of the name vector; otherwise it keeps growing!
+        if (index != fluid_map.size()){   // Fluid already in list if index was reset to something < fluid_map.size()
+            name_vector.pop_back();       // Pop duplicate name off the back of the name vector; otherwise it keeps growing!
             if (!get_config_bool(OVERWRITE_FLUIDS)){
                 throw ValueError(format("Cannot load fluid [%s:%s] because it is already in library; index = [%i] of [%i]; Consider enabling the config boolean variable OVERWRITE_FLUIDS", fluid.name.c_str(), fluid.CAS.c_str(), index, fluid_map.size()));
             }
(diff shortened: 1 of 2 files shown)


henningjp:
Additional verification provided by regenerating the /dev/ documentation.  Errors that were seen before in #1548 are gone and all examples on all pages provide expected results, even though the example on the HighLevelAPI page replaces the JSON for Water (with a copy of itself).

Here's the previously offending HighLevelAPI section on adding fluids:
hlapi_adding_fluids

And here's the first example on the Mixtures page with no errors:
mixtures_noerrors

@ibell , pull this in when ready or give me the OK and I'll do it.

henningjp:
bump.

ibell:
In principle that looks good, but what if you add a new fluid one million times (e.g. in fitting an EOS)?  Don't you want to erase the old one?

ibell:
Does this actually work properly?  If you overwrite the fluid a million times (e.g., in EOS fitting), won't the size of the library grow without bound?  That's why it is is key to overwrite (and replace) the fluid in the map....

henningjp:
@ibell ,
That is the beauty of the STL map class [] operator.  It does the replace for us at the [index] key specified if the key(=index) is already in the map; adds a new entry if the key(=index).  See the [MSDN reference](https://msdn.microsoft.com/en-us/library/s44w4h2s.aspx?f=255&mspperror=-2147217396#Anchor_35 "STL map Class reference") and this [reference](http://www.cplusplus.com/reference/map/map/operator[]/ "Cplusplus.com reference").  The first line of ``JSONFluidLibrary:add_one`` initializes index to the length of the map.  If the keys are sequential (which they are), index will equal the next key value to be _added_ into the list.
```cpp
    std::size_t index = fluid_map.size();
```
Then, we 
1) Create a new fluid object based on the passed in JSON fluid
2) Look for the fluid CAS, names, and aliases in the existing map and set index to the fluid_map key if the proposed fluid is found in the list.   

None of this has changed from the original code.  Now, however, 
```cpp
        // index now holds either 
        //    1. the index of a fluid that's already present, in which case it will be overwritten, or
        //    2. the fluid_map.size(), in which case a new entry will be added to the list
        
        // Add/Replace index->fluid mapping
        // If the fluid index exists, the [] operator replaces the existing entry with the new fluid;
        // if not, it will add the (index,fluid) pair to the map using the new index value (fluid_map.size())
        fluid_map[index] = fluid;
```
We don't have to erase the old entry manually, because the [] operator does the replace; assigning the reference at the iterator to our assigned new ``fluid`` and destructing the old one.   Don't confuse the index key with an array[index].  In an array, we'd have to do this erase and re-insert manually; in an STL map we don't.

The same thing applies to the other maps used to look up the JSONstring for a fluid index and reverse lookup map to find the fluid index given the name, CAS, or alias string.
```cpp
        // Add/Replace index->JSONstring mapping to easily pull out if the user wants it
        // Convert fuid_json to a string and store it in the map at index.
        // if the fluid index exists, the [] operator replaces the existing entry with the new JSONstring;
        // if not, it will add the new (index,JSONstring) pair to the map.
        JSONstring_map[index] = cpjson::json2string(fluid_json);
        
        // Add/Replace CAS->index mapping
        // This map helps find the index of a fluid in the fluid_map given a CAS string
        // If the CAS string exists, the [] operator will replace index with an updated index number;
        // if not, it will add a new (CAS,index) pair to the map.
        string_to_index_map[fluid.CAS] = index;
        
        // Add/Replace name->index mapping
        // This map quickly finds the index of a fluid in the fluid_map given its name string
        // Again, the map [] operator replaces if the alias is found, adds the new (name,index) pair if not
        string_to_index_map[fluid.name] = index;
```

So, I'll do this:
- [ ] Commit a bunch of comments explaining this logic in the ``JSONFluidLibrary:add_one`` routine.
- [ ] Test 1 Million adds of a fluid and watch for memory leaks  (this should be fun!)

henningjp:
Another proof that this works is the problem we were originally trying to fix #1548.  When the entry for a material was manually erased, the keys (index) got out of sync.  Adding the fluid back in set the index, incorrectly, to that of Xenon.  Setting
```cpp
              fluid_map[index] = fluid;
```
did not add anything to the fluid library but simply replaced the Xenon fluid with the fluid being updated.  Setting the index correctly to the fluid being replaced and not manually removing it has the desired effect and leave the map keys (index) in numerical order. 

Additionally, here's a test Mathcad worksheet using this patch to pull the JSON out for Water, add a couple of new aliases, and add the JSON back in.  Fluid is correctly added (replacing the old data) and fluids count does not change.  
[TestAddALIAS.pdf](https://github.com/CoolProp/CoolProp/files/1444273/TestAddALIAS.pdf)

henningjp:
OK. As promised.  I put a debug statement in the ``add_one()`` routine that prints the fluid.name being added and the total size of the fluid_map once the JSON fluid has been loaded into the map.  Here's the output:
```
(CP36) C:\Users\jeffp>python
Python 3.6.2 |Continuum Analytics, Inc.| (default, Jul 20 2017, 12:30:02) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import CoolProp.CoolProp as CP
Loaded 1-Butene . Size of fluid_map is: 1
Loaded Acetone . Size of fluid_map is: 2
Loaded Air . Size of fluid_map is: 3
Loaded Ammonia . Size of fluid_map is: 4
[...]
Loaded SulfurDioxide . Size of fluid_map is: 117
Loaded SulfurHexafluoride . Size of fluid_map is: 118
Loaded Toluene . Size of fluid_map is: 119
Loaded trans-2-Butene . Size of fluid_map is: 120
Loaded Water . Size of fluid_map is: 121
Loaded Xenon . Size of fluid_map is: 122
```
NOTE: Fluid library loaded in its entirety.  There are 122 JSON fluids loaded when the CoolProp module loads.  Python is taking up **_21.9 MB_** in memory.
```
# Now set the boolean to allow overwrites
>>> CP.set_config_bool(CP.OVERWRITE_FLUIDS, True)
# Grab the JSON string for Water
>>> jj = CP.get_fluid_param_string('Water', 'JSON')
# Now add Water, which already exists, back into the fluid library
>>> CP.add_fluids_as_JSON('HEOS',jj)
Loaded Water . Size of fluid_map is: 122
>>> CP.add_fluids_as_JSON('HEOS',jj)
Loaded Water . Size of fluid_map is: 122
>>> CP.add_fluids_as_JSON('HEOS',jj)
Loaded Water . Size of fluid_map is: 122
```
NOTE: Three loads and fluid_map.size() is still 122.  Python is still taking up **_21.9 MB_** in memory.

One million adds?  Really?  OK, but I won't list the full output here!  Ready?
```
>>> for i in range(0,1000000):
...     CP.add_fluids_as_JSON('HEOS',jj)
...    
Loaded Water . Size of fluid_map is: 122
Loaded Water . Size of fluid_map is: 122
[... 1 million output messages later ...]
Loaded Water . Size of fluid_map is: 122
Loaded Water . Size of fluid_map is: 122
>>> CP.get_fluid_param_string('Water','CAS')
'7732-18-5'
>>> CP.get_fluid_param_string('Xenon','CAS')
'7440-63-3'
>>> CP.PropsSI('Dmass','T',300,'P',101325,'Water')
996.556935265202
>>> CP.PropsSI('Dmass','T',300,'P',101325,'Xenon')
5.361213267804207
```
Hmm.  The fluid_map still has 122 elements in it and it is functioning properly, but the Python memory grew steadily during the 1 million iterations to _**52.3 MB**_.  

The fluid map is stable; however, I suspect that the map is just assigning the new fluid to the map location and is not properly disposing of the old one.  I'll do a little research into this and see if it is a known problem with the [] operator.

ibell:
Thanks for the research, that's good enough for me!

PR #1568, opened 2017-10-18
updated compile.sh with cmake build of coolprop libray

fsonnino:
This Script build the CoolProp sources into a shared library, then compile the fluent udf into shared objects and finally it links the udf shared objects with the CoolProp shared library into the libudf.so

### DIFF
--- a/wrappers/Fluent/compile.sh
+++ b/wrappers/Fluent/compile.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+# This Script build the CoolProp sources into a shared library, then it links it to the Fluent udf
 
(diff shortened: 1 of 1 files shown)


ibell (changes requested):

ibell:
Thanks for this, @fsonnino, and apologies for the very long delay. This script hardcodes site-specific cluster paths (Python 2.7.3, ANSYS v150, an absolute CoolProp 6.1.0 source path) and targets a now-obsolete toolchain, so it can't be merged as a general-purpose build script. The Fluent wrapper is also no longer actively maintained. Closing — but the diff stays on record here for anyone building UDFs who wants the cmake-based approach as a reference.

PR #1587, opened 2017-11-06
[msgpack] update dependency

### LINKED ISSUE #1586: Update msgpack
Would you mind updating msgpack to a more recent version? I ran into compilation errors when attempting to build with a newer version.

xoviat:
Closes #1586.


jowr:
Thank you for the PR, I haven't checked the msgpack code, but I assume that it was necessary to pass objects/references instead of pointers?

ibell:
Is this PR complete?

xoviat:
Yes.

> but I assume that it was necessary to pass objects/references instead of pointers

They changed it to a `const&`.

ibell:
That's actually a pretty good idea on their part.  Merging...

PR #1591, opened 2017-11-10
Added Python wrapper CMake generator for vc14 for when default is higher

henningjp:
With release of Visual Studio 2017 and 2018, VS2015 (vc14) may not be the highest default on the machine.  Python wrapper setup.py needed a cmake-compiler option for vc14 to force use of VS2015 (required for Python 3.5 & 3.6) for the build when higher compiler versions exist on the machine.

Also added some ignores for Python wrapper build files left behind when building right in the local git repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,16 @@
 /dev/builder
 
 /wrappers/Python/build
+/wrappers/Python/cmake_build
+/wrappers/Python/CoolProp.egg-info
+/wrappers/Python/dist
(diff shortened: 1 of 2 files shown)


jowr:
Looks good to me, any comments @ibell ?

ibell:
Looks good to me!  Also, if we switch over to pybind11 for the interface (rather than cython), a lot of these problems will go away.  I have no idea if anyone is actually using the cython interface, perhaps I should email the mailing list.

PR #1596, opened 2017-11-17
Fix slight memory leak when replacing fluid jSON.

henningjp:
While indexing was fixed in #1562 for replacing a JSON fluid, it created a slight memory leak (only detected when replacing a fluid several hundred thousand times).  When map[index] is used to replace an existing keyed entry and the second map element is a custom type/object, it must first be erased to properly deallocate the memory before assigning the new map entry at the same keyed index.  This only needs to be done for the ``fluid_map`` and the ``JSONstring_map`` because the map data contains a ``CoolPropFluid`` object and a dynamic ``string`` type, respectively.

Testing shows that after 1 million fluid replacements, the functionality has not changed, the map sizes stay constant, and the memory allocation for the CoolProp module remains rock steady.

### DIFF
--- a/src/Backends/Helmholtz/Fluids/FluidLibrary.cpp
+++ b/src/Backends/Helmholtz/Fluids/FluidLibrary.cpp
@@ -250,9 +250,12 @@ void JSONFluidLibrary::add_one(rapidjson::Value &fluid_json)
             }
         }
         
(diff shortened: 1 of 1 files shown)


henningjp:
@ibell Pretty confident with this one and it was tested thoroughly using my scripts from #1562 .  I'm going to just merge this in.

Issue #1254, opened 2016-10-06
Automatic code formatting?


jowr:
Should we somehow integrate ClangFormat into our workflow? It could be nice to unify code formatting and to make sure that new code complies with the specified format.

jowr:
Have a look at: http://clang.llvm.org/docs/ClangFormat.html

ibell:
Especially if we can ensure that code doesn't use tabs instead of spaces, I
would be interested.  How strict is the formatting checking?  Do you
recommend one style in particular?

On Oct 6, 2016 2:00 AM, "Jorrit Wronski" <notifications@github.com> wrote:

Have a look at: http://clang.llvm.org/docs/ClangFormat.html

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/CoolProp/CoolProp/issues/1254#issuecomment-251893214>,
or mute the thread

mmoelle1:
As a side remark, I use the LLVM default format and slightly adjusted it to what I like my source codes to look like. Tabs are replaced by spaces. If you store the .clang-format file in you git-repo you can even install a git-hook that basically auto-formats files before commit.

jowr:
Yes, good point. We might want to use https://clangformat.com/ to build our format definitions.

ibell:
I'm fine with someone taking the lead on this.  For me the only thing that I *really* care about is tabs being replaced with spaces and I can live with most everything else

thorade:
For the tabs/spaces thing, an editorconfig file should be sufficient:
http://editorconfig.org/

thorade:
The Python files could be formatted using the autopep8 script.
https://pypi.python.org/pypi/autopep8

jowr:
I really like the editorconfig thingy - can you make a style-file?

ibell:
Yeah I love this idea.  Can we get it to autofix the code?  The tabs/spaces
is the most pernicious issue in my experience

thorade:
So far, I have only seen this, but never used it, nor written a file, but it should not be too hard. I'll create a minimal file that can be extended later, e.g. for other languages, more styles...
So, a tab should be four spaces, in  all languages?

For fixing the existing code: Python code can be fixed with autopep8, C++ code can probably be fixed by an advanced IDE, like [CLion from Jetbrains](https://www.jetbrains.com/clion/features/).
https://www.jetbrains.com/help/clion/changing-indentation.html

[...]

PR #1609, opened 2017-12-04
editorconfig

thorade:
Opening the PR so changes can be seen and discussed.
This is for #1254 

Note: This will only help in case a contributor uses one of the editors or IDEs that support editorconfig files.
Adding the file will not change existing code, that would have to be addressed in a different PR.

- What extensions should be covered? (*.sh, *,bat, *.yml)
- Configurations for single files are possible (CMakeLists.txt or travis.yml)
- What parts of editorconfig should be used?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,11 +7,10 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 
 # settings per extension
-[*.bat]
-
-[*.cmd]
-
 [*.{cpp,h}]
+charset = utf-8
+indent_size = 4
+indent_style = space
 
 [*.md]
 charset = utf-8
@@ -21,9 +20,3 @@ trim_trailing_whitespace = false
 charset = utf-8
 indent_size = 4
 indent_style = space
-
-[*.sh]
-
-[*.yml]
-indent_style = space
-indent_size = 2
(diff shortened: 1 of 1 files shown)


jowr:
Good start! Just a few comments:

* CMake files have 2 spaces indentation (*.cmake, CMakeListst.txt)
* What about newline characters? That was and probably is one of the biggest issues.

henningjp:
I think python scripts have to have tabs, not spaces, to specify indent levels for loops.  Right?

ibell:
Never tabs in python for indentation!!  The recommendation is 4 spaces
instead of a tab.  All reasonable editors can handle that nowadays

henningjp:
Sorry, @ibell .  My bad.  It was the Makefile that didn't like the spaces when I was building the local docs, not the python scripts.

thorade:
@henningjp For Python there is PEP8 which has rules for almost all formatting:
https://www.python.org/dev/peps/pep-0008/
And there are tools that enforce these rules, like autopep8.

@jowr Regarding newline characters: git converts newline characters upon pulling and pushing. To my understanding, a comprehensive gitattributes file should be sufficient.

thorade:
Here is a list of extensions that appear in the repo, extracted with a small script:
[fileExtensions.txt](https://github.com/CoolProp/CoolProp/files/1531017/fileExtensions.txt)

thorade:
This PR is in a state that it could be merged, it does not break anything and can be enhanced/extended later on.

PR #1610, opened 2017-12-05
only clone the last 5 commits

thorade:


ibell:
That's slick!  Can you also make it only check out the head commit of the submodules?  I think checking out Eigen is the slowest part of the clone.

thorade:
The submodules are pointing to a certain commit and it is not always sure that commit is within the last 5 commits. It would probably be possible to use the master of the external projects and only do a shallow clone.
But Travis supports caching, maybe that could be an alternative? I'll check.

thorade:
Sorry, I currently don't see a way to speed up the cloning of the submodules

Issue #1612, opened 2017-12-06
pep8


thorade:
Somewhat related to #1609 and #1254 but only for the Python files:
For python there exists pep8, a comprehensive style guide:
https://www.python.org/dev/peps/pep-0008/
and a couple of tools to check and/or fix the style,
especially [flake8](https://pypi.python.org/pypi/flake8) and [autopep8](https://pypi.python.org/pypi/autopep8), both installable via pip.

flake8 is good to get a first overview, 
autopep8 is for fixing.

Open a terminal and `cd` to the Coolprop directory, then run 
```
flake8 --count --exclude="externals" --max-line-length=200 --statistics -qq
flake8 --count --exclude="externals" --max-line-length=200 --select="E101"
autopep8 --in-place --recursive --max-line-length=200 --exclude="externals" --select="E101" .
```
and repeat, and have a look at the git diff in between. The numbers should go down.

The following (groups of) fixes are purely whitespace at the line end or beginning (so git can hide them):
```
autopep8 --in-place --recursive --max-line-length=200 --exclude="externals" --select="E101,W1,W2" .
```
There are other very safe fixes, e.g. the `E3` and `W3`group.
Most of the issues reported are in the `E2` group (e.g. missing space after comma).

scls19fr:
A big cleanup of Python files is probably still required.
https://gist.github.com/scls19fr/07bbb950a95072f7ca5de5428f89085b

I also noticed that there is some

```
print "..."
```

instead of

```
print("...")
```

in many `.py` files (see https://docs.python.org/3.0/whatsnew/3.0.html#print-is-a-function )

https://github.com/CoolProp/CoolProp/search?q=print&unscoped_q=print

scls19fr:
Checking PEP8 with Travis CI and `flake8` could be a nice improvement but a lot of work is required before enabling such a tool!

ibell:
Closing as part of an issue-tracker cleanup — this issue has had no activity in over two years.

*To be transparent: this comment and the close action are being performed by Claude (Anthropic's AI coding assistant), running on behalf of @ibell. Ian reviewed and approved the bulk-close criteria; I am the one executing the action.*

If you would like this issue reconsidered, please feel free to **re-open it** — a fresh comment with current behaviour against the latest CoolProp release is the most useful way to revive the discussion.

— Claude (acting for @ibell)

PR #1613, opened 2017-12-06
autopep8 whitespace

thorade:
This is the first batch of autopep8 changes for #1612, the exact command used was
```
autopep8 --in-place --recursive --max-line-length=200 --exclude="externals" --select="E101,W1,W2" .
```

Note: I changed the commits (and force-pushed to the branch) so that now there are only those whitespace-changes that git can hide. To verify, append `?w=1` to the URL of the diff page:
https://github.com/CoolProp/CoolProp/pull/1613/files
https://github.com/CoolProp/CoolProp/pull/1613/files?w=1


Web/develop/web2py_online.py
```diff
@@ -27,7 +27,7 @@
 ---------------------------
```
thorade: Should this be a .py file?

dev/Tickets/1443.py
```diff
@@ -108,19 +106,19 @@ def calc_isolines(fluid="HEOS::water", kind_a="P", range_a=[1e6,10e6], kind_b="T
         elif CUR_CONF==CONF_PYT_DIR:
             start = timer()
             for i,two in np.ndenumerate(vector_in1):
-                try: 
+                try:
                     state.update(input_pair, two, one)
                     for kind_o_single in kind_o_index:
                         tmp_result = state.keyed_output(kind_o_single)
                 except:
                     ++err_count
```
thorade: What is this? It is strangely highlighted.

thorade:
@jowr should i change the extension of Web/develop/web2py_online.py to .rst?

jowr:
I have no idea about web2py - but yes, if it is an rst file, why no call it that?

ibell:
Yeah, that should be a RST file

PR #1622, opened 2017-12-14
Submodules track master branch

thorade:
This is a follow-up for #1610 
Turns out it does not save a significant amount of time.

One could also remove the submodule_depth again, and only keep the other changes:
sorted the submodules alphabetically, specified to use master branch, and pulled the latest master

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,6 +25,7 @@
 [submodule "externals/pybind11"]
 	path = externals/pybind11
 	url = https://github.com/pybind/pybind11.git
+	branch = master
 [submodule "externals/rapidjson"]
 	path = externals/rapidjson
 	url = https://github.com/Tencent/rapidjson.git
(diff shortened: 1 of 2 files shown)


thorade:
I removed the submodule_depth again.
This PR includes an update for ALL submodules to their respective latest master.
If nobody has objections, I would merge it.

ibell:
I'm not sure about bumping ALL repositories.  I seem to recall there were
issues with the tippy Eigen - maybe it didn't work in MSVC2008?  Can we get
all the buildbot slaves to run on your branch before we merge?  You might
want to merge it into another branch within CoolProp/CoolProp so that we
can run the tests and determine if it is safe or not.

thorade:
This is on a branch in Coolprop already, how can I run the buildbot to test it?

thorade:
I reverted the update of the submodules,
now the only changes left is the alphabetical sorting of the submodules,
and that the branch=master is specified.

Both is harmless, and I will squash&merge this once the tests have passed.

ibell:
Tests on your branch are building at http://www.coolprop.dreamhosters.com:8010/grid

We don't have it set up to auto-build all branches pushed, but I can give you the login if you want to be able to fire tests on branches.

PR #1628, opened 2018-01-05
Parse out zero mole fraction components in PropsSI()

### LINKED ISSUE #969: Support mixtures with component mole fractions of zero
ValueError: PropsSI failed ungracefully :: inputs were:"S","P",1.0132500000000000e+05,"T",2.8814999999999998e+02,"HEOS::Nitrogen[0.770375]&Oxygen[0.206696]&CarbonDioxide[0]&Water[0.0138572]&Argon[0.00907251]";

### LINKED ISSUE #1603: Parse out Zero Mass Fraction Components in High-Level Interface
PropsSI parses out the mixture string into components and mole fractions. Since mass fractions of 0.0 are not well behaved, modify the High-Level PropsSI to parse out any "zero" mole fraction components; leaving them out of the mixture list.

henningjp:
This PR closes #1603 and also closes #969 .

In High-Level interface only, when parsing the mixture string, do not add mixture components that have a mole fraction of 0 (or negative, which is nonsensical).  Unfortunately, cannot be applied to the low-level interface.

Tested in Mathcad wrapper (curiously, error could not be reproduced in Python wrapper, at least for this example).
zerocomponenttest

### DIFF
--- a/src/CoolProp.cpp
+++ b/src/CoolProp.cpp
@@ -148,11 +148,14 @@ std::string extract_fractions(const std::string &fluid_string, std::vector<doubl
             // If pEnd points to the last character in the string, it wasn't able to do the conversion
             if (pEnd == &(fraction[fraction.size()-1])){throw ValueError(format("Could not convert [%s] into number", fraction.c_str()));}
 
-            // And add to vector
-            fractions.push_back(f);
+            if (f > 0)  // Only push component if fraction is positive and non-zero
+            {
+                // And add to vector
+                fractions.push_back(f);
 
-            // Add name
-            names.push_back(name);
+                // Add name
+                names.push_back(name);
+            }
         }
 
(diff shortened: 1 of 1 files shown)


henningjp:
@ibell , This was easier than I thought it would be.  The check is against zero (0).  I've seen PropsSI() function properly with very small mole fractions, but we could put a check to 1E-12, or similar.  However, I don't know where the threshold would be for either (a) failure, or (b) no appreciable affect on mixture.  If you have a suggestion, I'll update it.

ibell:
Wow, thanks for doing this; you're right, it isn't too complicated!!  I would probably do f > 10*DBL_EPSILON rather than f > 0 - in general it is advised to never compare floats directly against integers

henningjp:
Excellent.  Will do.

Issue #1638, opened 2018-01-28
Bug in AbstractState.pyx (aka can't build or install CoolProp python wrapper for python 3.6 under linux)


notmyrealhandle:
arg1 and arg2 used with the fused type string_or_size_t in AbstractState.pyx do not match CAS1 and CAS2 used in AbstractState.pyd.

Using:

patch < [AbstractState.patch](https://gist.github.com/notmyrealhandle/77a2c04516aa235eeae928e628f20b01)

in the directory: CoolProp/wrappers/Python/CoolProp

resolves this for me.  Verified with:

import CoolProp.CoolProp as CP
CAS_He = CP.get_fluid_param_string('Helium','CAS')
CAS_Xe = CP.get_fluid_param_string('Xe','CAS')
CP.apply_simple_mixing_rule(CAS_He, CAS_Xe, 'linear')
AS = CP.AbstractState("HEOS","Helium&Xenon")
AS.set_binary_interaction_double(0, 1, 'betaT', 0.987)
AS.get_binary_interaction_double(0, 1, 'betaT')

Out[7]: 0.987

And now the bugreport:

Environment:
lubuntu 17.10 (artful) on a dell e6220: 
uname: 4.13.0-32-generic #35-Ubuntu SMP Thu Jan 25 09:13:46 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
gcc --version: gcc (Ubuntu 7.2.0-8ubuntu3) 7.2.0
python 3.6.4 in a pyenv virtual env

After 

git clone https://github.com/CoolProp/CoolProp.git --recursive

and

cd CoolProp/wrappers/Python/

python setup.py install 

will fail with the most helpful and easiest to debug error output I have ever come across:

version written to file: /home/ul/.local/src/CoolProp/include/cpversion.h
version written to hidden file: /home/ul/.local/src/CoolProp/.version for use in builders that don't use git repo
git version 2.14.1
git is accessible at the command line
git revision is d394aa1968ad59c6afd9153b353d4acb18cfffa2
*** Generating gitrevision.h ***
/home/ul/.local/src/CoolProp/include/gitrevision.h written to file
/home/ul/.local/src/CoolProp/include/all_fluids_JSON.h written to file
/home/ul/.local/src/CoolProp/include/all_incompressibles_JSON.h written to file
/home/ul/.local/src/CoolProp/include/mixture_departure_functions_JSON.h written to file
[...]

PR #1639, opened 2018-01-31
fix inconsistent function declarations between AbstractState.pyx and AbstractState.pxd

notmyrealhandle:
See https://github.com/CoolProp/CoolProp/issues/1638


notmyrealhandle:
hmm,

somehow this went from 1 commit (aSFix) to 57...

I think something has gone wrong and this should be abandoned.

notmyrealhandle:
Please do not merge.

PR #1640, opened 2018-01-31
fix inconsistent function declarations between AbstractState.pyx and AbstractState.pxd (#1638)

notmyrealhandle:
See https://github.com/CoolProp/CoolProp/issues/1638

Apologies for PR #1639.  Hopefully this one works for you.


ibell:
Interesting - the types are the same, but the symbol names are different, and that causes cython to fail?  That's odd.

In any case, it's best to have consistent names, so I'm going to merge this one.  Thanks for going through the hurdles to get the PR together.

PR #1656, opened 2018-02-26
Mistake in function 'inline_label' in CoolProp/Plots/Common.py

xaumex:
Hi,
This pull request corrects a little mistake in function "inline_label" in CoolProp/Plots/Common.py in the python wrapper


ibell:
Thanks! Looks good to me!

PR #1660, opened 2018-03-07
Update PropsSI() to Parse Imposed Phase Strings on Input Keys

### LINKED ISSUE #1279: Add phase specification to high-level interface
In the high-level interface there is currently no way to specify the phase of the fluid, which is problematic to say the least.  For instance, just like in refprop, maybe we should allow '<' or '>' to specify liquid or vapor phase respectively.

@jowr - what do you think?  I think this would be useful.

henningjp:
**_PropsSI()_** is updated to allow the phase to be imposed by tagging one of the input key strings with the pipe symbol, **"|"** followed by a phase string (e.g. "**T|liquid**" or "**P|supercritical_gas**").  If a valid phase string is found, **_State_**->**_specify_phase()_** is called with the corresponding phase index before the call to **_update()_**.  Only one Input key string can be tagged with a phase string and error checking makes sure that:

- Imposed phase is a valid phase string
- Only one Input key is tagged with a phase
- Phase strings prefixed with "phase_" or "iphase_" are acceptable

High-Level Interface documentation is also updated to explain how to use this feature.  It has been successfully tested in both Python and Mathcad wrappers (updated to catch phase syntax errors).
Closes #1279

### DIFF
--- a/Web/coolprop/HighLevelAPI.rst
+++ b/Web/coolprop/HighLevelAPI.rst
@@ -55,6 +55,55 @@ To retrieve either the *vapor* or *liquid* properties along the saturation curve
 .. note::
    The *latent heat of vaporization* can be calculated using the difference between the vapor and liquid enthalpies at the same point on the saturation curve.
 
+Imposing the Phase (Optional)
+-----------------------------
+
+Each call to ``PropsSI()`` requires the phase to be determined based on the provided input pair, and may require a non-trivial flash calculation to determine if the state point is in the single-phase or two-phase region and to generate a sensible initial guess for the solver. For computational efficiency, ``PropsSI()`` allows the phase to be manually imposed through the input key parameters.  If unspecified, PropsSI will attempt to determine the phase automatically. 
+
+Depending on the input pair, there may or may not be a speed benefit to imposing a phase.  However, some state points may not be able to find a suitable initial guess for the solver and being able to impose the phase manually may offer a solution if the solver is failing.  Additionally,  with an input pair in the two-phase region, it can be useful to impose a liquid or gas phase to instruct ``PropsSI()`` to return the saturated liquid or saturated gas properties.  
+
+To specify the phase to be used, add the "|" delimiter to one (and only one) of the input key strings followed by one of the phase strings in the table below:
+
++---------------------------------+----------------------------------------------------+
+| Phase String                    | Phase Region                                       |
++=================================+====================================================+
+| "liquid"                        | p < pcrit & T < Tcrit ; above saturation           |
++---------------------------------+----------------------------------------------------+
+| "gas"                           | p < pcrit & T < Tcrit ; below saturation           |
++---------------------------------+----------------------------------------------------+
+| "twophase"                      | p < pcrit & T < Tcrit ; mixed liquid/gas           |
++---------------------------------+----------------------------------------------------+
+| "supercritical_liquid"          | p > pcrit & T < Tcrit                              |
++---------------------------------+----------------------------------------------------+
+| "supercritical_gas"             | p < pcrit & T > Tcrit                              |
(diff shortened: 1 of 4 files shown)


henningjp:
A bit of the verification work using the Mathcad wrapper:

propssi verif1
propssi verif2

Identical results in Python.

ibell:
Brilliant! Nice work!  Looks good from my review

ibell:
Looks like a number of tests are failing - something to do with the incompressible backend:

http://www.coolprop.dreamhosters.com:8010/builders/Java-windows/builds/2204
http://www.coolprop.dreamhosters.com:8010/builders/Csharp-OSX/builds/2213

henningjp:
@ibell Yep, so my little routine makes calls to ``AbstractState->phase()`` and ``AbstractState->specify_phase(imposed)``.  I'm thinking this whole feature makes no sense if the backend is incompressible, in which case the phase is always **_liquid_**?

Two options:
1. Bypass this feature if the Backend == Incompressible.  **_Is there an easy function call to check for this?_**
2. In the incompressible backend, implement a ``calc_phase()`` for incompressible that always returns **_iphase_liquid_**, and a ``specify_phase()`` function that simply ignores the imposed phase.

NOTE: A quick check of IncompressibleBackend.cpp and the header shows that ``update()`` just sets ``_phase`` to liquid and leaves it.  There's no override of the AbstractState functions to get or set the phase.

So I can fix this tonight.  How do I test the backend name?

ibell:
I'm a bit confused myself.  The calls that were failing in the tests did
not specify phase, and therefore, should never have hit your code, so I
suspect that there is a logical flaw in the parsing of the input flags.
Here is an example that fails currently

PropsSI("D","T",300,"P",101325,"INCOMP::MEG-50%")

henningjp:
No, I know exactly where it is.  The first thing my StripPhase() routine does is save the phase that is already imposed by calling State->Phase().  It does this even if there are no phase strings to strip off.  It doesn't end up using it, and leaves the Input key strings alone.  However, if the State is Incompressible, then calc_phase() function is still the inherited one from AbstractState, which throws an error.   I'll put that check just inside the if structure that processes the phase string if it exists.  That should fix it.  I'll PR that tonight and we can re-run the test.

ibell:
Great!  Looking forward to it!

henningjp:
Hmmm.  Some other issues.  I wasn't even thinking about the other backends.

- [x] HEOS works fine
- [x] REFPROP works (now.  Needed a tweak.  No calc_phase() function there either.)
- [x] PR works
- [x] SRK works
- [x] INCOMP fixed, but will throw an error if trying to set phase (should always be a liquid)
- [x] IF97 same as INCOMP.  It needs to set it's own phase regions.
- [ ] TREND  ?  Is that a thing?  What is it?  I can't find any docs, throw error?
- [ ] TTSE & BICUBIC  I don't think these should be called from high-level anyway - will throw error.
- [ ] VTPR Another one I can't find any documentation for.  It has no phase functions, will throw error.

yodifone:
Hi all.

To reply to your question: 

TREND ? Is that a thing? What is it? I can't find any docs, throw error?

This is a fluid property modelling library actively developed at University of Bochum. These guys who have developed almost all highly sophisticated helmholtz eos (i.e. GERG etc.).

Regards.

ibell:
@yodifone is right.  They (RuB) and NIST are the two driving forces behind the vast majority of EOS that are in REFPROP, CoolProp, and TREND.

TMI: At one point I was trying to convince RuB to spend the necessary time to develop a backend for CoolProp that would call TREND.  That would dramatically increase the number of users that could use TREND.  Never came to pass though.  We should probably remove TREND backend entirely.  We can always add it back.  Nobody would miss it because it doesn't work.

PR #1665, opened 2018-03-09
Patches PropsSI imposed phase for backends other than HEOS

henningjp:
Patches #1660.  Mods to ``StripPhase()`` in CoolProp.cpp to handle (or reject) backends that don't have phase functions like calc_phase() and specify_phase().  Specifically REFPROP and INCOMP and IF97 backends were affected.  Tests should all pass now.

Error message was something like:
```python
In [1]: PropsSI('C','T',500,'P',101325,'INCOMP::DowQ')
ValueError: This backend does not implement calc_phase function : PropsSI("C","T",500,"P",101325,"INCOMP::DowQ")
```

With the patch, these backends are all passing tests:
```python
In [4]: PropsSI('C','T',500,'P',101325,'INCOMP::DowQ')
Out[4]: 2288.1643758645673

In [7]: PropsSI("D","P",101325,"T",375,"REFPROP::Water")
Out[7]: 0.5944246707421037

In [8]: PropsSI("D","P",101325,"T|gas",375,"REFPROP::Water")
Out[8]: 0.5944246707421037

In [9]: PropsSI("D","P",101325,"T|liquid",375,"REFPROP::Water")
Out[9]: 957.0103405813556

In [10]: PropsSI("D","P",101325,"T",375,"SRK::Water")
Out[10]: 709.0245321521222

In [13]: PropsSI("D","P|gas",101325,"T",375,"SRK::Water")
Out[13]: 0.5903825428777025

In [14]: PropsSI("D","P|liquid",101325,"T",375,"SRK::Water")
Out[14]: 709.0245321521222

In [15]: PropsSI("D","P",101325,"T",375,"IF97::Water")
Out[15]: 0.5943926157152021
```

### DIFF
--- a/src/CoolProp.cpp
+++ b/src/CoolProp.cpp
@@ -436,10 +436,23 @@ void _PropsSI_outputs(shared_ptr<AbstractState> &State,
 bool StripPhase(std::string &Name, shared_ptr<AbstractState> &State)
 // Parses an imposed phase out of the Input Name string using the "|" delimiter
 {
-    phases imposed = State->phase();                          // Initialize imposed phase
     std::vector<std::string> strVec = strsplit(Name, '|');    // Split input key string in to vector containing input key [0] and phase string [1]
     if (strVec.size() > 1) {                                  // If there is a phase string (contains "|" character)
-        if (strVec.size() > 2)                                //    If there's more than on phase separator, throw error
+        // Check for invalid backends for setting phase in PropsSI
+        std::string strBackend = State->backend_name();
+        if (strBackend == get_backend_string(INCOMP_BACKEND))
+            throw ValueError("Cannot set phase on Incompressible Fluid; always liquid phase");   // incompressible fluids are always "liquid".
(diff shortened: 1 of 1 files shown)


ibell:
I'm going to merge this so that we can make the users/buildslaves happy again.  

A question though: why do you worry about getting the phase at all?  If you are talking about the high-level interface, won't the phase ALWAYS be the default phase?  If my understanding is wrong, it would in any case better to just wrap the call in a try-except block, so you don't need to specialize manually.  That's usually not the right solution, though it is called for in some special cases.

henningjp:
Yeah, I wasn't sure.  I think my initial thought was if it was already set by the first parameter, but setting anything on the second parameter will get tossed anyway.

I don't get phase() now, I just initialize the temp variable to **iphase_not_imposed**.  Then if there's any way for it to fall through the phase checking process, the State phase will just get set to **iphase_not_imposed** again.

henningjp:
I also put checks in to throw errors if the user tries to enter an imposed phase on the backends where phase shouldn't be imposed.

PR #1675, opened 2018-03-17
Let DARWIN build with libc++

### LINKED ISSUE #1671: make with libc++ instead of libstdc++ on osx?
Is it possible to compile CoolProps with libc++ instead of libstdc++? I tried setting two flags:

> -DCMAKE_EXE_LINKER_FLAGS=-stdlib=libc++ -DCMAKE_CXX_FLAGS=-stdlib=libc++

but libstdc++ still is whats linked according to `otool -L libCoolProp.dylib`, 

The flag seems to make it to the ./CMakeFiles/CoolProp.dir/link.txt file, but both libs are on that line. 

> ./CMakeFiles/CoolProp.dir/link.txt:/opt/moose/mpich-3.2/clang-5.0.0/bin/mpicxx -stdlib=libc++ -dynamiclib -Wl,-headerpad_max_install_names  -stdlib=libstdc++  ...

veeshy:
Added DARWIN_USE_LIBCPP cmake flag to allow for OSX builds to use the libc++ library.

This was needed by me since another project was built with libc++ and it caused issues trying to call CoolProp from it, adding cxx make flags did not work since libstdc++ was hard coded in the CMakeLists. Allows for and closes #1671

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,6 +106,10 @@ option (COOLPROP_MSVC_DYNAMIC
 option (COOLPROP_MSVC_DEBUG
        "Link the debug version of Microsoft Standard library to the debug builds."
        ON)
+       
+option (DARWIN_USE_LIBCPP
+        "On Darwin systems, compile and link with -std=libc++ instead of the default -std=libstdc++"
+        OFF)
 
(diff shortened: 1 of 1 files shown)


ibell:
Thanks for the PR!

PR #1681, opened 2018-03-26
Patch HEOS phase determination when p = p_crit

### LINKED ISSUE #1678: errors when calculating properties of water at critical pressure
For plotting purposes, I'm trying to calculate properties of water at critical pressure. This works fine at the critical point, but not when I move away from it:

```
from CoolProp import AbstractState

backend="?"
fluid="water"

state = AbstractState(backend, fluid)

pcrit = CP.PropsSI(fluid, "pcrit")

print("critical point")
state.update(CP.PQ_INPUTS, pcrit, 0)

smass = 4000
print("pcrit, smass={}".format(smass))
state.update(CP.PSmass_INPUTS, pcrit, smass)
```
the last line throws an error message:
```
>>> state.update(CP.PSmass_INPUTS, pcrit, 5000)
  File "<stdin>", line 1, in <module>
  File "CoolProp/AbstractState.pyx", line 90, in CoolProp.CoolProp.AbstractState.update (CoolProp/CoolProp.cpp:17930)
  File "CoolProp/AbstractState.pyx", line 92, in CoolProp.CoolProp.AbstractState.update (CoolProp/CoolProp.cpp:17822)
ValueError: Residual function in secant returned invalid number
```
I also tried using PT_INPUTS, starting at minimum temperature for water (273.16 K). AFAIK Water should be liquid at critical pressure and triple point temperature, but CP outputs an empty string for the phase and throws the same error as above. Only when I use a higher temperature (starting between 300 and 350 K), things start to work without errors.

Increasing pressure by just 1 Pa also gets rid of the errors, as in
```
state2.update(CP.PT_INPUTS, pcrit+1, tmin)`
```

Is there a reliable way of evaulating at pcrit over the whole defined range of some other variable (for example entropy or temperature, as above)?

henningjp:
### Description of the Change

The reverse ancillaries fail for water when determining phase from Pressure and either H, S, or U when the pressure is equal to the critical pressure.  The error is coming from Secant() which only gets called when the Brent() solver fails when called from the ancillaries invert() function.  This occurs because pcrit is at the very edge of the ancillary curve and any values calculated that exceed this limit, even from round off error, will cause an error.  Solution is to check for p = pcrit and set T = Tcrit without calling the ancillary solver.

### Benefits

Prevents errors that occur at any values of H, S, or U when p = pcrit.

```python
from CoolProp import AbstractState

backend="?"
fluid="water"

state = AbstractState(backend, fluid)

pcrit = CP.PropsSI(fluid, "pcrit")

print("critical point")
state.update(CP.PQ_INPUTS, pcrit, 0)

smass = 4000
print("pcrit, smass={}".format(smass))
state.update(CP.PSmass_INPUTS, pcrit, smass)
  File "<stdin>", line 1, in <module>
  File "CoolProp/AbstractState.pyx", line 90, in CoolProp.CoolProp.AbstractState.update (CoolProp/CoolProp.cpp:17930)
  File "CoolProp/AbstractState.pyx", line 92, in CoolProp.CoolProp.AbstractState.update (CoolProp/CoolProp.cpp:17822)
ValueError: Residual function in secant returned invalid number
```
Pressures just above and just below (by as small an increment as 1E-9) do not return these errors:
```python
>>> state.update(CP.PSmass_INPUTS, pcrit, smass)
  File "<stdin>", line 1, in <module>
  File "CoolProp\AbstractState.pyx", line 94, in CoolProp.CoolProp.AbstractState.update
  File "CoolProp\AbstractState.pyx", line 96, in CoolProp.CoolProp.AbstractState.update
ValueError: Residual function in secant returned invalid number
>>> state.update(CP.PSmass_INPUTS, pcrit+0.01, smass)
>>> state.update(CP.PSmass_INPUTS, pcrit-0.01, smass)
```
Same error occurs for all values of smass as long as p = pcrit.  Same error occurs with PUmass_INPUTS and HmassP_INPUTS for all values of U and H respectively when p = pcrit. 

### Possible Drawbacks

Root cause lies in Brent() and Secant() which are failing and should handle an input end-point on the curve.  This PR patches the P_Phase_Determination at p = pcrit for ``other`` inputs of H, S, or U, but does not fix the underlying issue with Secant().

### Verification Process

Patch was tested with above python sample script as well as high level calls in Mathcad.  There are no errors being thrown when update is called at p = pcrit and the expected results are returned.

```python
>>> state.update(CP.PSmass_INPUTS, pcrit, smass)
>>> state.update(CP.PSmass_INPUTS, pcrit+0.01, smass)
>>> state.update(CP.PSmass_INPUTS, pcrit-0.01, smass)
```
State is also updated all values of smass (liquid and vapor) and for PUmass_INPUTS and HmassP_INPUTS as well.  

### Applicable Issues

Closes #1678.

### DIFF
--- a/src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
+++ b/src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
@@ -1498,9 +1498,15 @@ void HelmholtzEOSMixtureBackend::p_phase_determination_pure_or_pseudopure(int ot
         // First try the ancillaries, use them to determine the state if you can
         
         // Calculate dew and bubble temps from the ancillaries (everything needs them)
-        _TLanc = components[0].ancillaries.pL.invert(_p);
-        _TVanc = components[0].ancillaries.pV.invert(_p);
-        
+        if ( (psat_max-_p) < 10*DBL_EPSILON ) {                 // However, if right at the critical pressure...
+            _TLanc = components[0].ancillaries.pL.get_Tmax();   //     Secant method in ancillaries::invert(_p) will fail,
+            _TVanc = components[0].ancillaries.pV.get_Tmax();   //     so just set T to Tmax = Tcrit = "tip" of the curve.
+        }
+        else {                                                  // Otherwise, invert the ancillary with no problems
+            _TLanc = components[0].ancillaries.pL.invert(_p);   //     Find T on liquid saturation line
+            _TVanc = components[0].ancillaries.pV.invert(_p);   //     Find T on vapor saturation line
+        }
+
         bool definitely_two_phase = false;
         
         // Try using the ancillaries for P,H,S if they are there


ibell:
I'm in general in agreement with this, but what about pseudo-pure fluids like air or R410A?  Will this still work?

henningjp:
Hmm.  You're right.  Pseudo-pure R410A fails, but in a different way.  I'll see if I can find out why and add to this PR.

henningjp:
@ibell I'm going to need a little help here.  This issue with pseudo-pure fluids was a problem before I started messing with it.  For at least R410A pseudo-pure fluid, the value returned for pcrit is 4901200.0, which is actually (from the fluid JSON) the reducing pressure.  The actual critical pressure (from the JSON and returned by ``calc_pmas_sat()``) is 4901296.586

Everything between these two values (a span of about 96 Pa) fails in the Secant method when trying to determine phase.  p = pcrit works for pseudo pure, but just a smidge over fails - all the way up to p_reducing.  This is a very different issue than I was trying to fix for pure fluids, which were failing exactly at p = pcrit.

I don't understand why p_reducing is different from p_crit for this pseudo-pure fluid (and others?).  Can you explain?  I can fix the behavior if I know how to attack it.

ibell:
For pseudo-pure fluids, there are three important points in the critical
region:
1) The maximum temperature along the isopleth (line of constant composition)
2) The maximum pressure along the isopleth
3) The critical point.

You should check out the paper on the pseudo-pure fluids.  It gives some
helpful info on this topic.

Basically, the maximum temperature point is used in place of the critical
temperature for pure fluids.  Plotting the curves in the critical region is
very instructive - it was the only way that I was able to understand what
was going on when I was getting started

henningjp:
Had a two week setback while I debugged and rebuilt my python environment that went nuts during testing.  Running comprehensive tests this week.  I'll hopefully patch this up by next weekend.

henningjp:
Closing this PR based on findings from #1678.  This is a much bigger issue that requires further investigation.

PR #1693, opened 2018-05-10
InChi standard keys added

wandadars:
### Requirements

### Description of the Change

[ Updated the fluids database InChi keys & strings to use the standard values. ]

### Benefits

[ Fluids with standardized InChi values are probably expected by users ]

### Possible Drawbacks

[ An incorrect InChi value entered during the transcription process ]

### Verification Process

[ No regression tests or verification was performed to ensure that InChi values are correct(unsure even how to go about testing the InChi values in CoolProp) 
I used the CAS number to lookup the species from PubChem: https://pubchem.ncbi.nlm.nih.gov/compound/8255#section=Top

That way at least I could be sure that the InChi values that I obtained were 100% for the fluid that I thought it was.

]


wandadars:
I'm torn between:
1.) leaving the individual commits because each one contains the reference page where the InChi key & string were obtained.
2.) Squashing all the commits down & leaving a note with the PubChem website & a comment that the CAS codes were used to lookup the species to get the the InChi values.

ibell:
I can't tell you how appreciative I am of this work;  every time i would see the non-standard INChI fragments it made me uncomfortable.  I am going to make a merge commit with this unless you think I should squash/merge.  I'm good either way so long  as these commits get into the trunk.

ibell:
After we merge these, I can automatically check against ThermoDataEngine from NIST to ensure that we have all the right INChI keys again.

wandadars:
@ibell Thanks Ian. CoolProp's Python module really helped save me a lot of time for evaluating species critical properties, so hopefully this PR saved you some time(We'll see after that ThermoDataEngine check lol).

PR #1738, opened 2018-09-25
Force unloading of REFPROP upon path change

jowr:
### Description of the Change

Added functions to force the loading and unloading of the REFPROP library. This allows you to change the REFPROP version at runtime and not only at times when not REFPROPMixtureBackend instance is alive. 

### Possible Drawbacks

This is not thread-safe, but I guess all of the REFPROP-related is like this.

### DIFF
--- a/src/Backends/REFPROP/REFPROPMixtureBackend.cpp
+++ b/src/Backends/REFPROP/REFPROPMixtureBackend.cpp
@@ -140,7 +140,7 @@ std::string get_REFPROP_HMX_BNC_path()
 
 
 namespace CoolProp {
-    
+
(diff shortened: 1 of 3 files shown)


jowr:
What about this one? https://github.com/CoolProp/CoolProp/pull/1738/files?w=1

ibell:
Aha, that link is much better.  Is there a need for the whitespace changes
in general?

jowr:
Well, yes and no. Coding style says the spaces should be removed, but no, they do not hurt. If you prefer to keep them, I could modify the PR.

ibell:
What happens to already-existing copies of AbstractState wrapped versions of REFPROP?  Do they get super confused when the force unload happens?

PR #1746, opened 2018-10-18
Fix Python print

scls19fr:
### Requirements

Print is now a function since Python 3.0

### Description of the Change

```
2to3 -f print -n -l -w .
```

and some manual fixes

### Benefits

Should works both with Python 2.7 and 3.x

### Possible Drawbacks

Maybe additional spaces for 

```
print x, y
```

which are changed automatically to

```
print(x, y)
```

instead of manually fixing to

```
print("%s %s" % (x, y))
```

other possible drawback:

this fix will remove possible compatibility with very old Python 2.x version
(ie version which doesn't support print as a function)

other possible drawback:

with Python 2.7

```
print()
```

prints

```
()\n
```

but with Python 3.x

```
\n
```

so changing manually to

```
print("")
```

but I may have miss some

### Verification Process

Run "some" scripts (but not all)

### Applicable Issues

Related #1612


ibell:
I very much appreciate the fixing of this - can you also add
```
from __future__ import print_function
```
to the top of all scripts that you modified?  That will fix the print function in python 2.7.  Until it reaches end-of-life, python 2.7 is still a thing.

scls19fr:
Sorry but I'm not sure that I can do this automatically as there are some `.py` files that can have such an import.
I must admit that I'm not a `grep` / `sed` expert.
Anyway I don't think that printing sometimes not perfectly things (only with Py2.7) is something really annoying compared to the fact that all modified scripts are (were) broken with Py3.x.

ibell:
Fair enough - most of the scripts are not actually part of CoolProp anyhow,
but are/were used for development purposes.  And most of that has
historically at least been done in 2.7, but I'm finally leaving that all
behind.

scls19fr:
I think it's more "future" oriented to watch at diff and fix manual print calls that need to be fixed.

dev/Tickets/524.py
```diff
@@ -2,42 +2,42 @@
 from CoolProp.CoolProp import PropsSI
 from CoolProp.CoolProp import set_reference_state
 
-print "CoolProp version",CoolProp.__version__
-print "CoolProp revision",CoolProp.__gitrevision__
+print("CoolProp version",CoolProp.__version__)
```
scls19fr: This print call for example need to be fixed manually

dev/Tickets/524.py
```diff
@@ -2,42 +2,42 @@
...
+print("CoolProp revision",CoolProp.__gitrevision__)
```
scls19fr: also this one

ibell:
I'm inclined to say we merge this PR, but I wish you had only changed the
print statement to a print function and made no other changes; that's a
change that I would whole-heartedly endorse.  I'm not so sure about all
these modified string insertion updates, I wish they had been left
untouched since they are valid Python code.

dev/TTSE/speed_test_TTSE_BICUBIC.py
```diff
@@ -30,4 +30,4 @@ def time_check(N, h, p, TTSE = False, mode = 'TTSE'):
     time_TTSE, value_TTSE = time_check(N, h, p, TTSE = True)
     time_BICUBIC, value_BICUBIC = time_check(N, h, p, TTSE = True, mode='BICUBIC')
 
-    print k, h, p, (value_TTSE/value_EOS-1.0)*100, (value_BICUBIC/value_EOS-1.0)*100, time_EOS/time_TTSE, time_EOS/time_BICUBIC
+    print(k, h, p, (value_TTSE/value_EOS-1.0)*100, (value_BICUBIC/value_EOS-1.0)*100, time_EOS/time_TTSE, time_EOS/time_BICUBIC)
```
scls19fr: This one need to be fixed

thorade:
This looks all good to me, but I only looked at it, did not run any tests or similar.
I think the from future imports can be added using the futurize script (it is based on 2to3).

dev/pseudo-pure/fit_pseudo-pure_eos.py
```diff
@@ -358,7 +358,7 @@ def OBJECTIVE(self, N):
         residuals = np.r_[(PPF.p/self.p-1),(PPF.cv/self.cv-1),(PPF.cp/self.cp-1)]#,(PPF.w**2/self.speed_sound**2-1)]
         RMS = np.sqrt(np.mean(np.power(residuals, 2)))
 
-        print 'RMS:',RMS*100, '% Max',np.max(np.abs(residuals))*100,'%'
+        print('RMS: %s %% Max %s %' % (RMS*100, np.max(np.abs(residuals))*100))
```
scls19fr: Forget to escape second percent

dev/genetic.py
```diff
@@ -287,8 +287,8 @@ def run(self):
 
         self.fitness(samples[0])
 
-        print self.value
-        print '// Max error is ',samples[0].max_abserror,'% between',np.min(self.T),'and',np.max(self.T),'K'
+        print(self.value)
+        print('// Max error is ',samples[0].max_abserror,'% between',np.min(self.T),'and',np.max(self.T),'K')
```
scls19fr: an other one!

dev/scripts/check_mole_masses.py
```diff
@@ -10,4 +10,4 @@
         m += masses[el]*int(c)/1000.0
     err = m/CoolProp.CoolProp.PropsSI(fluid,"M")-1
     if abs(err) > 1e-3:
-        print(fluid, m/CoolProp.CoolProp.PropsSI(fluid,"M")-1)
+        print((fluid, m/CoolProp.CoolProp.PropsSI(fluid,"M")-1))
```
scls19fr: should be fixed also

dev/scripts/derivative_tester.py
```diff
@@ -39,38 +39,38 @@ def f(index, dx):
 rho = 1.5
 H = Props('H','T',T,'D',rho,fluid)
 P = Props('P','T',T,'D',rho,fluid)
-print T,rho,H,P
+print(T,rho,H,P)
```
scls19fr: same here

dev/scripts/derivative_tester.py
```diff
@@ -83,137 +83,137 @@ def f(dx):
     return DerivTerms('dphir_dDelta',T,rho,fluid)
 
 
-print 'd2phir_dDelta2'
-print (f(-2*dx)-8*f(-dx)+8*f(dx)-f(2*dx))/(12*dx)
-print DerivTerms('d2phir_dDelta2',T,rho,fluid)
+print('d2phir_dDelta2')
+print((f(-2*dx)-8*f(-dx)+8*f(dx)-f(2*dx))/(12*dx))
```
scls19fr: not sure about what 2to3 did here

scls19fr: that's ok sorry for noise

dev/scripts/fit_ancillary_ODRPACK.py
```diff
@@ -36,7 +36,7 @@ def saturation_density(Ref, ClassName, form = 'A', LV = 'L', perc_error_allowed
         pc = Props(Ref,'pcrit')
         rhoc = Props(Ref,'rhocrit')
         Tmin = Props(Ref,'Tmin')
-        print Ref,Tmin,Props(Ref,'Ttriple')
+        print(Ref,Tmin,Props(Ref,'Ttriple'))
```
scls19fr: to be fixed

dev/scripts/fit_ancillary_ODRPACK.py
```diff
@@ -86,7 +86,7 @@ def saturation_density(Ref, ClassName, form = 'A', LV = 'L', perc_error_allowed
 
     max_abserror = 0
     while len(n) > 3:
-        print max_abserror, len(n)
+        print(max_abserror, len(n))
```
scls19fr: same

scls19fr:
This is ready on my side! (I think)

wrappers/Python/CoolProp/__init__.py
```diff
@@ -10,7 +10,7 @@
         os.remove(constants.__file__)
         print("constants shared library has been removed.  Please restart your python code")
     except:
-        print("Unable to remove" + constants.__file__ + ". Please manually remove it")
```
ibell: Can you please revert the changes in this file?

scls19fr: What's wrong here?

ibell:
There was nothing wrong with the original, so I'd prefer to not touch any
file that was fine to begin with.  Makes me nervous, and its an unnecessary
clutter to the diff.

scls19fr:
I think the main problem is that CoolProp should have unit tests to run all these scripts.
Moreover there is so many differents way of using `print` in all these scripts!
We can find probably every way of doing... except f-strings (which are only for Python 3.6 and above).
- “Old-school” String Formatting with `%s`
- concatenation with `str` calls
- `str.format()`
- ...

scls19fr:
File reverted

ibell:
These scripts are mostly not used any more, so none of these are actually
terminal problems with CoolProp, and unit testing is a bit overkill as they
shouldn't actually run regularly.  But I agree that it would be good to fix
print function everywhere

ibell:
Here we go!

PR #1747, opened 2018-10-18
Update LibreOffice wrapper and documention

spinnau:
### Description of the Change

The LibreOffice wrapper is changed to use the CoolProp python package on Windows and Linux. The LibreOffice Windows installers contains a bundled python interpreter, so it is easy to set up. The necessary python script and the Basic macros to wrap the underlying python functions are directly integrated into a new ods template file that also contains some basic examples.

The documentation is updated to reflect the changes in the installation guide and usage instructions.

### Benefits

The same template file can be used on Linux and Windows. The necessary python script for LibreOffice is now directly bundled in the ods template file. Thus, there is no need anymore to install that python script in the user directory.

### Verification Process

I have tested the new wrapper with LibreOffice 6.1.2.1 on Arch Linux and with LibreOffice 6.1.2.1 64bit (bundled with python 3.5.5) on Windows 10.


ibell:
Thanks! This is a nice addition!

Issue #1748, opened 2018-10-19
Apostrophe should be escaped in '...' strings or be used in "..." string


scls19fr:
https://github.com/CoolProp/CoolProp/blob/5045c963222bff7ecfffaace63487e16b4d43ffe/wrappers/Python/examples/example.py#L87

ibell:
Better yet, just use double-quotes outside, fix welcomed

PR #1750, opened 2018-10-19
Python / Code formatting / PEP8

scls19fr:
Following #1748
Related #1254 #1612

Running: https://black.readthedocs.io with default settings

```
$ black --version
black, version 18.6b1
$ black wrappers/Python/
```


ibell:
I think black is too extreme in its code formatting in my opinion.  I started looking into that for another project i develop, and I am not a fan of some of the changes that it makes, so I stopped using black

I endorse a tool that gently pushes for PEP 8 compliance, but allows us flexibility when warranted.

PR #1765, opened 2018-12-05
Patch PropsSI for INCOMP fluid with zero mass fraction

### LINKED ISSUE #1756: Issue with incompressible fluid in v6.2.1
Calling

`PropsSI("H", "T", 325.15, "P", 3800, "INCOMP::LiBr[0]")`

throws an exception on the latest version (6.2.1):

**System.ApplicationException
  HResult=0x80131600
  Message=Initialize failed for backend: "INCOMP", fluid: "" fractions ""; error: key [] was not found in string_to_index_map in JSONIncompressibleLibrary**

On v6.1.0, it returns

`67181.93463102358`

Tested with the C# version on Windows and Darwin 64-bit.

henningjp:
### Description of the Change

Prior to 6.2.0, an error sometimes occurred if a user defined mixture with many components contained a component with a zero (0) mass fraction. PR #1628 modified PropsSI() to parse out any zero mass fraction components from the mixture fluid string and calculate the mixture based on the remaining components.  This had an adverse effect on incompressible fluids which can used the same [] syntax to specify the solute mass/volume fraction in water, which is allowed to be zero.  The change in PR#1628 would then strip out the incompressible fluid leaving an empty and invalid fluid string as reported in Issue #1756.

Change was to prevent a zero mass fraction component from being stripped if there is only one component in the fluid list (as is the case for incompressible fluids).  The alternative was to check the backend in use for "INCOMP", however this is a little rigid if other backends are added.

### Benefits

Single fluid components with a specified mixture fraction of [0] will not be stripped from the fluid list, allowing incompressible fluids to again specify a zero mass/volume fraction.

### Possible Drawbacks

Checks were made with all of the other backends to make sure they all behave as expected.  HEOS and REFPROP mixtures with a single component, although odd, are restored to the original behavior, even if the mixture fraction is zero.
mono-mixture

### Verification Process

Tested incompressible fluid behavior in both Mathcad and Python to ensure that a zero mass/volume fraction again returns a value and does not throw an error.

incompressible

Tested normal user defined mixtures in both HEOS and REFPROP backends using both the Mathcad and Python wrappers to ensure mixture behavior was unaltered.

mixtures

### Applicable Issues

Closes #1756

### DIFF
--- a/src/CoolProp.cpp
+++ b/src/CoolProp.cpp
@@ -168,7 +168,8 @@ std::string extract_fractions(const std::string &fluid_string, std::vector<doubl
                 throw ValueError(format("fraction [%s] was not converted to a value between 0 and 1 inclusive", fraction.c_str()));
             }
 
-            if (f > 10*DBL_EPSILON)  // Only push component if fraction is positive and non-zero
+            if ((f > 10*DBL_EPSILON) ||  // Only push component if fraction is positive and non-zero
+                (pairs.size() == 1))     // ..or if there is only one fluid (i.e. INCOMP backend )
             {
                 // And add to vector
                 fractions.push_back(f);


henningjp:
@ibell, @jowr, If how I did this is OK with you, I'll merge it.

Issue #1777, opened 2019-01-04
Building from the PyPI sdist on Python 3.7 results in compilation errors on macOS


bryanwweber:
### Description

After downloading the sdist of 6.2.1 from PyPI (BTW, there is no tag nor release on GH for 6.2.1), trying to build the package using Python 3.7 by running

    MACOSX_DEPLOYMENT_TARGET=10.9 python setup.py install

results in

```
gcc -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -I/Users/bryan/miniconda3/include -arch x86_64 -I/Users/bryan/miniconda3/include -arch x86_64 -I. -I./include -I./src -I./externals/Eigen -I./externals/fmtlib -I./externals/msgpack-c/include -I/Users/bryan/miniconda3/include/python3.7m -c CoolProp/CoolProp.cpp -o build/temp.macosx-10.7-x86_64-3.7/CoolProp/CoolProp.o
CoolProp/CoolProp.cpp:65531:21: error: no member named 'exc_type' in '_ts'
    *type = tstate->exc_type;
            ~~~~~~  ^
CoolProp/CoolProp.cpp:65532:22: error: no member named 'exc_value' in '_ts'; did you mean 'curexc_value'?
    *value = tstate->exc_value;
                     ^~~~~~~~~
                     curexc_value
/Users/bryan/miniconda3/include/python3.7m/pystate.h:240:15: note: 'curexc_value' declared here
    PyObject *curexc_value;
              ^
CoolProp/CoolProp.cpp:65533:19: error: no member named 'exc_traceback' in '_ts'; did you mean 'curexc_traceback'?
    *tb = tstate->exc_traceback;
                  ^~~~~~~~~~~~~
                  curexc_traceback
/Users/bryan/miniconda3/include/python3.7m/pystate.h:241:15: note: 'curexc_traceback' declared here
    PyObject *curexc_traceback;
...
```
(output truncated for space).

I can successfully build the same commit from a git checkout of the source tree and using Cython, so I presume the problem has something to do with not using Cython in the sdist. Is there a reason not to include the *.pyx files in the sdist?

### Versions

**CoolProp Version:** 6.2.1
**Operating System and Version:** macOS 10.14.2 Apple LLVM version 10.0.0 (clang-1000.10.44.4)
Target: x86_64-apple-darwin18.2.0

P.S. 
[...]

PR #1787, opened 2019-01-12
Add conda install instructions from conda-forge

bryanwweber:
### Description of the Change

As discussed in #1777, this PR introduces instructions to install the Conda package from the conda-forge channel.

### Benefits

This installation method supports those users with Conda distributions of Python.

### Possible Drawbacks

Conda-forge may decide to stop maintaining the package, resulting in these instructions becoming out of date.

### Verification Process

I did not do any verification of this change.


ibell:
Thanks, looks fine

PR #1791, opened 2019-01-15
Add isentropic expansion coefficient

jowr:
I added a shorthand function for the isentropic expansion coefficient. Sorry for the whitespace mess, please visit https://github.com/CoolProp/CoolProp/pull/1791/files?w=1 for a diff without whitespace changes.


ibell:
Looks good to me!  Can you perhaps run through all the C++ files and clean up whitespace so that the whitespace problem goes away in future PR?  I did some of that with the Python files already.

Issue #1781, opened 2019-01-10
C++11 compilation


ibell:
@jowr - I'm thinking to start adding C+11 as a hard requirement.  We'll need a major release for this.  It will by definition drop support for Python 2.7 compiled with visual studio 2008 which does not support C++11. But it will allow us to integrate a lot of the new things I have been working on:

* Chebyshev rootfinding for mixtures: https://www.sciencedirect.com/science/article/pii/S0378381218301730
* Psychrometric routines for humid air: https://www.sciencedirect.com/science/article/pii/S014070071730381X
* In-Memory copies of REFPPROP: https://github.com/usnistgov/REFPROP-manager

My tests with the mingwpy package on windows are very promising (they work fine).  We'll need to make sure that compilation with mingwpy is successful on windows, and that we can compile dependent libraries (PDSim) that use cython, but I have positive experience thus far with both.

jowr:
Thumbs up for this! I am more than happy move forward. However, may of our internal scripts are written in Python2 is that going to result in some funny surprises? What about our tests, are they all Python3?

ibell:
I think the changes will be relatively minor, so I'n not too concerned.  But a safe step #1 would be to ensure that all scripts run with python 3.6

jowr:
So I guess we can start to run all the buildbots with Python 3.6 - Maybe we can also rework the buildbot configuration and integrate TravisCI, AppVeyor and Azure DevOps for tests and deployment. I guess we could benefit from taking some load off the build machines and it would allow us to test different OSX version without messing with XCode and compiler flags...

ibell:
Closing as part of an issue-tracker cleanup — this issue has had no activity in over two years.

*To be transparent: this comment and the close action are being performed by Claude (Anthropic's AI coding assistant), running on behalf of @ibell. Ian reviewed and approved the bulk-close criteria; I am the one executing the action.*

If you would like this issue reconsidered, please feel free to **re-open it** — a fresh comment with current behaviour against the latest CoolProp release is the most useful way to revive the discussion.

— Claude (acting for @ibell)

PR #1810, opened 2019-03-24
Use std::shared_ptr if available in VS

beutlich:
### Requirements

### Description of the Change

Fix compilation with VS2017, where std::tr1 is no longer enabled by default

### Benefits

Compilation with VS2017 works flawlessly. `std::shared_ptr` should be used if available.

### Possible Drawbacks

Not known.

### Verification Process

Compilation works.

### Applicable Issues

Somehow related to #1781, but does not check for the presence of a C++11 compiler, but the first VS version (= 2008) supporting `std::shared_ptr`.

### DIFF
--- a/include/crossplatform_shared_ptr.h
+++ b/include/crossplatform_shared_ptr.h
@@ -42,7 +42,12 @@
     using namespace std::tr1;
 #elif defined(__ISWINDOWS__) && !defined(__MINGW32__)
     #include <memory>
-    using namespace std::tr1;
+    // VS2008 has std::shared_ptr from C++11 
+    #if defined(_MSC_VER) && _MSC_VER >= 1500
+        using std::shared_ptr;
+    #else
+        using namespace std::tr1;
+    #endif
 #else
     #pragma error
 #endif


ibell:
Works for me... Looking forward to the day that we can drop support for old-school C++

ibell:
This causes failure on old VS2008: http://www.coolprop.dreamhosters.com:8010/builders/Python-binaries-windows-32bit-py27/builds/2063/steps/shell_1/logs/stdio  Is the preprocessor macro handling not correct?

ibell:
I think it should be _MSC_VER > 1500 rather than _MSC_VER >= 1500

beutlich:
Hm, strange. I actually took it from the textbook "C++17 Praxiswissen zum neuen Standard" by Peter Pohmann, that `std:shared_ptr` is available by MSVC >= 2008 (and GCC >= 4.8). Since I have no VS2008 available, I tested in VS2010, where ist is available. Thus, I'd recommend to change it to _MSC_VER >= 1600 then.

ibell:
Ok, I bumped to checking for ``_MSC_VER >= 1600`` in 3c07cf89be423ec6b60605185a0d6f7b7919ae2c

PR #1813, opened 2019-04-01
Added limits to the humid air properties, closes #1811

### LINKED ISSUE #1811: Humid air properties above 388 K
Dear developers,

I`m using CoolProp high-level functions through Scilab and a problem emerged. According to the web site the properties of humid air are valid within -143.15 and 350 °C. However, the solutions for temperature close to the limits do not make sense. For instance, when the pressure is 101.33kPa and the temperature is 200K, the calculated enthalpy is -73574. On the other end, when the temperature is above 390K the enthalpy value also becomes negative. I hope you could help me figure out what is happening. 

Best regards,
Gustavo Bonolo de Campos

**Access Method:** Scilab wrapper

jowr:
### Description of the Change

I introduced checks for the limits mentioned in https://doi.org/10.1080/10789669.2009.10390874 

Use https://github.com/CoolProp/CoolProp/pull/1813/files?w=1 to ignore whitespace changes

### Benefits

Calls that are out of bounds return an error and no longer give nonsense values.

### Possible Drawbacks

Calls that are just slightly out of bounds will fail, but they might have worked before.

### Applicable Issues

Closes #1811


ibell:
This looks good, and is a good idea that I should have done long ago.  Conceptually I am struggling to see how one could run into trouble with this in the first place, but I see the point about the partial pressure being greater than the total pressure being a potential problem.

ibell:
One thing I was wondering about, should we skip the error check if the inputs are out of range but still technically possible?  A negative mole fraction for instance would be an impossible state.

PR #1823, opened 2019-04-23
Issues/1820

### LINKED ISSUE #1820: Humid air example fails due to new limits
The recently introduced limits for the HAPropsSi cause the examples to fail ungracefully. This should be fixed be fore we release 6.3.0.

ibell:
closes #1820

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,19 @@ language:
 
 matrix:
   include:
+    - compiler: clang
+      branches:
+        only:
+          - issues/1820
+      before_script:
+        - cmake --version
+        - echo "$CXX" && "$CXX" -v
+        - mkdir -p build && pushd build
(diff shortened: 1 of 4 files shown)


ibell:
This works perfectly now: 
Exceptions: 0 / 51894 = 0%
Bad Accuracy: 0 / 51894 = 0%

And much faster too now that I fixed some stupidity about dewpoint inputs.

If you want to run the tests again, that might be wise, but it should be 100% pass rate

jowr:
Great! I am going to fix the Travis builds and add some results to the humid air benchmark code.

jowr:
I changed the T limits and there are still some regions with problems. However, the overall performance and stability has improved a lot! Great work.

See https://travis-ci.org/CoolProp/CoolProp/jobs/523406260 for test logs.

ibell:
I think I'm done with this for now; these humid air routines are a massive time black home.  Does the model even make sense at 130 K and 373 K?  I'd say go ahead and merge.

Issue #1841, opened 2019-07-16
Doxygen: Fundamental derivative of gas dynamics


DGSEM:
Dear all,
i only want to make a note, that the third term in the formula is wrong:

http://www.coolprop.org/_static/doxygen/html/class_cool_prop_1_1_abstract_state.html#a6116c42705c1d1ae4dac622f998c7302

Regards

ibell:
What's wrong?

DGSEM:
The think the third term in the formule should be written without the 1+..., like

Gamma= v^3/(2*c^2)*(partial p^2 / (partial v^2))

The implementation in Coolprop is fine.

ibell:
Can you manage a pull request to fix this?  It would be a very fine starter
pull request.  Maybe you could fix the partial derivative too at the same
time?

thorade:
Also see equation 1 in https://doi.org/10.1016/j.fluid.2009.11.003

ibell:
Watch out for the units on volume or density, they matter when you use the
form with second derivatives, but cancel when you use the first

PR #1842, opened 2019-07-28
fixes in doxygen docu for fundamental derivative

DGSEM:
See https://github.com/CoolProp/CoolProp/issues/1841


ibell:
Thank you!

PR #1877, opened 2019-12-11
Adding PC-SAFT EOS

zmeri:
### Requirements

### Description of the Change

I have added the PC-SAFT equation of state to CoolProp. The code implements also the polar and electrolyte terms of the EOS. The functions implementing the equation of state are mostly in PCSAFTBackend.cpp. I also had to add a separate function for setting the PCSAFT binary interaction parameters since it uses different parameters than the other backends (these functions are named set_mixture_binary_pair_pcsaft and get_mixture_binary_pair_pcsaft). I also added PC-SAFT parameters for a good chunk of compounds and mixtures. I added Catch tests as well.

### Benefits

A new EOS

### Possible Drawbacks

I don't think it should interfere too much with the existing code since it is a separate backend, but there is the chance it could unexpectedly cause bugs when using other backends.

### Verification Process

I wrote Catch tests to check that the PC-SAFT backend gave accurate values. I also added a couple tests to ensure that the functions for setting and getting the PC-SAFT binary interaction parameters worked and that phases were calculated correctly. I compiled CoolProp with the Catch module (cmake .. -DCOOLPROP_CATCH_MODULE=ON) and ran the tests. All the PC-SAFT tests passed. When I initially ran the other CoolProp tests (before I made any changes) 204 of the 38415 assertions failed. The majority of them seemed to be related to the fluid libraries (i.e. "key [ n-Decane] was not found in string_to_index_map in JSONFluidLibrary"). When I ran all the tests afterwards 206 of the 38415 assertions failed. I looked through those that failed, but I could not see any related to the PC-SAFT backend. It seemed they were the same ones that failed before. So as far as I have tested the PC-SAFT backend works correctly.

### Applicable Issues

N/A


ibell:
Wow.....  This is so awesome.  How feasible would it be to revert the whitespace changes?

src/AbstractState.cpp
```diff
@@ -127,14 +127,17 @@ AbstractState * AbstractState::factory(const std::string &backend, const std::ve
 
     std::map<backend_families,shared_ptr<AbstractStateGenerator> >::const_iterator gen, end;
     get_backend_library().get_generator_iterators(f1, gen, end);
+    std::cout << "after get_generator_iterators" << std::endl; // !!! remove
```
ibell: Can you please remove these lines?

include/crossplatform_shared_ptr.h
```diff
@@ -20,8 +20,8 @@
         using std::shared_ptr;
     #endif
 #elif defined(__ISLINUX__) // GCC
-    #include <tr1/memory>
-    using namespace std::tr1;
+    #include <memory>
```
ibell: Are these changes necessary?  Until we switch the full codebase to C++11, I don't want to mess with this.

src/AbstractState.cpp
zmeri: I think I removed them all. At least I went through and tried to delete them all before submitting the pull request. I didn't see this line in the file anymore. Are you still seeing them?

include/crossplatform_shared_ptr.h
zmeri: I could not get the code to compile without making this change. I am on Ubuntu Linux (18.04.3 LTS) with GCC 7.4.0. Should we include an if statement like was done for MacOS that first checks if <tr1/memory> can be included?

ibell: I think you are right that on linux this isn't too serious, and should work.  This was for monkey-patching ancient versions of c++ compilers, shared_ptr is in the standard library now.

src/AbstractState.cpp
ibell: Yeah, I was looking at intermediate commit.

ibell:
Very nice work!  Thanks a bunch.

PR #1881, opened 2019-12-18
Fixed crash in PTflash_twophase::build_arrays

tobias-loew:
Fixed crash in SaturationSolvers::PTflash_twophase::build_arrays for mixtures with 3 or more components.

### Requirements

### Description of the Change

changed used index

### Benefits

code crashed hard because of wrong index

### Verification Process

Please check, if numerical properties of algorithm are now correct

*Describe the actions you performed (e.g. text you typed, commands you ran, etc.), and describe the results you observed.*  

*Please attach here any python test sessions or image captures of testing in other wrappers.* ]


tobias-loew:
THere is still the problem in src/Backends/Helmholtz/FlashRoutines.cpp that with phase-envelope built ONLY single phase calculation are supported. This seems weird because WITHOUT phase-envelope two-phase calculation are at least tried!

PR #1883, opened 2019-12-19
Testing pcsaft

zmeri:
### Requirements

### Description of the Change

Renamed entropy, enthalpy, and Gibbs energy functions for PC-SAFT because they only calculate the residual properties.

### Benefits

Clarity

### Possible Drawbacks

[ *What are the possible side-effects or negative impacts of the code change?* ]

### Verification Process

I ran the unit tests. The PC-SAFT unit tests passed (except for the entropy test, which has a separate issue that needs to be fixed), and the other test that failed did not seem related to the changes made.

### Applicable Issues

[ *Enter any applicable Issues here.  Use ``Closes #????`` if this PR closes an open issue.* ]


src/AbstractState.cpp
```diff
@@ -548,6 +552,10 @@ double AbstractState::hmolar(void){
     if (!_hmolar) _hmolar = calc_hmolar();
     return _hmolar;
 }
+double AbstractState::hmolar_residual(void){
```
ibell: also nice

src/AbstractState.cpp
```diff
@@ -556,6 +564,10 @@ double AbstractState::smolar(void){
     if (!_smolar) _smolar = calc_smolar();
     return _smolar;
 }
+double AbstractState::smolar_residual_trho(void){
```
ibell: I would remove this function

src/AbstractState.cpp
```diff
@@ -407,12 +407,14 @@ double AbstractState::keyed_output(parameters key)
...
         return hmolar();
+    case iHmolar_residual:
+        return hmolar_residual();
     case iHmass:
         return hmass();
     case iSmolar:
         return smolar();
-    case iSmolar_residual:
-        return gas_constant()*(tau()*dalphar_dTau() - alphar());
+    case iSmolar_residual_trho:
```
ibell: This should go back to smolar_residual

src/AbstractState.cpp
```diff
@@ -421,6 +423,8 @@ double AbstractState::keyed_output(parameters key)
         return umass();
     case iGmolar:
         return gibbsmolar();
+    case iGmolar_residual:
```
ibell: nice! Good idea

src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.h
```diff
@@ -290,31 +290,34 @@ class HelmholtzEOSMixtureBackend : public AbstractState {
...
+    CoolPropDbl calc_gibbsmolar_residual(void) { return gas_constant()*_T*(alphar() + delta()*dalphar_dDelta()); }
     CoolPropDbl calc_gibbsmolar_nocache(CoolPropDbl T, CoolPropDbl rhomolar);
 
     CoolPropDbl calc_helmholtzmolar(void);
     CoolPropDbl calc_cpmolar_idealgas(void);
     CoolPropDbl calc_pressure_nocache(CoolPropDbl T, CoolPropDbl rhomolar);
     CoolPropDbl calc_smolar(void);
     CoolPropDbl calc_smolar_nocache(CoolPropDbl T, CoolPropDbl rhomolar);
-    
+    CoolPropDbl calc_smolar_residual_trho(void) { return gas_constant()*(tau()*dalphar_dTau() - alphar()); }
```
ibell: Yup, this is the right approach for the default implementation

src/DataStructures.cpp
```diff
@@ -40,7 +40,9 @@ const parameter_info parameter_info_list[] = {
     {iCvmass,            "Cvmass",            "O", "J/kg/K",  "Mass specific constant volume specific heat",     false},
     {iCp0molar,          "Cp0molar",          "O", "J/mol/K", "Ideal gas molar specific constant pressure specific heat",false},
     {iCp0mass,           "Cp0mass",           "O", "J/kg/K",  "Ideal gas mass specific constant pressure specific heat",false},
-    {iSmolar_residual,   "Smolar_residual",   "O", "J/mol/K", "Residual molar entropy (sr/R = tau*dar_dtau-ar)",false},
```
ibell: Please put this back, but maybe write sr/R = s(T,rho)-s^0(T,rho) instead of what I did

ibell:
Nice work!!!!

Can you please make some changes to the smolar_residual handling as I indicated by the comments?  I should have been more explicit, but all residual properties in CoolProp/REFPROP are always defined as X^r(T,rho) = X(T,rho) - X^{(ig)}(T,rho), so I don't think we should fiddle with the public residual methods in the AbstractState.  I would fully endorse updating the docs to make clear that this is what is actually calculated for smolar_residual.  It would be good though to do what you started and make smolar_residual a generic method that calls calc_smolar_residual, and then calc_smolar_residual can have a default implementation that can be overwritten by other backends (e.g. PC-SAFT).  Any changes to the abstract base class needs to have some discussion before those changes are made, since it might have some surprising side-effects.

ibell:
But adding the residual methods as you did is a good idea, can you also add the default implementations?

ibell (changes requested):

src/AbstractState.cpp
zmeri: Good point, I will fix that.

src/DataStructures.cpp
zmeri: will do

ibell:
I think this PR can be closed, right?

zmeri:
Yes, that's right. I copied all the changes over to that new PR. Thanks!

PR #1885, opened 2019-12-25
fixed typo in getos() and else statement

tormec:
Fixed a couple of typos that prevent Scilab to run the Coolprop's library under Linux.


ibell:
Thanks!  Merging!

In general, please don't change whitespace (hard to tell what has actually change), and one theme per PR (issues with the typos should be another issue)

PR #1891, opened 2020-01-27
Fix uppercase-only fluid naming

alex-s-v:
### Description of the Change

It was annoying specifying all fluid names in caps when adding new fluids to the cubic backend. Simply removing the uppercasing when searching for the fluid in the database is not a good idea cause of the need to specify all possible name spellings in the aliases. Now all fluid names adding to the database uppercased and aliases stay the same.

### Benefits

No need to uppercase fluid names anymore.

### Possible Drawbacks

I can't come up with any.

### Verification Process

I just tried to add new fluid.

```python
import json


fluid_data = {
    'CAS': '79-01-6',
    'Tc': 571.0,
    'Tc_units': 'K',
    'acentric': 0.217,
    'aliases': ['tce', "TCE"],
    'molemass': 0.131387,
    'molemass_units': 'kg/mol',
    'name': 'trichloroethylene',
    'pc': 4910000.0,
    'pc_units': 'Pa'
}

CP.add_fluids_as_JSON("PR", json.dumps([fluid_data]))
T1 = CP.PropsSI("T", "P", 101325, "Q", 0,"PR::Trichloroethylene")
T2 = CP.PropsSI("T", "P", 101325, "Q", 0,"PR::tce")
T3 = CP.PropsSI("T", "P", 101325, "Q", 0,"PR::TCE")
print(T1 == T2 == T3)  # True
print(T3)  # 359.23423445653344
```

### Applicable Issues

I don't know any.

### DIFF
--- a/src/Backends/Cubics/CubicsLibrary.cpp
+++ b/src/Backends/Cubics/CubicsLibrary.cpp
@@ -51,11 +51,11 @@ public:
                 val.alpha0 = JSONFluidLibrary::parse_alpha0((*itr)["alpha0"]);
             }
             std::pair<std::map<std::string, CubicsValues>::iterator, bool> ret;
-            ret = fluid_map.insert(std::pair<std::string, CubicsValues>(val.name, val) );
+            ret = fluid_map.insert(std::pair<std::string, CubicsValues>(upper(val.name), val) );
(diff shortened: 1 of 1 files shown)


ibell:
This is an example of an excellent PR.  Thank you!

I'm still of the opinion that it would have overall been better to properly fix it by handling case correctly, but this solution is much better than no solution at all.

Issue #1876, opened 2019-12-09
Python 3.8 : Error in import


frisbee-w:
### Prerequisites

Checked.
    * Not reported in [CoolProp User Group](https://goo.gl/Pa7FBT)
    * Similar to #1872
    * Checked with the latest release 6.3.0 from [PyPI](https://pypi.org/project/CoolProp) and [development version](http://www.coolprop.org/dev) 6.3.1 (as suggested in https://github.com/CoolProp/CoolProp/issues/1872#issuecomment-560970428)

### Description

Hi,

I've tried installing the development build. It gets installed, but when importing the module, get the following.
```shell
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/user0/.local/lib/python3.8/site-packages/CoolProp/__init__.py", line 22, in <module>
    __fluids__ = CoolProp.get_global_param_string('fluids_list').split(',')
  File "CoolProp.pyx", line 287, in CoolProp.CoolProp.get_global_param_string
  File "stringsource", line 15, in string.from_py.__pyx_convert_string_from_py_std__in_string
TypeError: expected bytes, str found
```

### Steps to Reproduce

1. Install CoolProp from PyPI or the development server  
2. Import the module in Python 3.8
```python
import CoolProp.CoolProp as CP
```

**Expected behavior:** Successful import

**Actual behavior:** Error

### Versions

**CoolProp Version:** 6.3.0 or 6.3.1  
**Operating System and Version:** Arch Linux
**Access Method:** Python 3.8    

### Additional Information

None

frisbee-w:
Works with the build from GitHub sources.
Closing the issue.

ibell:
Actually, I think this is abug,

notmyrealhandle:
FWIW unsurprisingly this issues also exists for python 3.8.1 and pypi CoolProp 6.3.0.  

Building CoolProp 6.3.1.dev0 from (github) source does resolve the issue for me.  

I'm not sure its related but If it helps, I found [this](https://github.com/cython/cython/issues/2819) on the cython github issues site - perhaps when cython 3 is released there will be a change.

A few extraneous details
```
(p381.mkl) [160] $ gcc --version
gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
[...]

PR #1896, opened 2020-02-05
Update init.py byte string in split for Py3.8

dailyllama:
Fixes issue https://github.com/CoolProp/CoolProp/issues/1876


ibell:
Thank you!

ibell:
I think I was a bit too quick to accept this, I reverted the change since all the bytes seem no longer needed with up-to-date cython builds.

PR #1906, opened 2020-03-20
Update index.rst

HCSpin:
### Requirements

### Description of the Change

[ Addition of a step in setting up the Excel wrapper]

### Benefits

[ This action was required for me to get CoolProp to work on Excel 2019 for MacOSX. ]

### Possible Drawbacks

[ Unknown ]

### Verification Process

[ I tried locating the .dylib file elsewhere, providing the full path to it, but it only worked if I used the full path indicated in the rest of the installation guide.  It did not work for me w/o the full path.  It would be great if someone else could confirm on another mac.]

### Applicable Issues

[ NA]


ibell:
Thank you!

PR #1921, opened 2020-04-20
fix typos in pcsaft json

zmeri:
### Description of the Change

Two compounds were misnamed in the file containing PC-SAFT interaction parameters (mixture_binary_pairs_pcsaft.json). "HEPTENE" and "OCTENE" should be "1-HEPTENE" and "1-OCTENE". The .so pattern was also missing from .gitignore.

### Benefits

Now the names here match with those in the json file containing pure compound PC-SAFT parameters.

### Possible Drawbacks

I guess there's always the chance I accidentally added more typos...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@
 /wrappers/Python/CoolProp.egg-info
 /wrappers/Python/dist
 /wrappers/Python/CoolProp/*.o
+/wrappers/Python/CoolProp/*.so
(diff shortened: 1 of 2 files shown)


ibell:
Can you look into the resetting of the residual values that get cached?

Also, residual entropy is broken for REFPROP now, but was ok before.

zmeri:
Yeah, I can take a look into those things.

PR #1957, opened 2020-07-12
Update BaseObjects.py

### LINKED ISSUE #1931: Adding New Fluids - all_incompressibles.py
I wanted to add a new fluid as per the instructions on coolprop.org. After running all_incompressibles.py and reinstalling, an error appeared when importing and using PropsSI. The error also occurs without any modification to the source files (i.e. PureFluids.py).

Anaconda Python 3.7.4 on Windows 10

Steps taken:

clone repo and check out v6.3.0
run BuildCP.bat under wrappers/Python

-> no problem

run all_incompressibles.py
run ForceBuild.bat under wrappers/Python

-> problem with incompressible fluids, as follows:

In [1]: from CoolProp.CoolProp import PropsSI
Unable to load fluid: DowJ; error was input "[
    -250.0,
    1.5,
    10.0
]" is not a 2D array
input "[
    -250.0,
    1.5,
    10.0
]" is not a 2D array

In [2]: PropsSI("V","P",101325,"T",273.15,"INCOMP::DowJ")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-8652c2f6633c> in <module>
----> 1 PropsSI("V","P",101325,"T",273.15,"INCOMP::DowJ")

CoolProp\CoolProp.pyx in CoolProp.CoolProp.PropsSI()

CoolProp\CoolProp.pyx in CoolProp.CoolProp.PropsSI()

CoolProp\CoolProp.pyx in CoolProp.CoolProp.__Props_err2()

ValueError: Initialize failed for backend: "INCOMP", fluid: "DowJ" fractions "[ 1.0000000000 ]"; error: key [DowJ] was not found in string_to_index_map in JSONIncompressibleLibrary : PropsSI("V","P",101325,"T",273.15,"INCOMP::DowJ")

whunter7:
### Description of the Change

all_incompressibles.py was not completing properly due to a type error - a none integer was being passed to np.zeros((xLen, yLen)).

### Benefits

all_incompressibles.py completes properly

### Possible Drawbacks

none I can see

### Verification Process

Successfully followed documented procedure Adding New Fluids

### Applicable Issues

``Closes #1931``

### DIFF
--- a/dev/incompressible_liquids/CPIncomp/BaseObjects.py
+++ b/dev/incompressible_liquids/CPIncomp/BaseObjects.py
@@ -139,8 +139,8 @@ class IncompressibleData(object):
             elif self.type == IncompressibleData.INCOMPRESSIBLE_LOGEXPONENTIAL:
                 xLen = np.round([len(x) / 1.5])
                 yLen = np.round([len(y) / 1.5])
-                xLen = np.min([xLen, 4])
-                yLen = np.min([yLen, 6])
+                xLen = int(np.min([xLen, 4]))
+                yLen = int(np.min([yLen, 6]))
 
                 if (xLen + yLen) > 2:
                     if self.DEBUG: print("Poor solution found with log exponential, trying once more with exponential polynomial.")


ibell:
Should this maybe be a math.floor instead of an int?

whunter7:
That should work too, I think int vs math.floor only would make a
difference if the argument was negative, which, for defining the length of
an array, shouldn't happen.
No rounding operation is actually required, np.min() is operating on the
result of an np.round() call and a hard coded int. Therefore, there are no
decimal places. But the result needs to be made an int type to be used to
size the subsequent array.
Please trust your instinct though, I'm a very part-time hobby programmer at
best.

Issue #1941, opened 2020-05-28
n-D inputs for list are broken


ibell:
@jowr ran into an issue with #1932 when building docs.  Basically, list inputs to PropsSI are now broken :/  I guess they need to be special cased, or the fix needs to be special cased.

```
Executing logo_2014.py
Traceback (most recent call last):
  File "logo_2014.py", line 64, in <module>
    p_sat = CP.CoolProp.PropsSI('P', 'T', T_sat, 'Q', [0] * len(T_sat), 'Water')
  File "CoolProp/CoolProp.pyx", line 373, in CoolProp.CoolProp.PropsSI
  File "CoolProp/CoolProp.pyx", line 414, in CoolProp.CoolProp.PropsSI
AttributeError: 'list' object has no attribute 'shape'
```

ibell:
Another special case, when multiple outputs are provided.  I think 5 keys are provided here.
``` python
Traceback (most recent call last):
  File "D4.py", line 46, in <module>
    RPdata = CP.PropsSI(normalkeys, 'T', T, 'Dmolar', rho, RPfluid)
  File "CoolProp/CoolProp.pyx", line 373, in CoolProp.CoolProp.PropsSI
  File "CoolProp/CoolProp.pyx", line 462, in CoolProp.CoolProp.PropsSI
ValueError: cannot reshape array of size 250 into shape (50,)
```

jowr:
Hi - what would be the desired output here? It looks like my changes were not as well prepared as I thought.

ibell:
I think in this case the returned matrix should be 5 columns by 50 rows.

There are a lot of special cases to worry about that I didn't think about before....

jowr:
I was thinking something along these lines: 

1. Single output and only floats as input -> float
2. Single output and one nD array and one float as input -> nD array shaped like the input array
3. Single output and two nD arrays as input -> nD array shaped like the input arrays, error if shapes do not match
4. List of outputs and only floats as input -> list of floats (list is indexed like the list of outputs)
5. List of outputs and one nD array and one float as input -> list of nD arrays (list is indexed like the list of outputs, array is shaped like the input array)
6. List of outputs and two nD arrays as input -> list of nD arrays (list is indexed like the list of outputs, array is shaped like the input array, error if shapes do not match)

Would that work for you?

jowr:
I assume that my last description is what we want - I'll go ahead and start implementing the changes.

ibell:
[...]

PR #1959, opened 2020-07-28
WIP: n-D inputs for Python 

### LINKED ISSUE #1898: nD data as inputs in Python wrapper
Python array inputs are limited to 1D, shouldn't we extend that to nD arrays?

jowr:
Initially intended to close #1898, but causes trouble as discussed in #1941 .


ibell:
I think I'll close this one; the `AS.fast_evaluate` path is the preferred option going forwards, and opens the doors to potentially multi-threaded operation

Issue #1890, opened 2020-01-25
Nonsymmetric matrix or binary interaction parameters


alex-s-v:
Hello. Is it possible to set binary interaction parameters for EoSs independently? Or add a new function for that purpose?
For example it could be done in the following way:
```
AS.set_binary_interaction_double_nonsymmetric(0, 1, -0.002)  # Sets only 0 to 1 interaction
AS.set_binary_interaction_double_nonsymmetric(1, 0, 0.002)  # Sets only 1 to 0 interaction
# Above resulting in k equal to
# [[0, -0.002]
#  [0.002, 0]] 
```
I've tried to do it myself, but I'm no good in C++ and the build keep crushing on me every time with different errors.

ibell:
That's not possible presently, but a good idea.  Another idea would be to overload the setter function so that by default it would set both entries in the matrix, but a final argument could control whether to also set the partner to the same value (the current default is to do that)

PR #1967, opened 2020-08-31
Asymmetric BIPs

alex-s-v:
### Description of the Change

Some new mixing rules require the matrix of binary interaction parameters (BIPs) to be asymmetric, so I have done my best to add such functionality. Methods `set_binary_interaction_double` were overwritten to accept an additional (optional) parameter which specifies whether to set Kij symmetrically or not (symmetrically by default).

### Benefits

Now you can specify the matrix of BIPs asymmetrically.

### Possible Drawbacks

Adds new ways to inexperienced users to ruin their calculations =D

### Verification Process

The following python script shows the use and effect of the introduced changes.

```python

AS = CP.AbstractState("PR", "benzene&toluene")
AS.set_binary_interaction_double(0, 1, "kij", 0.05)  # symmetrically
k01 = AS.get_binary_interaction_double(0, 1, "kij")
k10 = AS.get_binary_interaction_double(1, 0, "kij")
print(f"k01 = {k01}; k10 = {k10}")
# Output: k01 = 0.05; k10 = 0.05
AS.set_mole_fractions([0.3, 0.7])
AS.update(CP.PQ_INPUTS, 101325, 0)
ys = AS.mole_fractions_vapor()
print(ys)
# Output: [0.5042805536100579, 0.49571944638994214]

AS.set_binary_interaction_double(0, 1, "kij", 0.1, False)  # asymmetrically
k01 = AS.get_binary_interaction_double(0, 1, "kij")
k10 = AS.get_binary_interaction_double(1, 0, "kij")
print(f"k01 = {k01}; k10 = {k10}")
# Output: k01 = 0.1; k10 = 0.05
AS.update(CP.PQ_INPUTS, 101325, 0)
ys = AS.mole_fractions_vapor()
print(ys)
# Output: [0.5050982763042872, 0.49490172369571284]
```

### Applicable Issues
My own old issue #1890

### DIFF
--- a/include/AbstractState.h
+++ b/include/AbstractState.h
@@ -574,6 +574,10 @@ public:
     virtual void set_binary_interaction_string(const std::string &CAS1, const std::string &CAS2, const std::string &parameter, const std::string &value){ throw NotImplementedError("set_binary_interaction_string is not implemented for this backend"); };
     /// Set binary mixture string parameter (EXPERT USE ONLY!!!)
     virtual void set_binary_interaction_string(const std::size_t i, const std::size_t j, const std::string &parameter, const std::string &value){ throw NotImplementedError("set_binary_interaction_string is not implemented for this backend"); };
+    /// Set binary mixture string parameter (EXPERT USE ONLY!!!)
+    virtual void set_binary_interaction_double(const std::string &CAS1, const std::string &CAS2, const std::string &parameter, const double value, const bool symmetric){ throw NotImplementedError("set_binary_interaction_double is not implemented for this backend"); };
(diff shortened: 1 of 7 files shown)


ibell:
I have a new idea: what if rather than adding a new argument, you could add ``:asymmetric`` (or something similar) to the key, so the API could be identical, but the same effect could be realized?  Boolean arguments are always challenging to document because I see a ``True`` in the code, and it is not clear what is going on there.  My proposal should propagate to the linked states for VLE, for instance, and I'm not sure that would necessarily work with what you have implemented.

PR #2005, opened 2021-03-05
Fix cxx17

ssg-aero:
### Requirements

None

### Description of the Change

Fix cxx11 detection

### Benefits

Shall be compilator agnotic

### Possible Drawbacks

Tests feature is not meaningful without comments.

### Verification Process

Compilation was tested successfully with and without -DCMAKE_CXX_STANDARD:STRING=17 cmake flag

### Applicable Issues

None


ibell:
That seems good.  Let's see how the compilers like this solution.

PR #1990, opened 2020-12-10
Fix cxx17

ssg-aero:
### Description of the Change
Trim functions are edited according the quoted stackoverflowpost.
[](https://stackoverflow.com/questions/216823/whats-the-best-way-to-trim-stdstring)
### Benefits

Support c++17

### Possible Drawbacks

Needs a c++11 compiler

PR #2008, opened 2021-03-15
Fix build on macOS

bryanwweber:
### Description of the Change

The changes in #1990 require C++11 compiler support. On macOS 10.14 with
XCode 11.3, the -std=c++11 flag must be specified to the compiler.

I was getting errors when compiling like:

```
clang -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include -I../.. -I../../include -I../../src -I../../externals/Eigen -I../../externals/fmtlib -I../../externals/msgpack-c/include -I/Users/bryan/.pyenv/versions/thermostate-dev/include -I/Users/bryan/.pyenv/versions/3.9.1/include/python3.9 -c ../../src/AbstractState.cpp -o build/temp.macosx-10.14-x86_64-3.9/../../src/AbstractState.o
In file included from ../../src/AbstractState.cpp:14:
In file included from ../../include/AbstractState.h:11:
In file included from ../../include/CachedElement.h:11:
In file included from ../../include/CoolPropTools.h:15:
../../include/CPstrings.h:45:61: error: expected expression
        s.erase(s.begin(), std::find_if(s.begin(), s.end(), [](unsigned char ch) {
                                                            ^
../../include/CPstrings.h:61:52: error: expected expression
        s.erase(std::find_if(s.rbegin(), s.rend(), [](unsigned char ch) {
```

### Benefits

Fix this compiling error on macOS

### Possible Drawbacks

This should only affect macOS, but I haven't tested on newer versions of macOS than 10.14, so I'm not sure if this is required for newer versions of XCode.

### Verification Process

I built the Python wrapper on macOS 10.14, as well as used CMake to build the shared libraries.

### Applicable Issues

#1990 introduced this change to strings.


ibell:
Yup, I think we are finally going to require c++11.

PR #2056, opened 2021-10-26
Fix typo in iQ description

Qi-Chu-woodside:
### Requirements

### Description of the Change

The description of Q in the http://www.coolprop.org/coolprop/HighLevelAPI.html#high-level-api is misleading, where it should be "Molar vapor quality" instead of "Mass vapor quality" given the unit description there and the source code implementation.  This commit is intended to fix that documentation by fixing the description implemented in the source code.

### Benefits

No more confusion when using this property.

### Possible Drawbacks

Null

### Verification Process

Null. One word fix so not tested.

### Applicable Issues

NA.


ibell:
I've merged this PR.  I would propose you implement Qmass as an AbstractState method, implement it, and add to Python interface as next step.  For mixtures, we will have to add all the Qmass inputs as options at some point, but one step at a time.

PR #2097, opened 2022-03-14
Add github action to build python wheels (including python 3.9 and 3.10)

jmarrec:
### Description of the Change

- Fix #1981
- Adds a github action with **cibuildwheel** to build wheels on all platforms
- Setting up some TWINE tokens would enable automatic upload to PyPi (for tags), and TestPyPi (for nightly builds / PR builds)
- **Note**: Right now I cannot get windows x86 to build...

### Benefits

Python 3.9 and 3.10 aren't supported right now.

As far as I can tell there is also no real CI running right now (nor has been for a while) but I could be wrong. At least this is what this comment seem to indicate: https://github.com/CoolProp/CoolProp/issues/1981#issuecomment-1009973498

This sets up github actions to produce new python wheels for all platforms and automatically upload them to PyPi/TestPyPi

I also enabled arm64 mac builds (for M1 macs / Apple silicon)

### Possible Drawbacks

Mac picks ups not `HAS_MOVE_SEMANTICS` but C++11 mode isn't enabled, so it chokes on lambdas. For now I generally enabled C++11, not sure if that's a big deal or not... Here it'll choke on line 45

https://github.com/CoolProp/CoolProp/blob/381c8535e5dec3eec27ad430ebbfff8bc9dfc008/include/CPstrings.h#L33-L50

### Verification Process

Right now I'm just looking at whether the CI build worked or not. I have not conducted extensive testing of whether the resulting wheels are valid or not.

### Applicable Issues

- Fix #1981

PR #2098, opened 2022-03-15
Github Actions: add shared library and doxygen workflows.

jmarrec:
Reacting to a comment on https://github.com/CoolProp/CoolProp/pull/2097#issuecomment-1067484987

@ibell  I honestly don't know if that's what you had in mind or not, and I definitely don't understand what the "Installer package" is supposed to be, is that the `COOLPROP_WINDOWS_PACKAGE` ?

Anyways, this will build the shared lib, and package it up nicely in a version-named tar.gz, uploaded as an artifact.
If you trigger a release (via github, but that effectively creates a tag which is the actual trigger), it will also upload them to the release page.

The HTML version of the documentation is also packaged in a tar.gz and has the same upload treatment. Not sure if that's what you needed either.

---

**Edit**:

- Windows isntaller package, checkout this run: https://github.com/jmarrec/CoolProp/actions/runs/1985940859
```
CoolProp-6.4.2-dev-WindowsInstaller.tar.gz
├── Installers
│   └── Windows
│       └── CoolProp_v6.4.2.0.exe
└── MicrosoftExcel
    ├── CoolProp.xla
    ├── CoolProp.xlam
    └── TestExcel.xlsx

3 directories, 4 files
```

- Checkout this release on my fork, to showcase the upload of artifacts to the release page: https://github.com/jmarrec/CoolProp/releases/tag/v6.4.2-dev

Screen Shot 2022-03-15 at 11 42 35

### DIFF
--- a/.github/workflows/BuildSharedLibrary.yml
+++ b/.github/workflows/BuildSharedLibrary.yml
@@ -58,24 +58,68 @@ jobs:
     #  # See https://cmake.org/cmake/help/latest/manual/ctest.1.html for more detail
     #  run: ctest -C $BUILD_TYPE
 
+
+    - name: Extract OSApp version from CMakeLists.txt
+      shell: python
+      run: |
+        import re
+        from pathlib import Path
+
+        with open('CMakeLists.txt', 'r') as f:
+            content = f.read()
+
+        no_comments_lines = []
+        for line in content.splitlines():
+            l = line.strip().split('#')[0]
+            if l:
+                no_comments_lines.append(l)
+        content = "\n".join(no_comments_lines)
+
+        m_major = re.search(r'set\s*\(COOLPROP_VERSION_MAJOR (\d+)\)', content)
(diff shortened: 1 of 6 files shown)


jmarrec:
I guess perhaps uploading the Windows installer .exe directly to the release page makes more sense?
What about the Excel stuff though?

ibell:
I had in mind the last few shared library options from http://www.coolprop.dreamhosters.com:8010/grid

ibell:
Would it be feasible to build 32-bit DLL? If not, I can live with that - 64-bit is now the standard.

ibell:
Yep, this is the installer package - you're amazing!  Do you want to dig into the C++? We really need someone to.

jmarrec:
@ibell I added a workflow to run the catch2 tests. Many are failling. Some are failing because of missing Refprop. I have tried to built it locally from https://github.com/usnistgov/REFPROP-cmake but it fails to work due to PASS_FNT.FOR when I use the instructions.

PR #2103, opened 2022-03-28
Lint: use automated tooling to reformat C++ and CMakeLists files

jmarrec:
### Requirements

Lint: use automated tooling to reformat C++ and CMakeLists files

### Description of the Change

This sets up config files for clang-format and cmake-format. It also adds a config file for clang-tidy

### Benefits

This ensures nicely formatted files. Note that it's possible to enforce this via a clang-format (or pre-commit hook) action if that's needed.

### Possible Drawbacks

Zero.

### Verification Process

I added a Catch2 runner on https://github.com/CoolProp/CoolProp/pull/2098 so we could compare results before/after. They are the same.


ibell:
I have mixed feelings about the style choices here. I would prefer 4 spaces per indent, and definitely as you did, no tabs. But I also think overall the changes are fine.

.clang-format
```diff
@@ -0,0 +1,99 @@
...
+ConstructorInitializerIndentWidth: 2
+ContinuationIndentWidth: 2
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+DisableFormat: false
+ExperimentalAutoDetectBinPacking: false
+FixNamespaceComments: true
+IncludeBlocks: Preserve
+IndentCaseLabels: true
+IndentWidth: 2
```
jmarrec, line 58: Change this to 4 if that makes you happier. You're the maintainer, pick something you like.

I tend to like 2 spaces better to avoid uncessary line wrapping, but that's my taste. The point is to have a **constant** enforceable style, so it's easy to read.

jmarrec:
@ibell Ok, I cleaned it up, ready for merge. Tests are the same before and after.

ibell:
Ok, here we go!

PR #2104, opened 2022-03-28
Initial review and refactoring of C++ code

jmarrec:
### Requirements

I didn't go everywhere. And I left some `TODO`s (search for TODO) for maintainers.

Note: this **builds upon the `lint`** branch, so merge https://github.com/CoolProp/CoolProp/pull/2103 and the diffs will get cleaned up, as can be seen here: https://github.com/jmarrec/CoolProp/compare/lint...cpp_review

### Description of the Change

Refactor the code where appropriate and review it.

### Benefits

Try to use modern C++. For eg the entire project is using deprecated headers (eg: `stdlib.h` instead of `<cstdlib>`). Lots of range-based for loops can be applied, some move semantics.  Virtual empty destructors were completely unecessary, overuse of C arrays, etc. 

### Applicable Issues

[ *Enter any applicable Issues here.  Use ``Closes #????`` if this PR closes an open issue.* ]

PR #2105, opened 2022-03-28
Bump Catch  1 to Catch v3.0.0-preview4

jmarrec:
### Description of the Change

Bump Catch to Catch 2.x
register catch2 tests as `ctest`

Note: this build upon the `cpp_review` branch, so review https://github.com/CoolProp/CoolProp/pull/2104 or look at https://github.com/jmarrec/CoolProp/compare/cpp_review...catch2 for a clean diff

### Benefits

More modernizing

### Possible Drawbacks

none.

### Verification Process

Ran the Catch2 tests

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,5 +1,5 @@
 [submodule "externals/Catch"]
-	path = externals/Catch
+	path = externals/Catch2
 	url = https://github.com/catchorg/Catch2
 	branch = master
 [submodule "externals/Eigen"]
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1964,21 +1964,27 @@ endif()
 
 ###      COOLPROP TESTING APP       ###
 if(COOLPROP_CATCH_MODULE)
+
+  add_subdirectory("externals/Catch2")
   enable_testing()
 
-  list(APPEND APP_SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/test_main.cxx")
   list(APPEND APP_SOURCES
        "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests.cpp")
 
   # CATCH TEST, compile everything with catch and set test entry point
   add_executable(CatchTestRunner ${APP_SOURCES})
-  add_dependencies(CatchTestRunner generate_headers)
-  set_target_properties(
-    CatchTestRunner PROPERTIES COMPILE_FLAGS "${COMPILE_FLAGS} -DENABLE_CATCH")
+
+  target_link_libraries(CatchTestRunner PRIVATE Catch2::Catch2WithMain)
+  target_compile_definitions(CatchTestRunner PRIVATE ENABLE_CATCH)
   if(UNIX)
     target_link_libraries(CatchTestRunner ${CMAKE_DL_LIBS})
   endif()
-  add_test(ProcedureTests CatchTestRunner)
+
+  target_link_libraries(CatchTestRunner PRIVATE ${COOLPROP_LIBRARY_NAME})
+
+  include(CTest)
+  include(${CMAKE_CURRENT_SOURCE_DIR}/externals/Catch2/extras/Catch.cmake)
+  catch_discover_tests(CatchTestRunner)
 
   if(COOLPROP_IWYU)
     find_program(iwyu_path NAMES include-what-you-use iwyu)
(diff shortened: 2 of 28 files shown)


ibell:
I'm really lost with all these PR.  Can we please keep them granular, so that each PR does one thing (only)? Upgrading Catch is fine, and even better would be to upgrade to Catch3

jmarrec:
@ibell I made **6** different PRs, for the sole purpose of making them do just one thing. I understand this is a lot to take in, but please trust that it would have been far easier to just do everything in one go. I ended up doing interactive rebases a bunch of times so I could split the work into discrete PRs

They are just stacked upon each other right now because it makes sense to be able to compare things (eg: I had a Catch test workflow in another PR, I kinda want to have it so I can compare the catch2 results before / after this update of catch), but they could always be cherry picked / rebased to stand in isolation.

jmarrec:
@ibell PAINFUL rebase, but I think I cleaned it up nicely. It's running locally and producing the same results

.gitmodules
```diff
@@ -1,5 +1,5 @@
 [submodule "externals/Catch"]
-	path = externals/Catch
+	path = externals/Catch2
```
jmarrec, line 2: Renamed the submodule. Locally developpers that already have a repo will need:

```
/bin/rm -Rf externals/Catch*
git submodule update --init --recursive --force
```

CMakeLists.txt
```diff
@@ -1964,21 +1964,27 @@ endif()
...
   list(APPEND APP_SOURCES
        "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests.cpp")
 
   # CATCH TEST, compile everything with catch and set test entry point
   add_executable(CatchTestRunner ${APP_SOURCES})
-  add_dependencies(CatchTestRunner generate_headers)
-  set_target_properties(
-    CatchTestRunner PROPERTIES COMPILE_FLAGS "${COMPILE_FLAGS} -DENABLE_CATCH")
+
+  target_link_libraries(CatchTestRunner PRIVATE Catch2::Catch2WithMain)
```
jmarrec, line 1977: Removing the test_main.cxx and linking to Catch2WithMain

CMakeLists.txt
```diff
@@ -1964,21 +1964,27 @@ endif()
...
+  target_compile_definitions(CatchTestRunner PRIVATE ENABLE_CATCH)
   if(UNIX)
     target_link_libraries(CatchTestRunner ${CMAKE_DL_LIBS})
   endif()
-  add_test(ProcedureTests CatchTestRunner)
+
+  target_link_libraries(CatchTestRunner PRIVATE ${COOLPROP_LIBRARY_NAME})
```
jmarrec: Will need to remove this as it's not needed since I now use APP_SOURCES instead of just using the files with tests in (per @ibell 's request/taste)

CMakeLists.txt
```diff
@@ -1964,21 +1964,27 @@ endif()
...
+
+  include(CTest)
+  include(${CMAKE_CURRENT_SOURCE_DIR}/externals/Catch2/extras/Catch.cmake)
+  catch_discover_tests(CatchTestRunner)
```
jmarrec, line 1985: Add to ctest

dev/generate_headers.py
```diff
@@ -292,9 +292,10 @@ def combine_json(root_dir):
 
 def generate():
 
-    import shutil
-    shutil.copy2(os.path.join(repo_root_path, 'externals', 'Catch', 'single_include', 'catch.hpp'), os.path.join(repo_root_path, 'include', 'catch.hpp'))
```
jmarrec, line 296: Removed

src/AbstractState.cpp
```diff
@@ -1064,7 +1064,7 @@ CoolPropDbl AbstractState::calc_second_partial_deriv(parameters Of1, parameters
 
 #ifdef ENABLE_CATCH
 
-#include "catch.hpp"
+#include <catch2/catch_all.hpp>
```
jmarrec, line 1067: Fixup includes. Technically we could use a more granular header such as `<catch2/catch_test_macros.hpp>`, but I didn't bother (link time isn't such an issue anyways)

src/Backends/Helmholtz/MixtureDerivatives.cpp
```diff
@@ -1321,7 +1321,7 @@ class DerivativeFixture
             }
             CAPTURE(name);
             CAPTURE(i);
-            CAPTURE(analytic)
+            CAPTURE(analytic);
```
jmarrec, line 1324: missing semi colons

.gitmodules
jmarrec: @ibell FYI

ibell: Thanks!

PR #2106, opened 2022-03-28
Cppcheck workflow

jmarrec:
Note: this builds upon https://github.com/CoolProp/CoolProp/pull/2103

The clang-format will run only on PRs, and only on files touched by this PR (to avoid false positives).

It adds a cppcheck one, with a custom formatter I built.

image

The cppcheck one is (logically) failing. Errors should be fixed one by one as possible, the rest should be suppressed (either for entire files or inline), until everything passes. Then you can start enforcing. (This is the same as the Catch2 tests, 100% should be passing if you want them to be useful, so you have a Yes or No answer right away)

https://github.com/jmarrec/CoolProp/runs/5724426268?check_suite_focus=true

### DIFF
--- a/.github/workflows/cppcheck.yml
+++ b/.github/workflows/cppcheck.yml
@@ -1,0 +1,45 @@
+name: cppcheck
+
+on:
+  push:
+    branches: [ master, develop, cppcheck_clang_format ]
+  pull_request:
+    branches: [ master, develop ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container:
+      image: nrel/cppcheck:2.3
+    steps:
+    - uses: actions/checkout@v3
+    - name: Run cppcheck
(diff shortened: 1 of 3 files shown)


ibell:
Nice to add this, same comment as other PR, I am a bit lost, would be nice to have granular PR, and then rebase so that each PR is more clear about what it is actually doing.

jmarrec:
The Lint PR is safe, so I assume it'll be merged eventually. This PR needs the .clang-format file I added in #2103, and I do not want to have to have one in both PRs and keep them in sync. Hence why this PR is there.

I suggest you start by merging #2103 after any adjustments you see fit. Then it will all become a lot clearer.

jmarrec:
Ok, rebased, moved the clang-format workflow on the Lint PR.

PR #2109, opened 2022-03-31
Configure upload to pypi/testpypi

jmarrec:
Configure upload to pypi/testpypi

Figure out the next available version to avoid override (especially for testpypi, shouldn't happen for pypi as we should bump in CMakeLists)


jmarrec:
@ibell I see you fixed the issues on master, thanks for that.

PR #2116, opened 2022-04-11
Fix mac sed

jmarrec:
@ibell mac has a weird sed version, it doesn't behave like the one I have on Ubuntu and git bash... I found the syntax to make it work, but I chose to do the replacement from python instead of using sed to avoid having obscure issues.


jmarrec:
@ibell Did you set the `PYPI_TOKEN` and `TESTPYPI_TOKEN` to a valid token in github secrets?

ibell:
> @ibell Did you set the `PYPI_TOKEN` and `TESTPYPI_TOKEN` to a valid token in github secrets?

Yes, I did, for TESTPYPI at least

PR #2118, opened 2022-04-11
Python bindings upload to (test)pypi fixes

jmarrec:
Trying to get the upload to work. The current one is having problems in the upload step: https://github.com/CoolProp/CoolProp/runs/5977281354?check_suite_focus=true#step:6:8


jmarrec:
@ibell I think with this, once merge it should work. The upload was skipped on this repo because this is a PR.

On my fork, it went to the upload but failed since I do not have a TESTPYPI_TOKEN secret there since I don't have access to the CoolProp package on test.pypi.org

PR #2122, opened 2022-04-12
Simplify CoolProp python bindings cibuildwheel

jmarrec:
I'm opening this because I've done the legwork and we discussed it on a couple of occasions already.

Pros:
* It's less verbose and therefore perhaps simpler to read.

Cons:
* It's slower (about 16minutes versus 8minutes in the current setup)
    * Primarily because one runner will build both x86_64 and win32 for eg
    * because it doesn't use the COOLPROP_CMAKE (which will build the shared library bit in parallel with `-j $(nproc)`)
* I'm pretty sure there is something quite wrong with the linux configs... the whl ends up being 43 MB versus 6MB in the current cibuildwheel setup (and 2MB on mac, 1.7 on windows...), I don't know why. (nor am I interested in figuring it out)
    * **edit**: fixed by putting back COOLPROP_CMAKE for linux

I'm not really in favor of making this "simplification" (edit: now the binary size is fixed, it's a matter of taste)

**edit2:** actually this solves the arm64 while my minimum fix didn't... so that makes me a lot more enclined to swallow the small runtime cost. I would now recommend we go with this one @ibell

### DIFF
--- a/.github/workflows/python_cibuildwheel.yml
+++ b/.github/workflows/python_cibuildwheel.yml
@@ -12,35 +12,21 @@ on:
 jobs:
 
   python_bindings:
-    name: Build wheel for ${{ matrix.os }}
+    name: py${{ matrix.python-version }}-${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       # Ensure that a wheel builder finishes even if another fails
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-2019, macos-latest]
-        include:
-          # Window 64 bit
-          # Note: windows-2019 is needed for older Python versions:
-          # https://github.com/scikit-learn/scikit-learn/issues/22530
-          - os: windows-2019
-            archs: 'AMD64,x86'
-
-          # Linux 64 bit manylinux2014
-          - os: ubuntu-latest
-            archs: 'x86_64,aarch64'
-
-            # MacOS x86_64
-          - os: macos-latest
-            archs: 'x86_64,arm64'
-
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        python-version: [36, 37, 38, 39, 310]
 
     steps:
     - uses: actions/checkout@v3
       with:
         submodules: recursive
 
-    - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/python_cibuildwheel.yml
```diff
@@ -12,173 +12,29 @@ on:
...
-            cmake_compiler: default
-            allow_failure: false
-          - os: macos-latest
-            bitness: 64
-            python: 310
-            platform_id: macosx_arm64
-            cmake_compiler: default
-            allow_failure: false
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        python-version: [36, 37, 38, 39, 310]
```
jmarrec, line 22: I'm still breaking it up by python version within a given OS or it is super slow, cf commit message in a2cd8b4c572bb445e7d65ff6713638193dcc9270 

> About 1h15 on mac, 45 min on windows

jmarrec:
@ibell the good news is that I can successully install the mac arm64 on my M1 mac, whereas when COOLPROP_CMAKE is used, I would get this error

```
In [1]: from CoolProp.CoolProp import get_global_param_string
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-26173437d578> in <module>
----> 1 from CoolProp.CoolProp import get_global_param_string

~/Virtualenvs/py39/lib/python3.9/site-packages/CoolProp/__init__.py in <module>
     14     quit()
     15
---> 16 from .CoolProp import AbstractState
     17 from . import CoolProp
     18 from . import HumidAirProp

ImportError: dlopen(/Users/julien/Virtualenvs/py39/lib/python3.9/site-packages/CoolProp/CoolProp.cpython-39-darwin.so, 0x0002): symbol not found in flat namespace '__Z5PropsPKccdcdS0_'
```

jmarrec:
I put the arm64 minimal fix into a PR at https://github.com/CoolProp/CoolProp/pull/2123 should we decide we do not want this one

ibell:
So merge this one, correct?

ibell:
I'm used to 1 hour cibuildwheels on other projects, so 16 minutes not too bad

PR #2132, opened 2022-05-07
Bump IF97 to included reverse T(P,H) patch [skip ci]

henningjp:
### Description of the Change

Bump IF97 submodule to v2.1.3 incorporate latest patch.

### Benefits

Fixes error in reverse lookup, T(P,H), in a narrow pressure band,  Region check issues was causing wrong sub-region equation to be used.  Error carried through to the IF97 Backend.  This update corrects issue reported in CoolProp/IF97#35 .

### Possible Drawbacks

None.

### Verification Process

Regression testing against IAPWS-IF97 verification values performed in stand-alone IF97 wrapper (see CoolProp/IF97#36)

### Applicable Issues

Reported in CoolProp/IF97#35.


ibell:
Ok, but watch out for the units in the comments, all the ``MPa*`` should be ``MPa``

henningjp:
@ibell - It was intentional. See comment note just above the constants definitions

```cpp
    // * IAPWS constants use units of MPa and kJ and are entered as such, but converted as needed to SI
    //         depending on the definition of IAPWS_UNITS below.  Main program can define IAPWS_UNITS
    //         to leave all input/output values in IAPWS units, which is handy for results verification
    //         against values printed in the IAPWS documents.  CoolProp will never use this definition.
    //         Converted constants below are commented with an *.
```

😃

ibell:
Gotcha, I should RTFM

PR #2134, opened 2022-05-09
Add fluid_param_string and get_JSONstring to cubic backend

friederikeboehm:
### Description of the Change

fluid_param_string and get_JSONstring were added to the cubic backend. Implementation is comparable to HelmholtzEOSMixturebackend See PR #1371.

Minor changes to CubicBackend.cpp were made:

- If phase is imposed, it is returned instead of always returning gas phase for mixtures
- TODO to return the correct phase is not properly fixed though

### Benefits

Enables the user to get the JSON for a cubic fluid during runtime

### Possible Drawbacks

None

### Verification Process

Functions were tested using the Mathematica wrapper, see PR #2085

### Applicable Issues

based on PR #2133

PR #2133, opened 2022-05-09
New functions for CoolPropLib

friederikeboehm:
Props1SImulti new function in CoolProp and Props1SImulti and PropsSImulti to CoolPropLib

### Description of the Change

Function Props1SImulti was added to CoolProp. It is the Props1SI function but with multiple outputs possible, comparable to PropsSImulti.
Props1SImulti and PropsSImulti were added to CoolPropLib

### Benefits

Enables the user to acces multiple non-thermodynamic-state-dependend fluid properties at once.

### Possible Drawbacks

None

### Verification Process

Functions were tested using the Mathematica wrapper, see PR #2085

PR #2085, opened 2022-01-31
AbstractState Mathematica Wrapper

friederikeboehm:
### Description of the Change

[ *We must be able to understand the design of your change from this description. If we can't get a good idea of what the code will be doing from the description here, the pull request may be closed at the maintainers' discretion. Keep in mind that the maintainer reviewing this PR may not be familiar with or have worked with the wrapper code being implemented, so please walk us through the concepts.* ]

### Benefits

Adds a lot of functionality to the Mathematica Wrapper. Especially the use of the interface layer to the low-level interface was not implemented in Mathematica beforehand.

### Possible Drawbacks

None

### Verification Process

I implemented a mathematica package to use the Library Functions (CoolProp.wl). This also adds error recognition and error messages to the Mathematica Wrapper. This was used in the example.nb Notebook. There I used examples from the CoolProp Homepage ([Access from High-Level Interface](http://www.coolprop.org/coolprop/LowLevelAPI.html#access-from-high-level-interface) and [Cubic Equations of State - Mixtures](http://www.coolprop.org/coolprop/Cubics.html#mixtures)) and received the same results.

grafik
grafik
grafik

### Applicable Issues

Changes in PR #2084 are used in this one
#2065

PR #2135, opened 2022-05-09
AbstractState functions for CoolPropLib

friederikeboehm:
### Description of the Change

AbstractState_get_mole_fractions_satState returns mole fractions for saturated "liquid" or "gas" phases
AbstractState_keyed_output_satState returns keyed output for saturated "liquid" or "gas" phases
AbstractState_backend_name returns the backend name for an AbstractState
add_fluids_as_JSON can be called from the High-Level.

Changes to AbstractState_get_phase_envelope_data

### Benefits

Adds functionality to the High-Level-Interface.

AbstractState_get_phase_envelope_data returns the actual length of the array and throws an error should there not be enough memory allocated.

### Possible Drawbacks

Changes to AbstractState_get_phase_envelope_data might imply refactoring of wrappers (could only find it in Julia wrapper)

### Verification Process

Functions were tested using the Mathematica wrapper, see PR #2085

### Applicable Issues

Based on PR #2133 and #2134

### DIFF
--- a/src/CoolPropLib.cpp
+++ b/src/CoolPropLib.cpp
@@ -568,6 +568,40 @@ EXPORT_CODE void CONVENTION AbstractState_get_mole_fractions(const long handle, 
         HandleException(errcode, message_buffer, buffer_length);
     }
 }
+EXPORT_CODE void CONVENTION AbstractState_get_mole_fractions_satState(const long handle, const char* saturated_state, double* fractions,
+                                                                      const long maxN, long* N, long* errcode, char* message_buffer,
+                                                                      const long buffer_length) {
+    *errcode = 0;
+
+    try {
+        shared_ptr<CoolProp::AbstractState>& AS = handle_manager.get(handle);
+        std::vector<double> _fractions;
+        double quality = AS->Q();
+        if (0 <= quality && quality <= 1) {
+            if (strcmp("liquid", saturated_state) == 0) {
+                _fractions = AS->mole_fractions_liquid();
+            } else if (strcmp("gas", saturated_state) == 0) {
+                _fractions = AS->mole_fractions_vapor();
+            } else {
+                throw CoolProp::ValueError(
+                  format("Bad info string [%s] to saturated state mole fractions, options are \"liquid\" and \"gas\"", saturated_state));
+            }
+        } else {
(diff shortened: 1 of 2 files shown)


ibell:
Can you please rebase or merge so that we don't have all the PropsSImulti stuff in this PR?

ibell:
Can you please rebase or merge so that we don't have all the PropsSImulti stuff in this PR?

ibell:
I prefer no breaking changes to the C interface exposed in the DLL, so please make a new function rather than modifying the existing one.

friederikeboehm:
> I prefer no breaking changes to the C interface exposed in the DLL, so please make a new function rather than modifying the existing one.

That makes sense. I've implemented an overloaded function (as it basically is the same function, just more input and output variables). If this is not wanted and I should reimplement it as a function with a different name, I'll do that of course.

ibell:
Thanks for the updates.  Please address my comments and then we can merge

friederikeboehm:
I can't see any comments, what do you mean?

ibell:
See above in this thread, or the Files Changed tab

friederikeboehm:
That's were I looked and I can't see any comments
There are no conversations going on according to GitHub:
grafik

ibell:
Sorry my bad, they were pending

src/CoolPropLib.cpp
```diff
@@ -568,6 +568,41 @@ EXPORT_CODE void CONVENTION AbstractState_get_mole_fractions(const long handle,
...
+                                                                      const long maxN, long* N, long* errcode, char* message_buffer,
+                                                                      const long buffer_length) {
+    *errcode = 0;
+
+    try {
+        shared_ptr<CoolProp::AbstractState>& AS = handle_manager.get(handle);
+        std::vector<double> _fractions;
+        double quality = AS->Q();
+        if (0 <= quality && quality <= 1) {
+            if (strcmp("liquid", saturated_state) == 0) {
```
ibell, line 581: It is bad practice to use ``strcmp`` in C++, please change to normal comparisons (e.g. ``saturated_state == "liquid" `` throughout)

ibell: I hadn't realized you were doing the string comparison with the C char buffer, anyhow better with the ``std::string``

PR #2152, opened 2022-08-11
PhaseEnvelopeRoutines::is_inside above Pcr

keirstitt84:
Finds nearest point on envolope even if we are above the critical pressure. I've done this the lazy way of looping the entire envolope and for each point finding the closest point, however it would be more efficient to e.g. start at the critical pressure and go left or right.

Also I'm noting that the code for below critical pressure but gas or liquid uses the envelope point for the pressure and calls this the closest. I'm not sure if this is entirely correct - many phase envelopes are highly asymmetrical and the nearest point might not share the same pressure. My understanding is that the isotherms contour around the phase envelope so I’m assuming my approach works – but I’m probably making a horrible mistake here so a bit of feedback from someone with a deeper understanding of the physics would be helpful.
Also I’m doing this to fix the handling of supercritical mixtures – I.e. blends of room temperature gases raised to high pressure such as in a cylinder. And I’m not sure if this routine is called from anywhere using it differently. But because FlashRoutines::PT_flash_mixtures calls this and then with previous code deduces that the supercritical fluid is a liquid, it handles it as such. I'm not sure if PT_flash_mixtures should in fact have a different handling for superitical states. At present it can't handle anything except solutions right of the phase envolope.

### Requirements

### Description of the Change

PhaseEnvelopeRoutines::is_inside now finds closest point on envolope when pressure is above critical pressure - previously it abandoned and merely returned false - which is the primary object of the function but the caller typically also depends upon closest_state being updated.

### Benefits

Supercritical mixtures above critical pressures are now calculated - although there may be errors in the results and maybe cautions need to be given. But previous functionality was to throw an error without any description at all.

### Possible Drawbacks

My search routine is the lazyest going and clearly needs to be changed for something more efficient. This inevitibly adds to computation time.

### Verification Process

I've ran a python routine to plot the isotherms. These outputs appear to curve as you'd expect.

In terms of regressions I don't think this routine is used elsewhere and making changes in an area where things appear to be largely not fully functional anyway. Mixtures are described as experimental only by various contributors so my concern about regressions is limited. The main return value of the this function will remain the same.

### Applicable Issues

I didn't open an issue on this - I was hunting for a similar issue but it didn't seem anyone had put it in appropriately similar terms. But I got a fix together myself within a couple of hours so I just ran with it.

### DIFF
--- a/src/Backends/Helmholtz/PhaseEnvelopeRoutines.cpp
+++ b/src/Backends/Helmholtz/PhaseEnvelopeRoutines.cpp
@@ -696,6 +696,42 @@ bool PhaseEnvelopeRoutines::is_inside(const PhaseEnvelopeData& env, parameters i
         return false;
     }
     if (iInput1 == iP && 0 < env.ipsat_max && env.ipsat_max < env.p.size() && value1 > env.p[env.ipsat_max]) {
+        
+        //written by Keir Stitt 10/8/22
+        //finds nearest point on envolope even if we are above the critical pressure.
+        //I've done this the lazy way of looping the entire envolope and for each point finding the closest point,
+        //however it would be more efficient to e.g. start at the critical pressure and go left or right.
+        //
+        //Also I'm noting that the code for below critical pressure but gas or liquid uses the envolope point 
+        //for the pressure and calls this the closest. I'm not sure if this is entirely correct - many phase envolopes
+        //are highly asymetrical and the nearest point might not share the same pressure. My understanding is that the 
+        //isotherms contour around the phase envelope so I’m assuming my approach works – but I’m probably making a 
+        //horrible mistake here so a bit of feedback from someone with a deeper understanding of the physics would 
+        //be helpful.
+        
+        double maxdistance = std::numeric_limits<double>::infinity();
+        double distance;
+        std::size_t iclosest;
+
+        for (std::size_t i = 0; i < env.p.size() - 1; ++i) {
+            distance = sqrt(pow(value1 - env.p[i], 2) + pow(value2 - env.T[i], 2));
+            if (distance < maxdistance) {
+                maxdistance = distance;
+                iclosest = i;
+            }
+        }
+
+        if (iclosest >= 0 && iclosest < env.p.size()) {
+            closest_state.T = env.T[iclosest];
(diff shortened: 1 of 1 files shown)


keirstitt84:
Just on testing - So I used the below code to produce the below Ph diagram for an N2, O2, Ar blend [air]. I also put on the phase envolope. I don't think it looks too bad considering it gives us a region of operation we didn't previously have.


import CoolProp as CP
import math
import matplotlib.pyplot as plt

backend = "HEOS"
mixture = "Nitrogen&Oxygen&Argon"
ratios = [.78084, .20946, .00934]

ASmix = CP.AbstractState(backend, mixture)
ASmix.set_mole_fractions(ratios)
ASmix.build_phase_envelope("dummy")
phasedata = ASmix.get_phase_envelope_data()

pressurestart = math.pow(10, 3)
pressureend = math.pow(10, 8)

presssuresteps = (math.log10(pressureend) - math.log10(pressurestart)) * 10



temperature = 273.15 + 15

for temperature in range(10, 500, 10):

    print("plotting for t = ", temperature, "K")

    pressurevalues = range(int(math.log10(pressurestart)*presssuresteps), int(math.log10(pressureend)*presssuresteps))

    h = []
    p = []
    t = []

    for a in pressurevalues:
        pressure = math.pow(10, a / presssuresteps)
        try:
            ASmix.update(CP.PT_INPUTS, pressure, temperature)
            h.append(ASmix.hmolar())
            p.append(pressure)
            t.append(temperature)
        except ValueError as e:
            print ("Value Error : ", e.args[0] )
            break


    plt.plot(h, p, t)

plt.plot(phasedata.hmolar_vap, phasedata.p)

plt.xlabel('Molar Enthalpy [J/mol]')
plt.ylabel('Pressure [Pa]')
plt.yscale('log')

plt.savefig('Ph Diagram.png')

print ("end")


Ph Diagram_N2_O2_Ar

keirstitt84:
I'm going to close this draft pull request because in its current form needs a bit more work.

I've had some great ideas now to get this working which I hope to post in a more full feature request.

If anyone else wishes to contribute they can pick up on the phase envelope for >2 component mixtures.

PR #2159, opened 2022-08-25
Improve pcsaft flash

zmeri:
### Description of the Change

The flash algorithm for the PC-SAFT backend was updated to improve its performance. Additionally, the association term for the PC-SAFT EOS was expanded to also support other association schemes, in addition to the default 2B scheme used in the earlier code.

### Benefits

The flash algorithm is, generally, faster and more reliable, although it could still be improved. Also, a wider range of association schemes can be used.

### Possible Drawbacks

New bugs might have been introduced.

### Verification Process

I ran the Catch tests in CoolProp-Tests.cpp. All the PC-SAFT tests passed. Of all the tests only 285/51440 assertions failed, and it appeared that many of these were because I didn't have Refprop installed. For comparison, when I run the tests on the current master branch 306/51436 fail, and some of those that fail are for PC-SAFT tests. So, this update should help correct some of the errors in the current master branch.


ibell:
Can you please rebase so that we get only your changes? 50 files are changed in this PR

PR #2176, opened 2022-10-19
add PCSAFT page in docs

zmeri:
### Description of the Change

I added a page in the docs for the PC-SAFT equation of state. I also added the references for the PC-SAFT fluid parameters to the BibTeX file.


ibell:
Nice, thanks for adding this.  Merging.

PR #2202, opened 2022-12-28
Robust hsu D flash method

### LINKED ISSUE #1965: Low-level Interface error when using DmassUmass_INPUTS
Please excuse my first GitHub Issue report...

### Description

Using CoolProp to provide fluid thermodynamic property data to simple Nitrogen tank blow-down model.  Model equations of state track Conservations of Mass and Energy (Enthalpy -> Internal Energy form).  Differential equation solver integrates tank fluid total internal energy, while the CoolProp low-level interface is requested to update based on mass density and specific internal energy.  During the course of a simulation, the following traceback error occurs:

  File "\Documents\fluidprops\FluidProps.py", line 152, in update
    AS.update(CoolProp.DmassUmass_INPUTS, rho, u)

  File "CoolProp\AbstractState.pyx", line 102, in CoolProp.CoolProp.AbstractState.update

  File "CoolProp\AbstractState.pyx", line 104, in CoolProp.CoolProp.AbstractState.update

ValueError: Zero occurred in row 1, the matrix is singular. 

The last echoed tank condition prior to this error being thrown is:
Mass Density: 313.305125 kg/m3
Specific Internal Energy: 154834.285193 J/kg

### Steps to Reproduce

within Spyder IDE (4.1.3):

1. import CoolProp
2. AS = CoolProp.AbstractState("HEOS","nitrogen")
3. rho = 313.305125
4. u = 154834.285193
5. AS.update(CoolProp.DmassUmass_INPUTS, rho, u)

**Expected behavior:**
No error. The ability to call "AS.T()" and get Nitrogen temperature, in Kelvin

**Actual behavior:** [What actually happens]

Spyder consolde returns:

  File "<ipython-input-23-2c3f11f50064>", line 1, in <module>
    AS.update(CoolProp.DmassUmass_INPUTS, rho, u)

  File "CoolProp\AbstractState.pyx", line 102, in CoolProp.CoolProp.AbstractState.update

  File "CoolProp\AbstractState.pyx", line 104, in CoolProp.CoolProp.AbstractState.update

ValueError: Zero occurred in row 1, the matrix is singular. 

### Versions

**CoolProp Version:** 6.4.1  
**Operating System and Version:** Windows 10 Enterprise, 10.0.18362
**Access Method:** Python 3.7.6

### Additional Information

If possible, please post examples and/or screenshots of the issue.
image

### LINKED ISSUE #2157: Calculating pressure from density and internal energy within two-phase region fails within certain ranges of vapour quality. Range increases for higher pressures to.
I have had some issues with determining the pressure in the two-phase region based on the density and internal energy. Looking into previously asked questions I noticed this sometimes happens near the phase boundary regions, which indeed occurs irregularly at those points.
However, creating arrays of density and internal energy over the pressure range of [0.1*p_crit;0.8*p_crit] and with Q between [0;1] in coolprop and then doing the inverse (ie getting the pressure out of each point of density and internal energy) a pattern emerges where either the secant residual function returns an invalid number or the brent method reaches its maximum number of steps of 20. 

I tried both R1233ZDE and R32 both with 1000 subdivisions in both pressure and quality range.
For R1233ZDE the PropsSI command failed for the D,U combo from Q=0.024 for the lowest pressures to Q=0.22 to Q=0.24 for highest pressures. This trend is spread over the entire pressure range and moves slowly to higher Q's and a wider Q range for increasing pressure. (See small part of the resulting pressure array where -1 is used when the PropsSI function failed the calculation.)
The results for R32 are similar.

(in the .py file ill add I made 100 subdivisions each to shorten the calculation time, but the trend can still be seen. Mostly at the higher pressures where the problematic range is bigger.)

Has anyone stumbled across this issue before or know why this happens for a certain quality range, and if so, how did you fix it?
For now, I am circumventing it via interpolation as proposed by jowr in issue #1054

Thanks in advance for any help.
Xander

CoolProp Version: 6.3.0
Access Method: Python
[CoolpropPfromDU.zip](https://github.com/CoolProp/CoolProp/files/9373763/CoolpropPfromDU.zip)
LowPressures

HighPressures

msaitta-mpr:
### Description of the Change

Providing inputs to CoolProp of D and U (or any of HSU) is usually stable, but sometimes can cause annoying issues. This is noted in issues #1965 and #2157 (as well as some of my personal projects). This PR attempts to solve these issues once and for all as an extension to the work done in PR #2173. This is done in 3 main changes:

* Add additional checks and attempts at solving the ancillary equations. Adds more damping attempts and a way to force a guess.
* Add limits to the extrapolated ancillary guesses. If the ancillary becomes very flat near the edges (e.g., near the triple point or critical point), it might not have a solution, but the extrapolated guess could be very far off. We limit the extrapolated guess so that the Akshasa solver can at least attempt something. 
* If all else fails, we give up on finding the saturation curve, and apply a Brent method directly to find T by iterating H/S/U. We do not know the phase, but at least we have everything else. 

### Benefits

We should now always get a solution for D_HSU inputs. 

### Possible Drawbacks

It is possible that one of our fallbacks does not provide the user with a phase, instead setting it to iphase_unknown. 

The fallback solver can be slow because of needing a ton of Brent iterations, after already doing thousands of other iterations trying to find the saturation curve (which is now allowed to go to many of thousands of evaluations). However, a slow solution is almost always better than no solution. 

### Verification Process

I ran large sweeps of inputs to ensure that no errors were observed. I also integrated the results with another code and it seems to behave acceptably. I made custom test codes for issues #1965 and #2157. See the attached C++ code. 

[coolPropTester.zip](https://github.com/CoolProp/CoolProp/files/10314853/coolPropTester.zip)

### Applicable Issues

Closes #1965
Closes #2157

### DIFF
--- a/include/Ancillaries.h
+++ b/include/Ancillaries.h
@@ -9,6 +9,15 @@
 
 namespace CoolProp {
 
+struct SaturationAncillaryFunction_invert_options
+{
+    double min_bound = -1;
+    double max_bound = -1;
+    int max_iter = 200;
+    double omega = 1.;
+    bool best_guess = false;
+};
+
 /**
 The surface tension correlation class uses correlations for the surface tension that are all
 of the form
@@ -137,7 +146,7 @@ class SaturationAncillaryFunction
     /// @param min_bound (optional) The minimum value for T; ignored if < 0
     /// @param max_bound (optional) The maximum value for T; ignored if < 0
     /// @returns T The temperature in K
-    double invert(double value, double min_bound = -1, double max_bound = -1);
+    double invert(double value, SaturationAncillaryFunction_invert_options options = SaturationAncillaryFunction_invert_options()) ;
 
     /// Get the minimum temperature in K
     double get_Tmin(void) {
--- a/include/Solvers.h
+++ b/include/Solvers.h
@@ -73,7 +73,7 @@ class FuncWrapperND
 double Brent(FuncWrapper1D* f, double a, double b, double macheps, double t, int maxiter);
 double Secant(FuncWrapper1D* f, double x0, double dx, double ftol, int maxiter);
 double BoundedSecant(FuncWrapper1D* f, double x0, double xmin, double xmax, double dx, double ftol, int maxiter);
-double ExtrapolatingSecant(FuncWrapper1D* f, double x0, double dx, double ftol, int maxiter);
+double ExtrapolatingSecant(FuncWrapper1D* f, double x0, double dx, double ftol, int maxiter, bool best_guess);
 double Newton(FuncWrapper1DWithDeriv* f, double x0, double ftol, int maxiter);
 double Halley(FuncWrapper1DWithTwoDerivs* f, double x0, double ftol, int maxiter, double xtol_rel = 1e-12);
 double Householder4(FuncWrapper1DWithThreeDerivs* f, double x0, double ftol, int maxiter, double xtol_rel = 1e-12);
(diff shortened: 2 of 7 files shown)


ibell:
This is an impressive PR, addressing a particular point that has caused me a lot of pain.  Thank you very much.  Are there any points to discuss before merging?

@jowr Do you want to take a look?

jowr:
Hi - I just surfaced again after a very busy period. I do not see any problems, but I haven't checked in detail. @msaitta-mpr  - I guess this PR is still relevant?

msaitta-mpr:
Both of the referenced issues are still open, so I assume that it is still a needed change. However, I have not attempted to rebase in the last 2 years, so I am not sure if there will be any unforseen merge issues.

ibell:
I have added superancillary functions and was getting ready to address this issue for pure fluids. @msaitta-mpr do you want to team up to do that (merged with this PR)?

msaitta-mpr:
OK. I will take a look and see what I can do.

ibell:
I think this PR has been superceded by #2995. Only helps when superancillaries are available, but that is basically all pure fluids now. The speedup is meaningful, and the reliability is very good

PR #2203, opened 2022-12-28
Provide better feedback for bad DQ inputs

msaitta-mpr:
### Description of the Change

DQ is poorly defined (not unique) for densities above critical density. Provide the user better feedback when they try to do this, but do not give them a non-unique answer that might give them trouble later. This PR provides the user more direct feedback warning them about this, but does not try to guess at one of the non unique solutions.

#### A bit more about DQ
The HEOS defines properties in terms of density (D) and temperature (T). If we provide a DQ input pair, the first thing that the solver attempts to do is to find the associated temperature. However, above the critical density, temperature is not a function of density and therefore no unique state exists. 

For all (most?) fluids, DQ inputs are not unique for densities above the critical density. Take for example the following plots of R134a, water, hydrogen, and propane.
r134
water
hydrogen
propane

Things become a bit clearer if we plot it on a linear scale (water for remaining plots). Note that the curves are curves of constant quality. The orange curve is 5% quality and it is clear that it does not pass the vertical line test. Therefore, we can see that temperature is not a function of density for a given quality. 
linear_quality

To put it another way, the solver can calculate quality for a given DT pair. We know D, so we iterate on T until we find the quality that we want. Plotted below is Q for D=450 and a range of temperatures. If we want to find the temperature for which the quality is 0.05, there is no unique solution. This lack of a solution is why the current version of the code typically crashes due to Brent failing to bracket. 
solver_issue

### Benefits

Provide the user more robust feedback so that they can make a decision about how to handle their thermodynamic state. E.g., they may choose to build their own solver, ignore the error, or recognize that they need a new method. 

### Possible Drawbacks

This does not actual prevent an exception from being thrown, but simply adds additional text helping explain to the user what went wrong. 

### Verification Process
I wrote a quick code the proves out of range inputs and gets the new error text as expected. Because we are not changing much, more robust regression is not needed. 

```
#include "CoolProp.h"
#include "AbstractState.h"
#include <iostream>
#include <memory>

int main(){
    std::shared_ptr<CoolProp::AbstractState> AS = std::shared_ptr<CoolProp::AbstractState>(CoolProp::AbstractState::factory("HEOS", "water"));
    double rho = 450;
    double q = 0.05;
    try{
        AS->update(CoolProp::DmassQ_INPUTS, rho, q);
        std::cout << AS->T() <<std::endl;
    } catch (CoolProp::CoolPropBaseError e){
        std::cout << e.what() << std::endl;
    }
    return 0;
}
```

### Applicable Issues

No issue, but resolves the discussion from PR #2173 .

### DIFF
--- a/src/Backends/Helmholtz/FlashRoutines.cpp
+++ b/src/Backends/Helmholtz/FlashRoutines.cpp
@@ -260,22 +260,17 @@ void FlashRoutines::DQ_flash(HelmholtzEOSMixtureBackend& HEOS) {
         double Tmin = HEOS.Tmin() + 0.1;
         double rhomolar = HEOS._rhomolar;
         double Q = HEOS._Q;
-        DQ_flash_residual resid(HEOS, rhomolar, Q);
-        try {
-            Brent(resid, Tmin, Tmax, DBL_EPSILON, 1e-10, 100);
-            HEOS._p = HEOS.SatV->p();
-            HEOS._T = HEOS.SatV->T();
-            HEOS._rhomolar = rhomolar;
-            HEOS._Q = Q;
-            HEOS._phase = iphase_twophase;
-        } catch (ValueError e) {
-            const double eps = 1e-12; // small tolerance to allow for slop in iterative calculations
-            if (rhomolar >= HEOS.rhomolar_critical() && Q > (0 + eps) && Q < (1 - eps)){
-                throw CoolProp::OutOfRangeError(format("DQ inputs are not defined for density above critical density (%g)", HEOS.rhomolar_critical()).c_str());
-            } else {
-                throw;
-            }
+        const double eps = 1e-12; // small tolerance to allow for slop in iterative calculations
+        if (rhomolar >= (HEOS.rhomolar_critical() + eps) && Q > (0 + eps)){
+            throw CoolProp::OutOfRangeError(format("DQ inputs are not defined for density above critical density (%g)", HEOS.rhomolar_critical()).c_str());
         }
+        DQ_flash_residual resid(HEOS, rhomolar, Q);
+        Brent(resid, Tmin, Tmax, DBL_EPSILON, 1e-10, 100);
+        HEOS._p = HEOS.SatV->p();
+        HEOS._T = HEOS.SatV->T();
+        HEOS._rhomolar = rhomolar;
+        HEOS._Q = Q;
+        HEOS._phase = iphase_twophase;
     } else {
         throw NotImplementedError("DQ_flash not ready for mixtures");
     }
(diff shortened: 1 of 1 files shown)


ibell:
Nice idea, but should only be applied if density is above the critical
density and the quality is above zero. The saturated liquid curve is
monotonic and invertible, at least in the ideal case.

msaitta-mpr:
That is a good point. I added a commit ([689e8ea](https://github.com/CoolProp/CoolProp/pull/2203/commits/689e8eaac69b8afb2e5adb08f2f0b88736da8849)) which should prevent it from ever throwing this new error on the saturation boundaries. Note that it will still always try to find a solution, but will only throw this new error when it fails.

ibell:
You might want a bit of slop in your check of the domain (0,1). Usually I do (-epsilon, 1-epsilon), with epsilon something like 1e-12.  This is because iterative calculations will often yield not precisely Q of 1 or 0.

jowr:
Thank you for contributing! Would it make sense to skip the calculation when we have Q>0 and rho >= rho_crit? Is there any benefit in letting the solver try to find a solution?

msaitta-mpr:
@ibell I pushed a commit (f964db3) which adds 1e-12 slop per your suggestion. 

@jowr I considered this, but was unable to prove that a solution does not exist if and only if rho>rho_crit and Q>0. My fear was that by changing the actual value that is returned, I would miss some small, valid region and break someone's work. My assumption is that in most cases, we do not care about a few extra microseconds of calculation when we are going to have to throw an exception. If someone is very worried about speed, they can perform the check themselves before calling the CoolProp flash routine. Of course, if you are confident that the density always makes the solution invalid, then we can swap to an early return easily enough.

ibell:
I think I have settled on agreeing with @jowr. I think we should never allow the calculation if rho > rhoc and Q > 0.

msaitta-mpr:
@ibell I trust you to know the right thing to do and have added a commit (818f454) to implement this requirement.

src/Backends/Helmholtz/FlashRoutines.cpp
```diff
@@ -260,6 +260,10 @@ void FlashRoutines::DQ_flash(HelmholtzEOSMixtureBackend& HEOS) {
         double Tmin = HEOS.Tmin() + 0.1;
         double rhomolar = HEOS._rhomolar;
         double Q = HEOS._Q;
+        const double eps = 1e-12; // small tolerance to allow for slop in iterative calculations
+        if (rhomolar >= (HEOS.rhomolar_critical() + eps) && Q > (0 + eps)){
+            throw CoolProp::OutOfRangeError(format("DQ inputs are not defined for density above critical density (%g)", HEOS.rhomolar_critical()).c_str());
```
ibell, line 265: The message should also indicate the Q part of the check (that Q >= 0)

ibell:
Otherwise, I think this is good to go.  

@msaitta-mpr Thank your for your very high-quality contributions. Do you want to introduce yourself at https://github.com/CoolProp/CoolProp/discussions/2206 ?

ibell:
Thanks!

PR #2207, opened 2023-01-22
Verify that mole fractions are set before using them

### LINKED ISSUE #2205: Python silently crashes when calling trivial_keyed_output on binary mixtures
### Description

I tried to fiddle with binary mixtures to potentially include them in TESPy (https://github.com/oemof/tespy/discussions/375) and found, if any `trivial_keyed_output` is called on an instance of `AbstractState` with a binary mixture, python just crashes silently for me.

### Steps to Reproduce

For example (I tested it with multiple pairs):

```python

AS = CP.AbstractState("HEOS", "R134a&R1234YF")

AS.trivial_keyed_output(CP.iT_min)

print("Silent crash")
```

**Expected behavior:** Script crashes with error

**Actual behavior:** No error is raised, script execution stops at `AS.trivial_keyed_output(CP.iT_min)`.

### Versions

**CoolProp Version:** 6.4.3.post1
**Operating System and Version:** Windows 10 (Cannot test on different OS at the moment)
**Access Method:** Python

### Additional Information

Terminal output of the above code:

grafik

msaitta-mpr:
### Description of the Change

Add a check for mole fractions being set before attempting to use them.

### Benefits

Prevents hard crash in Python module. 

Currently, if mole fractions are not set, then there is an exception thrown in the vector's out of bounds check (`"vector subscript out of range"`). In the created Python module, there is no handling for this, so the code just crashes. We now check for the mole fraction size before using and throw a CoolProp::ValueError. This is recognized by the Python module and allows for Python side exception handling. 

### Possible Drawbacks

* I may have missed a usage of the mole fractions and this issue may reappear somewhere in the future. 
* It is awkward to have the check called at all places where mole fractions are used. It would be more elegant to place the check at a higher level so that future modifications do no need to manually include this check. However, I am not sure if there are alternative code paths that lead to the mole fractions, so I put the check close to the actual usage. 

### Verification Process
The test from #2205 was executed:
``` python
AS = CP.AbstractState("HEOS", "R134a&R1234YF")
AS.trivial_keyed_output(CP.iT_min)
```

Now, an exception is thrown.
```
>>> import CoolProp.CoolProp as CP
>>> AS = CP.AbstractState("HEOS", "R134a&R1234YF")
>>> AS.trivial_keyed_output(CP.iT_min)
  File "<stdin>", line 1, in <module>
  File "CoolProp\AbstractState.pyx", line 227, in CoolProp.CoolProp.AbstractState.trivial_keyed_output
  File "CoolProp\AbstractState.pyx", line 229, in CoolProp.CoolProp.AbstractState.trivial_keyed_output
ValueError: mole fractions are not set for all components
```

This can be handled if desired:
``` python
try:
    AS.trivial_keyed_output(CP.iT_min)
except Exception as e:
    print("Handle this:", e)
```

### Applicable Issues

Closes #2205

### DIFF
--- a/src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
+++ b/src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
@@ -483,6 +483,7 @@ CoolPropDbl HelmholtzEOSMixtureBackend::calc_gas_constant(void) {
         } else {
             // mass fraction weighted average of the components
             double summer = 0;
+            verify_mole_fractions_set();
             for (unsigned int i = 0; i < components.size(); ++i) {
                 summer += mole_fractions[i] * components[i].gas_constant();
             }
@@ -492,6 +493,7 @@ CoolPropDbl HelmholtzEOSMixtureBackend::calc_gas_constant(void) {
 }
 CoolPropDbl HelmholtzEOSMixtureBackend::calc_molar_mass(void) {
     double summer = 0;
+    verify_mole_fractions_set();
     for (unsigned int i = 0; i < components.size(); ++i) {
         summer += mole_fractions[i] * components[i].molar_mass();
     }
@@ -660,6 +662,7 @@ CoolPropDbl HelmholtzEOSMixtureBackend::calc_viscosity(void) {
     } else {
         set_warning_string("Mixture model for viscosity is highly approximate");
         CoolPropDbl summer = 0;
+        verify_mole_fractions_set();
         for (std::size_t i = 0; i < mole_fractions.size(); ++i) {
             shared_ptr<HelmholtzEOSBackend> HEOS(new HelmholtzEOSBackend(components[i]));
             HEOS->update(DmolarT_INPUTS, _rhomolar, _T);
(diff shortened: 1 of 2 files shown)


ibell:
I think your solution is a good one for now. I think it would be better, in general, to make the mole fractions object totally private, with a getter function that you call to get a reference to it. In that way, the check could happen in the getter. Not sure if you want to take that on, or go with your patch.

msaitta-mpr:
I think that your idea is a good one. I have tried to implement it (450624c), but admit that I am a bit new to C++ so may have made an error somewhere.

ibell:
This is a very nice solution, even better than the idea I had. Does it pass the tests?

Your coding skills are impressive, especially if C++ is not your native working environment.

@jowr can you review this PR?

msaitta-mpr:
I thought that this is exactly what you were suggesting, as I would not have come up with this on my own. 

It does not pass all of the automated test cases. cppcheck and Catch2 fail, but these tests also fail (in the same way) for the master branch. If there are other regression tests, I am not sure where they are.

ibell:
@jowr ping for review. It looks ok to me, wanted to see if you had any thoughts.

jowr:
I also think it looks good. Is there any reason why the new fraction object is only used in the HEOS backend and not globally?

PR #2213, opened 2023-02-07
Use lazy initialization and avoid static objects

jowr:
Use lazy initialization and getter functions to avoid problems with the order of initialization when linking statically.

I have had some trouble with the functions that use global maps (get_param_information and friends) since they are not guaranteed to be initialised at runtime. In my case, some constants should be initialised early using these functions and I had mysterious crashes caused by empty maps. 

This PR fixes that problem with almost no overhead.


src/DataStructures.cpp
```diff
@@ -153,9 +154,16 @@ class ParameterInformation
     }
 };
 
-static ParameterInformation parameter_information;
+std::unique_ptr<ParameterInformation> parameter_information_p;
+ParameterInformation& get_parameter_information() {
```
ibell, line 158: Can this be a const ref here?

src/DataStructures.cpp
```diff
@@ -379,12 +389,21 @@ class PhaseInformation
         }
     }
 };
-static PhaseInformation phase_information;
+
+std::unique_ptr<PhaseInformation> phase_information_p;
+PhaseInformation& get_phase_information() {
```
ibell, line 394: same question as above

src/DataStructures.cpp
```diff
@@ -437,13 +456,22 @@ class SchemeInformation {
         }
     }
 };
-static SchemeInformation scheme_information;
+
+std::unique_ptr<SchemeInformation> scheme_information_p;
+SchemeInformation& get_scheme_information() {
```
ibell, line 461: again

src/DataStructures.cpp
```diff
@@ -541,9 +569,16 @@ class InputPairInformation
     }
 };
 
-static InputPairInformation input_pair_information;
+std::unique_ptr<InputPairInformation> input_pair_information_p;
+InputPairInformation& get_input_pair_information() {
```
ibell, line 573: etc.

src/DataStructures.cpp
```diff
@@ -763,10 +800,17 @@ class BackendInformation
     }
 };
 
-static BackendInformation backend_information;
+std::unique_ptr<BackendInformation> backend_information_p;
+BackendInformation& get_backend_information() {
```
ibell, line 804: last one

ibell:
I see the point; a similar approach was used to manage auto-registration of backends.  Once we deal with the question about const ref, I think we can merge.

ibell:
The performance hit should only be a single check of the pointer, right?

jowr:
Thanks for reviewing the changes!

PR #2214, opened 2023-02-17
Change links from Google group to GitHub discussions

raphaeltimbo:
As discussed here:
https://groups.google.com/g/coolprop-users/c/l_aY4vTT58Q

### Description of the Change

Change links in readme and docs to point to github discussions.

### Benefits

I think the github discussions has a better format to discuss programming questions, it is easier to read, and we have options such as choosing the right answer, upvote etc. which are similar to stackoverflow.

Additionally, we would have all content (issues + discussions) in only one place, making it easier to search things.

It would also make it easier to reference things, for example, issues that could be open based on discussions.


ibell:
I'm good with this. Thank you!

PR #2221, opened 2023-03-14
Add a new REFPROP configuration parameter to use mass-based vapour quality

jowr:
### Description of the Change

A new configuration parameter to enforce using mass-based vapour quality calculation in the REFPROP backend.

### Benefits

Calls using mass-based quality can now be calculated inside CoolProp and one no longer needs to perform additional calculations.

### Possible Drawbacks

Checking the configuration variable is expected to have a minimal impact on performance.


ibell:
I have thought about this problem before, and came up with a different solution. I think the more maintainable solution is to add input keys of ``QmassT_INPUTS`` and ``QmolarT_INPUTS``, where ``QT_INPUTS`` is an alias to ``QmolarT_INPUTS``. Same thing for all other input pairs including ``Q``.  I think this approach is more flexible because the conversion function can be implemented in the ``AbstractState`` class and used in all backends.

kaern:
I found this branch just after committing this proposal https://github.com/kaern/CoolProp/tree/mass-quality-methods
The refprop_mass_quality branch appears empty to me. I actually used a flag in the beginning, passed similar ast he Qmass method, but finally decided to go for defining QmassT_inputs and PQmass_inputs, as I believe that is more coolpropic. If you look into my branch, then try to look at test_phase_envelope.py (don't think about the name). let me know if you think of something better. I also fixed surface tension to use densities when possible. With this branch I could finally get exactly the same values as in the refprop gui. I also noticed mole concentrations seem to be cut first component in mixtures, but i did not dare to look into that at this stage.

kaern:
I see the contents of the branch feature/refprop_mass_quality now, but it was never merged into master. I would need this feature in CoolProp (moving away from my own branch), so any changes that it might be merged @jowr branch or mine: https://github.com/kaern/CoolProp/tree/mass-quality-methods .. either way work for me.
image

ibell:
I took a look at the branch and I think it is like 80% of the way there. If you want to make a PR I can review it, or you can give me a couple of days and I can get Claude to start with your code and implement for both HEOS and REFPROP backends.

kaern:
Thanks a lot @ibell - I will make a PR and link to this issue.. 
Thanks for adding this feature, your time and availability.. looking forward to it - a lot  :-)
Thanks again ---- appreciated

ibell:
Up to you whether you want me to help you get the REFPROP part going, or whether you rather I just hand it off to Claude to go crazy and do it CoolProp-wide. If you are interested in further contributions, I would prefer the first path.

kaern:
I saw your comments to the PR and your suggestions sounds good, however, I think I have to let it up to You and Claude. Me and Claude would be much more cracy (it is basically what you already have seen). But as you know the complete structure of CoolProp, I think it is best Claude with your quidance, so everything go smoother.

ibell:
One thing that I only realized when getting into this the weeds of this implementation, is that it is a bad idea to use Qmass in general unless you must. The reason is that while it is easy to calculate Qmass if you know Qmolar and the mole fractions, if you know the bulk phase mole fractions and Qmass, you have to iterate (as part of the flash calculation) to find Qmolar. That comes at a relatively high computational cost I think.

ibell:
First PR version is out, additional review and testing still needed though

ibell:
I think we can close this PR now, given that the comprehensive solution with Qmass is now available. Ended up being a bit more subtle than I had appreciated

PR #2225, opened 2023-03-21
update cyclopentane.json

### LINKED ISSUE #2051: Cyclopentane EOS needs to be updated
In 9fbc32c77cf26775505bf6a3da069494b7e9a7ab the reference for cyclopentane was updated, but the EOS was also changed from the preprint, and so needs to be updated.

longemen3000:
### Requirements

### Description of the Change

Updates the Cyclopentane JSON data to add use the reference equation in http://dx.doi.org/10.1021/je5010164

### Benefits

should solve #2051

### Applicable Issues

closes #2051


ibell:
Thank you! Is the entry in the BibTeX file also updated?

longemen3000:
I don't think so, let me check and update if necessary

longemen3000:
the bibtex entry (`Gedanitz-JCED-2015`) seems correct

PR #2230, opened 2023-03-29
Topic-2200: Correct typo in n-Hexane rhoV auxilliary

### LINKED ISSUE #2200: CoolProp pure Hexane bug
### Description
It seems like a error on my part but for some reason CoolProp seems unable to calculate the vapor temperature of hexane at low pressures.

### Steps to Reproduce
I just want to convert vapor pressures of pure n-hexane into the corresponding temperature.
6065 Pa is the Vapor pressure of Hexane at 0°C.
 Code:
  
CP.PropsSI('T','P',6065,'Q',0,"Hexane")

**Expected behavior:** [What you expect to happen]

I expect it to return 273. 
As this is the vapor temperatur in Kelvin.

**Actual behavior:** [What actually happens]

ValueError: solver_rho_Tp was unable to find a solution for T=    507.82, p=      6065, with guess value     5522.8 with error: The molar density of -9.484198 mol/m3 is below the minimum of 0.000000 mol/m3 : PropsSI("T","P",6065,"Q",0,"Hexane")

### Versions

**CoolProp Version:** 6.4.3.post1
**Operating System and Version:** Windows 10 Enterprise
**Access Method:** Im using python and spyder

### Additional Information

I also tried to calculate the pressure for 273K and input that pressure. 
but that is also not possible:
Pressure = CP.PropsSI('P','T',273,'Q',0,"Hexane")

print(Pressure)

Temp = CP.PropsSI('T','P',Pressure,'Q',0,"Hexane")

print(Temp)

solver_rho_Tp was unable to find a solution for T=    507.82, p=   5992.24, with guess value    5459.22 with error: The molar density of -10.228558 mol/m3 is below the minimum of 0.000000 mol/m3 : PropsSI("T","P",5992.242029,"Q",0,"Hexane")

msaitta-mpr:
The rhoV auxiliary for n-Hexane had accidentally been set to use reduced tau. This commit fixes the typo. This prevents bad guesses for a number of property evaluations which use saturated states. See also https://github.com/usnistgov/teqp/commit/7de5755594f0c3ba201a9ecd368be5dd0543d7a7 .

### Description of the Change

Correct a typo in the rhoV auxiliary for n-Hexane. The value had been accidentally set improperly. This was discovered during investigation of issue #2200.

### Benefits

n-Hexane now gets better guesses when calculating saturated states and has fewer crashes.

### Possible Drawbacks

I have only performed testing for the specific state point mentioned in #2200 (see below). There may be other side effects that have not been tested for explicitly. 

### Verification Process

I ran the following C++ code from #2200. Before this fix, it caused an uncaught ValueError to be raised. It now returns the expected value. 

```C++
// Test issue #2200
#include "CoolProp.h"
#include "AbstractState.h"
#include <iostream>
#include <memory>
int main()
{
    std::shared_ptr<CoolProp::AbstractState> AS = std::shared_ptr<CoolProp::AbstractState>(CoolProp::AbstractState::factory("HEOS", "Hexane"));
    AS->update(CoolProp::PQ_INPUTS, 6065, 0);
    std::cout << AS->T() << std::endl;
    return 0;
}
```

### Applicable Issues

Closes #2200

### DIFF
--- a/dev/fluids/n-Hexane.json
+++ b/dev/fluids/n-Hexane.json
@@ -96,7 +96,7 @@
       "T_r": 507.82,
       "Tmax": 507.82,
       "Tmin": 177.83,
-      "description": "rho'' = rhoc*exp(Tc/T*sum(n_i*theta^t_i))",
+      "description": "rho'' = rhoc*exp(sum(n_i*theta^t_i))",
       "max_abserror_percentage": 1.0104195769602642,
       "n": [
         -3.4056,
@@ -116,7 +116,7 @@
         30.0
       ],
       "type": "rhoV",
-      "using_tau_r": true
+      "using_tau_r": false
     },
     "sL": {
       "A": [


ibell:
Thank you!

PR #2252, opened 2023-05-12
Update fmt submodule to 10.0.0

### LINKED ISSUE #2251: Unable to compile with fmt 10.0.0
Do you want to ask a question? Are you looking for technical support? For any kind of question regarding CoolProp and its usage, you can ask the [CoolProp user group](https://goo.gl/Pa7FBT)

If you have an issue to report, please continue and fill out the applicable sections below.  The details provided will help to resolve this issue as quickly as possible.

### Prerequisites

Before posting an issue, please:  
    * Check the [CoolProp User Group](https://goo.gl/Pa7FBT) for common solutions.  
    * Check that your issue isn't already filed under [CoolProp Issues](https://github.com/CoolProp/CoolProp/issues).  
    * Check the CoolProp [latest release](http://www.coolprop.org) and [development version](http://www.coolprop.org/dev) documentation.  

### Description

Using the latest version of fmt (10.0.0), compilation fails on all platforms due to breaking changes. (See https://github.com/microsoft/vcpkg/pull/31378)

### Steps to Reproduce

1. Install fmt 10.0.0
2. Compile CoolProp 6.4.3

**Expected behavior:** Compilation succeeds

**Actual behavior:** Compilation fails with the following message:

```
D:\installed\x64-windows\include\fmt\core.h(1691): error C2338: static_assert failed: 'Cannot format an argument. To make type T formattable provide a formatter<T> specialization: https://fmt.dev/latest/api.html#udt'
D:\installed\x64-windows\include\fmt\core.h(1711): note: see reference to function template instantiation 'fmt::v10::detail::value<Context> fmt::v10::detail::make_value<Context,const CoolProp::input_pairs&>(T)' being compiled
        with
        [
            Context=context,
            T=const CoolProp::input_pairs &
        ]
```

### Versions

**CoolProp Version:** 6.4.3
**Operating System and Version:** vcpkg CI runners (Windows, macOS, Linux)
**Access Method:** N/A

### Additional Information

The changelog for fmt [10.0.0](https://github.com/fmtlib/fmt/releases/tag/10.0.0) contains:

```
Removed deprecated implicit conversions for enums and conversions to primitive types for compatibility with std::format and to prevent potential ODR violations. Use format_as instead.
```

The compilation error seems related to this change as `CoolProp::input_pairs` is an enum.

A `format_as` overload should be provided for every enum that needs to be formatted, such as:

```cpp
#include <fmt/format.h>

enum class Foo {
    BAR = 0
};

auto format_as(Foo f) { return fmt::underlying(f); }
```

FtZPetruska:
### Requirements

### Description of the Change

With the recent breaking changes made with fmt, it is no longer possible to build CoolProp using the latest release of fmt (10.0.0).

The main issue is that enums no longer get automatically converted to int, even when using `sprintf` and the `%d` specifier explicitly.

To resolve this issue, every enum that is used in a format call needs to have a `format_as` overload **in the same namespace**. This feature is only available since fmt 9.0.0, hence the `FMT_VERSION >= 90000` guard.

As suggested, the module was also updated to the latest version.

### Benefits

Support newer versions of fmt, without dropping support for older versions.

### Possible Drawbacks

There is sort of a pattern that cannot be abstracted away with a template (to my knowledge) and that is repeated for every enum:
```cpp
int format_as(TheEnum foo) { return fmt::underlying(foo) }
```

Updating the fmt submodule beyond 4.x raises the minimum to  GCC 4.8, Clang 3.4, and MSVC 19.0 (2015). This change is not strictly necessary so it can be reverted if needed.

### Verification Process

Admittedly, I am not a CoolProp user, so the only testing that was done was ensuring the project successfully built with a newer version of fmt.

### Applicable Issues

Fixes #2251

### DIFF
--- a/include/CPstrings.h
+++ b/include/CPstrings.h
@@ -74,17 +74,16 @@ inline bool endswith(const std::string& s1, const std::string& s2) {
 #if defined(NO_FMTLIB)
 // Missing string formatting function, this old guy is needed for ancient gcc compilers on PowerPC for VxWorks
 inline std::string format(const char* fmt, ...);
+#elif FMT_VERSION >= 50000
+template <typename... Args>
+inline std::string format(const char* format_str, const Args&... args) {
+    return fmt::sprintf(format_str, args...);
+}
 #else
-// Missing std::string formatting function - provided by the fmtlib library
 inline std::string format(const char* format, fmt::ArgList args) {
     return fmt::sprintf(format, args);
 }
 FMT_VARIADIC(std::string, format, const char*)
-// For latest FMTLIB
-/*template <typename... Args>
-    inline std::string format(const char *format_str, const Args & ... args) {
-        return fmt::sprintf(format_str, args);
-    }*/
 #endif
 
 // Missing string split - like in Python
--- a/include/CoolPropFluid.h
+++ b/include/CoolPropFluid.h
@@ -562,5 +562,19 @@ class CoolPropFluid
     };
 };
 
+#if !defined(NO_FMTLIB) && FMT_VERSION >= 90000
+static int format_as(ViscosityDiluteVariables::ViscosityDiluteType type) {
+    return fmt::underlying(type);
+}
+
+static int format_as(TransportPropertyData::ViscosityHardcodedEnum viscosity) {
+    return fmt::underlying(viscosity);
+}
+
+static int format_as(TransportPropertyData::ConductivityHardcodedEnum conductivity) {
+    return fmt::underlying(conductivity);
+}
+#endif
+
 } /* namespace CoolProp */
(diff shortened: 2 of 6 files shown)


ibell:
Thanks for this! We can also consider to instead force casts of enum to int for format, but function, but this seems harmless enough.

One thing I am worried about, will all compilers accept your perfect forwarding with C++11 standard set?  Parameter packs are part of c++11: https://en.cppreference.com/w/cpp/language/parameter_pack

ibell:
GCC 4.8, Clang 3.4, and MSVC 19.0 (2015) are pretty old now, so I feel good with this change. Merging.

Thank you!

PR #2270, opened 2023-07-26
Fix failing python_cibuildwheel workflows

Arnab-Mahanti:
This may fix the failing python_cibuildwheel workflows.
Changed to older Cython version to fix `cpdef` deprication.


ibell:
Better would be to fix the code to compile with Cython 3.0, but this is a good start for now.

Thanks!

Arnab-Mahanti:
This should be Cython<3.0 instead of Cython<=3.0

PR #2271, opened 2023-07-26
Revert "Fix failing python_cibuildwheel workflows" and update

Arnab-Mahanti:
Reverts CoolProp/CoolProp#2270
and update Cython version to be strictly less than 3.0


ibell:
Right I missed that.

PR #2275, opened 2023-08-04
Fix Missing SWIG Wrapper

### LINKED ISSUE #2274: SWIG misses apply_simple_mixing_rule when building Java wrapper
### Description

I'm attempting to follow through the examples ([http://www.coolprop.org/coolprop/Cubics.html#adding-your-own-fluids](http://www.coolprop.org/coolprop/Cubics.html#adding-your-own-fluids)) for adding a custom fluid and then obtaining mixture properties. I'm using a Java wrapper for CoolProp. When adding a custom fluid it seems that this is an essential step to set the dummy interaction parameters with apply_simple_mixing_rule. Unfortunately, the method apply_simple_mixing_rule appears to be missing from CoolProp.Java generated by SWIG. Without it, PropsSI will throw an error that it could not match the binary pair. I presume that this is because no binary pair data has been set between the custom fluid and other fluids?

I thought the issue might be related to my installation of cmake, SWIG etc. as I've compiled the CoolProp code myself in order to assign a specific package (org.coolprop) to the generated code. Yet when I look back at the pre-compiled Java wrappers from 6.4.1 and earlier, this method is also missing. It appears that for some reason SWIG is not including apply_simple_mixing_rule in the JNI stubs.

### Steps to Reproduce

1. Code to add custom fluid in Java:
    ```java
   String fluidstring = "[\n  {\n    \"CAS\": \"000-0-00\",\n    \"Tc\": 400.0,\n    \"Tc_units\": \"K\",\n"
                + "    \"acentric\": 0.1,\n    \"aliases\": [\n    ],\n    \"molemass\": 0.04,\n"
                + "    \"molemass_units\": \"kg/mol\",\n    \"name\": \"FAKEFLUID\",\n    \"pc\": 4000000.0,\n"
                + "    \"pc_units\": \"Pa\"\n  }\n]";
   CoolProp.add_fluids_as_JSON("PR", fluidstring);
   System.out.println(CoolProp.PropsSI("T", "P", 101325, "Q", 0.0, "SRK::FAKEFLUID")+ " K");
    ```
   This step works as expected and produces the text output "247.0950826605273 K"
2. Method CoolProp.apply_simple_mixing_rule does not exist, so try skipping and just do a mixture calculation using our new custom fluid.
   ```java
    System.out.println(CoolProp.PropsSI("T", "P", 101325, "Q", 0.0, "HEOS::FAKEFLUID-SRK[0.3]&Ethane-SRK[0.7]"));
    ```
   This throws an error.

**Expected behavior:** Ideally the apply_simple_mixing_rule method would be accessible via Java using SWIG. Otherwise, in the event that the interaction parameters cannot be set, defaulting to a value of 1.0 for all interaction pairs between an existing fluid and a custom fluid might be sensible?

**Actual behavior:** Throws an error: java.lang.RuntimeException: Initialize failed for backend: "HEOS", fluid: "FAKEFLUID-SRK&Ethane-SRK" fractions "[ 0.3000000000, 0.7000000000 ]"; error: Could not match the binary pair [000-0-00,74-84-0] - for now this is an error.

### Versions

**CoolProp Version:** 6.5.0dev. 
**Operating System and Version:** Windows 11
**Access Method:** Java wrapper compiled using cmake 3.26.4 and swigwin-4.1.1. 

### Additional Information

This is my first experience with SWIG and C++, so I'm not sure where the best place to look is. I noticed that in /src/pybind11_interface.cxx there is an include at the top of the file for "Backends/Helmholtz/MixtureParameters.h". This is not included in CoolProp.i, so I tried adding it there, and to SWIG_DEPENDENCIES in CMakeLists.txt. After rebuilding and recompiling, neither change made apply_simple_mixing_rule appear.

I'm wondering if it's because the header files listed in CoolProp.i are all located in the /include/ folder, and that I should therefore be using a path like "../src/Backends/Helmholtz/MixtureParameters.h"?

pkirkham:
Added pass-through function to `CoolProp.cpp` for apply_simple_mixing_rule which was not recognised by SWIG and included in wrappers. Had to rename original function to apply_basic_mixing rule to avoid naming conflict. Using a similar approach as `CoolProp::add_fluids_as_JSON` and keeping the same function name failed to compile. Also added `HumidAirProp.h` to the SWIG_DEPENDENCIES so that these functions are included in the wrapper.

### Description of the Change

The Java Wrapper was missing the function `CoolProp::apply_simple_mixing_rule` function. It appears that SWIG was not including this function when generating the wrappers. Upon examination of earlier releases, it appears that this function has been missing for a while. Initially the workaround used was to add a pass-through function using a different name which called `apply_simple_mixing_rule`. It was suggested that if the the function name was preserved, with the arguments mirrored, then a PR could be considered. It was suggested that the `add_fluids_as_JSON` in `CoolProp.cpp` could be used as an example.

Attempts to use a similar approach as `add_fluids_as_JSON` were unsuccessful. Adding a new namespace around the function being called in `MixtureParameters.h` did not work. The only way this would successfully compile and run, and have SWIG pick up the function name for inclusion in the wrapper, was to change the name of the function that is called in `MixtureParameters.h`.

### Benefits

SWIG now correctly identifies the `apply_simple_mixing_rule` function and creates a wrapper for it in the CoolProp.java and CoolPropJNI.java. This means that custom fluids can be used in mixtures.

### Possible Drawbacks

The original function name in `MixtureParameters.h` has been changed. Any code that relies on calling that specifically, and in particular `mixturebinarypairlibrary.add_simple_mixing_rule(identifier1, identifier2, rule)` will no longer work. Also, whilst the renamed function `add_basic_mixing_rule` is not exposed in the Java wrapper created, it may appear in other wrappers and is exposed in the C++. Thus it appears there is a duplicate and redundant function; not DRY code.

### Verification Process

  No new functionality,

  Compiled changes successfully and ran test case using the newly generated Java wrapper to add a custom fluid, create the dummy interaction parameters using the new function `add_simple_mixing_rule`, and finally added fluid to a mixture as per (http://www.coolprop.org/coolprop/Cubics.html#adding-your-own-fluids) -- REFERENCE PAGE. Results obtained match those published on the REFERENCE PAGE.

  After compilation, I have not been able to run the commands:

  ```
  # Run the integration tests
  ctest --extra-verbose
  ```
  I receive the error "Test not available without configuration.  (Missing "-C <config>"?)". For the moment I've ignored this. However, I'm conscious that the changes have not been thoroughly tested and I've not been able to verify that the change has not introduced any regressions.

- *Describe the actions you performed (e.g. text you typed, commands you ran, etc.), and describe the results you observed.*  
Ran test suite based on standard CoolProp example functions in the generated `Example.java` plus code to add a custom fluid, apply the mixing rule and access a property for a mixture that includes the custom fluid. I have also tested with a range of different fluid compositions in my own fluid test suite for hydrocarbon fluids which range from dry gases through to crude oils. These mixtures generally include a C7+ pseudo-component which I usually then also split into 7 further pseudos. No errors were encountered with the nearly 250 different pseudos that were generated -- all had dummy interaction parameters successfully set using `add_simple_mixing_rule`.

Code to replicate results from REFERENCE PAGE:

```java
System.out.println("*********** CUSTOM FLUID AND MIXTURE *****************");
String fluidstring = "[\n  {\n    \"CAS\": \"000-0-00\",\n    \"Tc\": 400.0,\n    \"Tc_units\": \"K\",\n"
    + "    \"acentric\": 0.1,\n    \"aliases\": [\n    ],\n    \"molemass\": 0.04,\n"
    + "    \"molemass_units\": \"kg/mol\",\n    \"name\": \"FAKEFLUID\",\n    \"pc\": 4000000.0,\n"
    + "    \"pc_units\": \"Pa\"\n  }\n]";
CoolProp.add_fluids_as_JSON("PR", fluidstring);
System.out.println("Boiling temperature of custom fluid at 101325 Pa: " + CoolProp.PropsSI("T", "P", 101325, "Q", 0.0, "SRK::FAKEFLUID")+ " K");
String cas_ff = "000-0-00";
String cas_c2 = CoolProp.get_fluid_param_string("Ethane", "CAS");
CoolProp.set_config_bool(configuration_keys.OVERWRITE_BINARY_INTERACTION, true);
CoolProp.apply_simple_mixing_rule(cas_ff, cas_c2, "Lorentz-Berthelot");
System.out.println("Boiling temperature of mixture with custom fluid at 101325 Pa: " + CoolProp.PropsSI("T", "P", 101325, "Q", 0.0, "HEOS::FAKEFLUID-SRK[0.3]&Ethane-SRK[0.7]") + " K");
```

Results of this test are:

image

### Applicable Issues

``Closes #2274``

### DIFF
--- a/src/Backends/Helmholtz/MixtureParameters.cpp
+++ b/src/Backends/Helmholtz/MixtureParameters.cpp
@@ -195,7 +195,7 @@ class MixtureBinaryPairLibrary
         }
     }
     /// Add a simple mixing rule
-    void add_simple_mixing_rule(const std::string& identifier1, const std::string& identifier2, const std::string& rule) {
+    void add_basic_mixing_rule(const std::string& identifier1, const std::string& identifier2, const std::string& rule) {
         // Get the empty dictionary to be filled by the appropriate reducing parameter filling function
         Dictionary dict;
 
@@ -288,8 +288,8 @@ static MixtureBinaryPairLibrary mixturebinarypairlibrary;
 static MixtureBinaryPairLibrary mixturebinarypairlibrary_default;
 
 /// Add a simple mixing rule
-void apply_simple_mixing_rule(const std::string& identifier1, const std::string& identifier2, const std::string& rule) {
-    mixturebinarypairlibrary.add_simple_mixing_rule(identifier1, identifier2, rule);
+void apply_basic_mixing_rule(const std::string& identifier1, const std::string& identifier2, const std::string& rule) {
+    mixturebinarypairlibrary.add_basic_mixing_rule(identifier1, identifier2, rule);
 }
 
 std::string get_csv_mixture_binary_pairs() {
--- a/src/Backends/Helmholtz/MixtureParameters.cpp
+++ b/src/Backends/Helmholtz/MixtureParameters.cpp
@@ -287,14 +287,10 @@ static MixtureBinaryPairLibrary mixturebinarypairlibrary;
 // A fixed parameter library containing the default values
 static MixtureBinaryPairLibrary mixturebinarypairlibrary_default;
 
-namespace MixParameters {
-
 /// Add a simple mixing rule
 void apply_simple_mixing_rule(const std::string& identifier1, const std::string& identifier2, const std::string& rule) {
     mixturebinarypairlibrary.add_simple_mixing_rule(identifier1, identifier2, rule);
 }
-
-} /* namespace MixParameters */
 
 std::string get_csv_mixture_binary_pairs() {
 
--- a/src/Backends/Helmholtz/MixtureParameters.cpp
+++ b/src/Backends/Helmholtz/MixtureParameters.cpp
@@ -195,7 +195,7 @@ class MixtureBinaryPairLibrary
         }
     }
     /// Add a simple mixing rule
-    void add_basic_mixing_rule(const std::string& identifier1, const std::string& identifier2, const std::string& rule) {
+    void add_simple_mixing_rule(const std::string& identifier1, const std::string& identifier2, const std::string& rule) {
         // Get the empty dictionary to be filled by the appropriate reducing parameter filling function
         Dictionary dict;
 
@@ -287,10 +287,14 @@ static MixtureBinaryPairLibrary mixturebinarypairlibrary;
 // A fixed parameter library containing the default values
 static MixtureBinaryPairLibrary mixturebinarypairlibrary_default;
 
+namespace MixParameters {
+
 /// Add a simple mixing rule
-void apply_basic_mixing_rule(const std::string& identifier1, const std::string& identifier2, const std::string& rule) {
-    mixturebinarypairlibrary.add_basic_mixing_rule(identifier1, identifier2, rule);
+void apply_simple_mixing_rule(const std::string& identifier1, const std::string& identifier2, const std::string& rule) {
+    mixturebinarypairlibrary.add_simple_mixing_rule(identifier1, identifier2, rule);
 }
+
+} /* namespace MixParameters */
 
 std::string get_csv_mixture_binary_pairs() {
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,7 +272,8 @@ set(SWIG_DEPENDENCIES
     ${CMAKE_CURRENT_SOURCE_DIR}/include/CoolProp.h
     ${CMAKE_CURRENT_SOURCE_DIR}/include/AbstractState.h
     ${CMAKE_CURRENT_SOURCE_DIR}/include/Configuration.h
-    ${CMAKE_CURRENT_SOURCE_DIR}/include/PhaseEnvelope.h)
+    ${CMAKE_CURRENT_SOURCE_DIR}/include/PhaseEnvelope.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/include/HumidAirProp.h)
 
 set(COOLPROP_APP_SOURCES
     ${APP_SOURCES}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,8 +272,7 @@ set(SWIG_DEPENDENCIES
     ${CMAKE_CURRENT_SOURCE_DIR}/include/CoolProp.h
     ${CMAKE_CURRENT_SOURCE_DIR}/include/AbstractState.h
     ${CMAKE_CURRENT_SOURCE_DIR}/include/Configuration.h
-    ${CMAKE_CURRENT_SOURCE_DIR}/include/PhaseEnvelope.h
-    ${CMAKE_CURRENT_SOURCE_DIR}/include/HumidAirProp.h)
+    ${CMAKE_CURRENT_SOURCE_DIR}/include/PhaseEnvelope.h)
 
 set(COOLPROP_APP_SOURCES
     ${APP_SOURCES}
--- a/include/CoolProp.h
+++ b/include/CoolProp.h
@@ -131,6 +131,14 @@ bool is_valid_fluid_string(const std::string& fluidstring);
 bool add_fluids_as_JSON(const std::string& backend, const std::string& fluidstring);
 
 /**
+ * @brief Apply Lorentz-Berthelot mixing rule (all interaction parameters = 1.0) for a given binary pair
+ * @param identifier1 The CAS # (or name) for the first fluid
+ * @param identifier2 The CAS # (or name) for the second fluid
+ * @param rule The simple mixing rule to be used ("linear", "Lorentz-Berthelot")
+ */
+void apply_simple_mixing_rule(const std::string& identifier1, const std::string& identifier2, const std::string& rule);
+
+/**
     \brief Set the reference state based on a string representation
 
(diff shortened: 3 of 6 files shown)


ibell:
I think changing the core code of CoolProp is not the right solution. There must be something else afoot that does not permit the compilation. Can you please try with the same-named function and show what compilation error you get?

pkirkham:
Understood. It's not ideal. I'll educate myself more about SWIG and the compilation process and try to determine the issue.

pkirkham:
I managed to get a pass through function, with no changes to the function names, to compile using the same approach as `add_fluids_as_JSON`. I feel this is a much better solution. I have re-tested the wrapper code and .dll that is generated and it is working as expected.

src/Backends/Helmholtz/MixtureParameters.cpp
```diff
@@ -287,11 +287,15 @@ static MixtureBinaryPairLibrary mixturebinarypairlibrary;
 // A fixed parameter library containing the default values
 static MixtureBinaryPairLibrary mixturebinarypairlibrary_default;
 
+namespace MixParameters {
```
ibell, line 290: This namespace will break existing code. Please find a way to avoid it.

ibell (changes requested):
The addition of the namespace will break existing code. Please show what errors you get when you add only the passthrough function in CoolProp.h and CoolProp.cpp, and also remove changes to CoolProp.i

pkirkham:
> Please show what errors you get when you add only the passthrough function in CoolProp.h and CoolProp.cpp

Compiling with Visual Studio 2022 on Windows 11:

D:\Dropbox\GitHub\CoolProp\src\CoolProp.cpp(681): warning C4717: 'CoolProp::apply_simple_mixing_rule': recursive on all
 control paths, function will cause runtime stack overflow [D:\Dropbox\GitHub\CoolProp\build\CoolProp.vcxproj]
MixtureParameters.obj : error LNK2005: "void __cdecl CoolProp::apply_simple_mixing_rule(class std::basic_string<char,st
ruct std::char_traits<char>,class std::allocator<char> > const &,class std::basic_string<char,struct std::char_traits<c
har>,class std::allocator<char> > const &,class std::basic_string<char,struct std::char_traits<char>,class std::allocat
or<char> > const &)" (?apply_simple_mixing_rule@CoolProp@@YAXAEBV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2
@@std@@00@Z) already defined in CoolProp.obj [D:\Dropbox\GitHub\CoolProp\build\CoolProp.vcxproj]

Compiling with GCC on Linux:

/usr/bin/ld: CMakeFiles/CoolProp.dir/src/Backends/Helmholtz/MixtureParameters.cpp.o: in function \`CoolProp::apply_simple_mixing_rule(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
MixtureParameters.cpp:(.text+0xe40): multiple definition of \`CoolProp::apply_simple_mixing_rule(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'; CMakeFiles/CoolProp.dir/src/CoolProp.cpp.o:CoolProp.cpp:(.text+0x320): first defined here
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/CoolProp.dir/build.make:759: libCoolProp.so] Error 1
make[1]: *** [CMakeFiles/Makefile2:108: CMakeFiles/CoolProp.dir/all] Error 2
make: *** [Makefile:141: all] Error 2

Unsurprisingly it appears that we can't have the same function `apply_simple_mixing_rule` in CoolProp.cpp and MixtureParameters.cpp, hence the use of a different function name or namespace.

Perhaps I'm not properly understanding what you mean by using a passthrough function. The implementation is very similar to `add_fluids_as_JSON` where it appears there is also a nested namespace for `CubicLibrary` within the `CoolProp` namespace in CubicsLibrary.cpp.

All said and done, it seems the next step is to try to figure out why SWIG is not wrapping the `apply_simple_mixing_rule` in MixtureParameter.cpp and avoid the use of any passthrough function at all.

src/Backends/Helmholtz/MixtureParameters.cpp
pkirkham: Removed change to namespace and all other changes to the code. After learning more about how SWIG works, I realised the answer was to simply add the missing method into the CoolProp.i SWIG interface file and it now gets wrapped. SWIG won't recursively wrap methods in the header files, and thus it doesn't see the `apply_simple_mixing_rule` method unless it is explicitly defined.

ibell:
I think we are good now, right? It looks like you have figured out how to do this properly

pkirkham:
Correct. This is the proper way to fix the missing wrapping of the function by SWIG. The previous changes were less than satisfactory. I'm much happier with this approach. I've tested that the function is working properly when compiled using this approach.

PR #2286, opened 2023-08-23
Add fugacity functions needed for compatability with `CoolProp.jl` pkg

### LINKED ISSUE #2284: Getting fugacity from the shared library
### Description

We are trying to use CoolProp through the Julia interface, and there seems to be no way to get the fugacity from the shared library (http://www.coolprop.org/_static/doxygen/html/_cool_prop_lib_8h.html) since it is not included as keyed output or returned in a separate wrapped function. We are unsure of how to wrap the existing fugacity function in Julia without using PyCall.

### Steps to Reproduce

`CoolProp.AbstractState_keyed_output(handle, CoolProp.get_param_index("fugacity"))`

**Expected behavior:** fugacity (or fugacity vector) returned

**Actual behavior:** `CoolProp: Unknown parameter: fugacity`

or

`CoolProp.AbstractState_get_fugacity(handle, i)`

**Expected behavior:** fugacity of species i returned

**Actual behavior:** `ERROR: UndefVarError: AbstractState_get_fugacity not defined`

### Versions

**CoolProp Version:** 6.5.0  
**Access Method:** Julia wrapper

chris-hampel-CA:
### Description of the Change

Fugacity lookups `get_fugacity`, `get_fugacity_coefficient`, and `get_fugacity_coefficients` cannot be wrapped with Julia functions as they currently are in this repo. This is in effort to enable using `CoolProp.jl` to return fugacity which is currently  a missing feature (an issue was created [here](https://github.com/CoolProp/CoolProp.jl/issues/31) in `CoolProp.jl`). 

### Benefits

Allows fugacity calls to be wrapped with Julia code directly in `CoolProp.jl`. We have found that using low-level Julia-wrapper commands is approximately 10x faster than using `Pycall` within Julia to call CoolProp low-level commands. This is a surmountable benefit when incorporating CoolProp calls with iterative numerical models of fluid systems within the Julia language that require fugacity lookups.

### Possible Drawbacks

Unsure of any possible drawbacks.

### Verification Process

TODO: 
- [x] test that the C++ package works
- [ ] test that the envisioned CoolProp.jl Julia wrapper functions work properly with the new C++ code

### Applicable Issues

Closes #2284

### DIFF
--- a/include/CoolPropLib.h
+++ b/include/CoolPropLib.h
@@ -369,6 +369,38 @@ EXPORT_CODE void CONVENTION AbstractState_get_mole_fractions_satState(const long
                                                                       const long maxN, long* N, long* errcode, char* message_buffer,
                                                                       const long buffer_length);
 /**
+     * @brief Return the fugacity of the i-th component of the mixture.
+     * @param handle The integer handle for the state class stored in memory
+     * @param i i-th component of the mixture
+     * @param errcode The errorcode that is returned (0 = no error, !0 = error)
+     * @param message_buffer A buffer for the error code
+     * @param buffer_length The length of the buffer for the error code
+     * @return 
+     */
+EXPORT_CODE double CONVENTION AbstractState_get_fugacity(const long handle, const std::size_t i, long* errcode, char* message_buffer,
+                                                         const long buffer_length);
+/**
+     * @brief Return the fugacity coefficient of the i-th component of the mixture.
+     * @param handle The integer handle for the state class stored in memory
+     * @param i i-th component of the mixture
+     * @param errcode The errorcode that is returned (0 = no error, !0 = error)
+     * @param message_buffer A buffer for the error code
+     * @param buffer_length The length of the buffer for the error code
+     * @return 
+     */
+EXPORT_CODE double CONVENTION AbstractState_get_fugacity_coefficient(const long handle, const std::size_t i, long* errcode, char* message_buffer,
+                                                                      const long buffer_length);
+// TODO: update once the AbstractState_get_fugacity_coefficients function is verified to be correct
+// /**
+//      * @brief Return a vector of the fugacity coefficients for all components in the mixture.
+//      * @param handle The integer handle for the state class stored in memory
+//      * @param errcode The errorcode that is returned (0 = no error, !0 = error)
+//      * @param message_buffer A buffer for the error code
+//      * @param buffer_length The length of the buffer for the error code
+//      * @return 
+//      */
+// EXPORT_CODE double CONVENTION AbstractState_get_fugacity_coefficients(const long handle, long* errcode, char* message_buffer, const long buffer_length);
+/**
      * @brief Update the state of the AbstractState
      * @param handle The integer handle for the state class stored in memory
      * @param input_pair The integer value for the input pair obtained from XXXXXXXXXXXXXXXX
--- a/include/CoolPropLib.h
+++ b/include/CoolPropLib.h
@@ -377,7 +377,7 @@ EXPORT_CODE void CONVENTION AbstractState_get_mole_fractions_satState(const long
      * @param buffer_length The length of the buffer for the error code
      * @return 
      */
-EXPORT_CODE double CONVENTION AbstractState_get_fugacity(const long handle, const std::size_t i, long* errcode, char* message_buffer,
+EXPORT_CODE double CONVENTION AbstractState_get_fugacity(const long handle, const long i, long* errcode, char* message_buffer,
                                                          const long buffer_length);
 /**
      * @brief Return the fugacity coefficient of the i-th component of the mixture.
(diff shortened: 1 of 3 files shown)


include/CoolPropLib.h
```diff
@@ -368,6 +368,38 @@ EXPORT_CODE void CONVENTION AbstractState_get_mole_fractions(const long handle,
...
+/**
+     * @brief Return the fugacity of the i-th component of the mixture.
+     * @param handle The integer handle for the state class stored in memory
+     * @param i i-th component of the mixture
+     * @param errcode The errorcode that is returned (0 = no error, !0 = error)
+     * @param message_buffer A buffer for the error code
+     * @param buffer_length The length of the buffer for the error code
+     * @return 
+     */
+EXPORT_CODE double CONVENTION AbstractState_get_fugacity(const long handle, const std::size_t i, long* errcode, char* message_buffer,
```
ibell, line 380: everywhere you want long instead of std::size_t. size_t is not a standard C type

src/CoolPropLib.cpp
```diff
@@ -604,6 +604,43 @@ EXPORT_CODE void CONVENTION AbstractState_get_mole_fractions_satState(const long
...
+        shared_ptr<CoolProp::AbstractState>& AS = handle_manager.get(handle);
+        return AS->fugacity_coefficient(i);
+    } catch (...) {
+        HandleException(errcode, message_buffer, buffer_length);
+    }
+    return _HUGE;
+}
+// TODO: verify this is correct
+// Am I returning vectors correctly??
+// EXPORT_CODE vector<std::double> CONVENTION AbstractState_get_fugacity_coefficients(const long handle, const std::size_t i, long* errcode, char* message_buffer,
```
ibell, line 633: I would say let's leave this out until implemented throughout CoolProp

ibell (changes requested):

include/CoolPropLib.h
chris-hampel-CA: the functions `fugacity_coefficient` and `fugacity` are defined to expect `std::size_t i`. Does it not have to be the same? 

from `AbstractState.cpp` 
```
double AbstractState::fugacity(std::size_t i)
double AbstractState::fugacity_coefficient(std::size_t i)
```

src/CoolPropLib.cpp
chris-hampel-CA: Ok, I will remove the commented block

src/CoolPropLib.def
```diff
@@ -15,6 +15,8 @@ EXPORTS
   AbstractState_update_and_1_out = _AbstractState_update_and_1_out@40
   AbstractState_update_and_5_out = _AbstractState_update_and_5_out@56
   AbstractState_update_and_common_out = _AbstractState_update_and_common_out@52
+  AbstractState_get_fugacity = _AbstractState_get_fugacity@20
+  AbstractState_get_fugacity_coefficient = _AbstractState_get_fugacity_coefficient@20
```
chris-hampel-CA, line 19: I am also wondering if we need to add export lines to the `CoolPropLib.def` file. I noticed some functions in `CoolPropLib.cpp` that are missing from the `.def` file, but if it is required, can you check the integer value used after the `@`? I believe it is 4 bytes per arg

chris-hampel-CA:
Also, I noticed there were not any tests for any functions that start with `AbstractState_`. Should there be, and if so, how can I best test that the new functions are working properly?

src/CoolPropLib.def
ibell: Hmm good question, I don't think I have touched that in ages. Can you check if anywhere in the CMakeLists.txt references this def file? I think it is not used anymore...

include/CoolPropLib.h
ibell: No, the external interface should only use C datatypes, and cast C++ ones to C ones as needed

ibell:
> Also, I noticed there were not any tests for any functions that start with `AbstractState_`. Should there be, and if so, how can I best test that the new functions are working properly?

It would definitely be a good idea to have some tests, but I am not sure the best place to put them to be honest. Might be best to have a Python test script since calling shared libraries from Python is so easy.

src/CoolPropLib.def
chris-hampel-CA: There is not a reference to the `.def` file in the the file you mention. I will go ahead and remove this. Other functions written within `CoolPropLib.cpp` are left out of the `.def` file and work properly already.

include/CoolPropLib.h
chris-hampel-CA: Ok thanks for clarifying. I have made that change as you requested, so I will resolve this thread.

src/CoolPropLib.def
chris-hampel-CA: I have removed them now, so I am resolving this thread

chris-hampel-CA:
> > Also, I noticed there were not any tests for any functions that start with `AbstractState_`. Should there be, and if so, how can I best test that the new functions are working properly?
> 
> It would definitely be a good idea to have some tests, but I am not sure the best place to put them to be honest. Might be best to have a Python test script since calling shared libraries from Python is so easy.

We are working on testing this now either directly in C++ or perhaps with your Python suggestion. Assuming that all goes well, we will report back with evidence of the success. I believe all else is in order for completing the pull request. 

I am also curious about when/how quickly you all decide to release new versions. Ideally, a new version could be released after this is merged. The `CoolProp.jl` repo has a pkg dependency called `CoolProp_jll` which I assume gets released once a new version of this repo is released. Is that correct? If it's not possible to do the release after the completion of this pull request, is there a way to release `CoolProp_jll` locally such that `CoolProp.jl` can reference our changes.

longemen3000:
I normally update the CoolProp_jll package when it is released, it only needs a pull request to Yggdrasil.jl specifying the source of the new version. (it takes about a day after the release of a new version of CoolProp)

jyoungca:
Wrote a quick test file, and confirmed that the correct fugacity and fugacity coefficients are returned (for N2 and CO2 at [0.5, 0.5] mole frac, 250 K, and 1e6 Pa):
image

ibell:
Looks good, thank you

chris-hampel-CA:
@ibell Thanks again for the review! 

I was curious what pull requests qualify for being made available in a new release. 

My motivation stems from wanting to be able to use code for calling fugacity in the downstream pkg, `CoolProp.jl`. Recreating the development chain locally is a bit cumbersome while a release would make it easy and allow me to develop the julia wrapper in CoolProp.jl for all others to use too. 

To my knowledge, testing this new code locally in CoolProp.jl would require developing this chain: CoolProp -> Yggdrasil -> CoolProp_jll -> CoolProp.jl

ibell:
> @ibell Thanks again for the review!
> 
> I was curious what pull requests qualify for being made available in a new release.
> 
> My motivation stems from wanting to be able to use code for calling fugacity in the downstream pkg, `CoolProp.jl`. Recreating the development chain locally is a bit cumbersome while a release would make it easy and allow me to develop the julia wrapper in CoolProp.jl for all others to use too.
> 
> To my knowledge, testing this new code locally in CoolProp.jl would require developing this chain: CoolProp -> Yggdrasil -> CoolProp_jll -> CoolProp.jl

All the PR that are merged will be included. But the timeline for release is always uncertain. Can we try to add all the desired ``AbstractState_XXXX`` methods that are desired (or might be desired) for this version?

PR #2288, opened 2023-08-26
Google Apps Scipts function and python flask api.

qnikil7:
### Description of the Change

Coolprop for google sheets.

I wanted to use coolprop in google sheets and wasn't able to find a way to call shared libraries in google sheet. So this is an alternate approach.

I have created a web app in python using flask which which is called through an api from google sheets using GET. I have hosted this currently in python anywhere. Link for [google sheet.](https://docs.google.com/spreadsheets/d/1vU6Hx7oFZxdVOvrh6zusoLPGNxDD-TWY5mbbEt8bUOU/edit#gid=0)

Currently I have added only PropsSI and HAPropsSI functions.

### Benefits

Use CoolProp in google sheets.

### Possible Drawbacks

Sometimes the functions are slow to load.

### Verification Process

- There is a google sheet that demonstrates the functionality.
- The google sheets have the https links to get the values.
- test_api script to verify the values.

### Applicable Issues

- Probably calling a shared library using Apps Script (If possible) would be preferred.
- If there is interest in this, probably hosting it somewhere and creating an addin in google workspace.
- Also I selected python flask based on ease of use but may not be the most efficient.

### DIFF
--- a/wrappers/GoogleAppsScript/Code.js
+++ b/wrappers/GoogleAppsScript/Code.js
@@ -1,0 +1,72 @@
+function PropsSI(output,property_1,value_1,property_2,value_2,fluid, opt_url) {
+
+  /* This function fetches the state property for a fluid from the web api
+
+  Args:
+  output: str, name of the output property
+  property_1: str, name of the first input property
+  value_1: float, value of the first input property
+  property_2: str, name of the second input property
+  value_2: float, value of the second input property
+  fluid: str, name of the fluid
+  opt_url: str optional, url of the web api
+
+  Returns:
+  float, value of the state property
+  
+  */
+
+  if (opt_url==null) {
+    var baseUrl = "https://qnikil7.pythonanywhere.com"
+  } else {
+    var baseUrl = opt_url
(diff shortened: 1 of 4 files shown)


ibell:
This is a very good idea and would be generally useful to have this service around and well-documented and supported. I found getting things working on PA to be not always so easy. I remember I had to re-write the app from scratch at least one time.

ibell:
This is a nice idea! I wonder about accepting the PR though given that we cannot guarantee that your python anywhere instance will stay up. But it would indeed be handy to have a running API somewhere. Flask is a perfectly fine tool for this job.

qnikil7:
Thank you for the feedback.

I agree that the python anywhere instance would not be reliable. Moreover, would be good if its running somewhere where multiple people have access in case there are issues. I am not sure where to host this though.

On the google sheets side, would be good if its a google sheets add-in, maintained as a google cloud project, again multiple people having access to maintain.

I would like to improve it with better documentation and handling errors. Marking this PR as draft.

Issue #2307, opened 2023-10-19
Feature Request: Implementation of the novel Refrigerant R1336mzz(e) into HEOS Backend


chrishauptmannAtEsiGroup:
Dear CoolProp team,

Please consider adding R1336mzz(e) to the HEOS-Backend. 

Do you think it's possible for me to do the implementation with the sources provided below?

- https://www.nist.gov/publications/helmholtz-energy-equation-state-trans-111444-hexafluoro-2-butene-r-1336mzze

- https://tsapps.nist.gov/publication/get_pdf.cfm?pub_id=935812

longemen3000:
the suplementary information on the first link has a json file that has all the necessary information to add that fluid to CoolProp. if you are using CoolProp along a programing language, you can load that json directly: http://www.coolprop.org/coolprop/HighLevelAPI.html#adding-fluids

ibell:
@longemen3000 is right. I just didn't add it to CoolProp, but PR adding would be welcomed. Make sure to update the reference in the BibTeX library

chrishauptmannAtEsiGroup:
Thank you for your recommendations, it helped a lot. =) 

I implemented R1336mzze succesfully to my local CoolProp. 
But for my use case i also need additional equations e.g. for viscosity and conductivity, the warnings of CoolProp say's this is not implemented. 

Is it correct that the R1336mzze.json file of Akasaka-IJT-2023 has not this equations, or is it my fault in implementation? 

Furthermore i implemented the R1336mzze.FLD File to Refprop v10.0. This worked really well for my application, so currently i have no need to implement it in HEOS. If there would be a sufficient solution for the viscosity and conductivity, i would do it. 

Thank you so much for your support. =)

ibell:
Well, an EOS has nothing to do with the transport properties (viscosity and
thermal conductivity), at its core.

The FLD file has an ECS model that you could translate to CoolProp format,
or you could implement one the entropy scaling models that work quite well
for refrigerants. For instance this one:
https://pubs.acs.org/doi/abs/10.1021/acs.jced.0c01009

jowr:
I think this has been fixed in #2309 - but the transport properties are still missing though.

Issue #2127, opened 2022-04-20
Adding R1336zmm(E)


eos86:
Hello,

I would like to know if adding the equation of state of R1336zmm(E) is planned or in progress.
If not, is there any documentation of how to include it?
I was planning to use the eos by Tanaka et al. 2017 IJR.

Kind regards,

Egoi

ibell:
The new eos is complete, but not yet available in CoolProp

eos86:
nice! Do you know when it will be available?

thank you!

ibell:
When someone adds it

jfreissmann:
Hey guys, are there any updates on the addition of R1336mzz(E) (which seems to be misspelled within the issue)? I have found this NIST paper (https://doi.org/10.1021/acs.jced.6b00169) which describes the EOS between 200.15 K and 410 K.
In addition to that, is the implementation of R1336mzz(Z) planned as well? It is available within REFPROP10, so I would think that there is an EOS published somewhere.

Thanks in advance!

ibell:
There is an updated EOS for R-1336mzz(E), but it is only available for
REFPROP users currently. The one for R-1336mzz(Z) is also available for
REFPROP. Someone (not me) would need to translate the EOS parameters to
CoolProp form.

jfreissmann:
Thank you for the response. Is there any documentation or instructions for someone that has never worked on CoolProp, besides the existing source code?

ibell:
Please see this paper's SI for the JSON file for CoolProp:
https://doi.org/10.1007/s10765-022-03143-5

longemen3000:
There was a PR that added the fluid, maybe this issue can be closed?

MariaAguiar12:
Hey! What is meant by PR? I am trying to use the R1336mzz(Z) but it was not added yet. Anyone knows something about when it will be added? Thank you.

ibell:
PR: pull request, a recommended change to the code

PR #2309, opened 2023-10-24
Addition of EoS JSON of R1336mzz(E) from Akasaka-IJT-2023

jfreissmann:
As requested in Issues #2127 and #2307, this pull request aims to add R1336mzz(E) to the list of fluids of CoolProp.

I added the JSON file of the supplementary information zip-file from Akasaka-IJT-2023 (https://doi.org/10.1007/s10765-022-03143-5) in the fluids directory. Then i created the corresponding BibTeX entry and referenced it in the JSON file at parameter "BibTeX_EOS".

Please let me know, if there is anything else to do in order to complete the addition of R1336mzz(E).

### DIFF
--- a/CoolPropBibTeXLibrary.bib
+++ b/CoolPropBibTeXLibrary.bib
@@ -104,6 +104,18 @@
   Timestamp                = {2013.04.10}
 }
 
+@Article{Akasaka-IJT-2023,
+  Title                    = {{A Helmholtz Energy Equation of State for trans-1,1,1,4,4,4-Hexafluoro-2-butene [R-1336mzz(E)]}},
+  Author                   = {Ryo Akasaka and Marcia L. Huber and Luke Simoni and Eric W. Lemmon},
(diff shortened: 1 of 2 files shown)


ibell:
Can you please confirm that you can do VLE calculations over the entire temperature range? If so, I'll merge.

jfreissmann:
Sorry for the amateurish question, but from what i've understood, i will have to build the package again first, right? Could you point me towards information on how to do that? I'm mostly working with python, not C++.

ibell:
You'll need to move into wrappers/python and run:

pip install .

Or the old school:

python setup.py install

ibell:
I'm glad I checked. The saturated liquid density needs to be fixed in two places, which I did, I think we can merge.

jfreissmann:
Sorry for the extra work, I haven't found the time to validate the changes myself. Thank you for doing it anyway.

PR #2317, opened 2023-11-23
New C Interface Functions

CSchulzeTLK:
### Description of the Change

A few new C interface functions have been added. They enable accessing some information which is currently not available in the C interface. I guess they were just missing.
I did not yet add docstrings to the headers. I didn't know where to copy the descriptions from (as it surely exists somewhere in the library).

### Benefits

A more complete C interface?

### Possible Drawbacks

Future maintenance?

### Verification Process

We already use this interface in the TILMedia library. So, the interface is ready to use. However, not all failure cases (try/catch) are tested. The interface functions have been duplicated and modified from existing functions in the same file, to meet the same requirements.

### DIFF
--- a/src/CoolPropLib.cpp
+++ b/src/CoolPropLib.cpp
@@ -1029,7 +1029,7 @@ EXPORT_CODE void CONVENTION add_fluids_as_JSON(const char* backend, const char* 
 }
 
 
-EXPORT_CODE long long CONVENTION AbstractState_phase(const long handle, long *errcode, char *message_buffer, const long buffer_length) {
+EXPORT_CODE int CONVENTION AbstractState_phase(const long handle, long *errcode, char *message_buffer, const long buffer_length) {
     *errcode = 0;
     try {
         shared_ptr<CoolProp::AbstractState> &AS = handle_manager.get(handle);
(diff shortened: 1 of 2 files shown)


ibell:
Are the integer types handled properly? The C integer types should match the approach in the other exposed functions. For instance I don't think you want ``long long``?  Also, do we really want ``const * const`` rather than ``const *``? I am always confused by that.

jowr:
Thank you for the contribution. Yes, I think it makes sense to align the variable types. Could you please do that?

src/CoolPropLib.cpp
```diff
@@ -1027,3 +1027,81 @@ EXPORT_CODE void CONVENTION add_fluids_as_JSON(const char* backend, const char*
...
+
+EXPORT_CODE int CONVENTION C_is_valid_fluid_string(const char* const fluidName)
+{
+    return CoolProp::is_valid_fluid_string(fluidName);
+}
+
+EXPORT_CODE int CONVENTION C_extract_backend(const char* const fluid_string, char* backend, const long backend_length, char* fluid, const long fluid_length)
+{
+    std::string _fluid, _backend;
+    CoolProp::extract_backend(fluid_string, _backend, _fluid);
```
ibell, line 1093: This function can throw I believe, so I think you need to refactor the code with error handling like the others.

ibell (changes requested):

src/CoolPropLib.cpp
ibell: Nevermind, it can't throw.

ibell:
I guess we can merge, right?

PR #2322, opened 2023-11-28
Actions for Mathcad

jowr:
### Description of the Change

We reintroduce the Mathcad interface as requested by @henningjp and others

### Benefits

Include Mathcad in the released binaries

### Possible Drawbacks

Extra maintenance.

### Verification Process

The build seems to be fine, but @henningjp should review the actual binaries to make sure they work.


henningjp:
Thanks for figuring this out!  

Where am I looking for the binaries?  On SourceForge I don't see them in the Nightly Build and the ones in the 6.5.0 release look like the ones we uploaded manually back on 9/18 (that's the date on the files).

henningjp (commented):
Downloaded the v6.6.0 Mathcad DLL and verification sheet from Source Forge.  The newly built DLL works perfectly.

PR #2329, opened 2023-12-02
Fix the triple points densities for R-1336mzz(E)

ibell:
Oops.  Now fixed.

Consistency: [R-1336mzz(E).pdf](https://github.com/CoolProp/CoolProp/files/13537277/R-1336mzz.E.pdf)


ibell:
@jowr not sure how best to handle this because you sensibly disabled the file in 6.6.0

PR #2342, opened 2024-01-09
Include stdbool for C interface

k-collie:
### Description of the Change

Conditionally include `stdbool.h` when including `CoolPropLib.h` from C. 

### Benefits

`stdbool.h` will not need to be included before `CoolPropLib.h` in C programs. This conforms to the [include what you use](https://google.github.io/styleguide/cppguide.html#Include_What_You_Use) rule from the style guide.

### Possible Drawbacks

None I can think of.

### Verification Process

- Compiled a C99 program including `CoolPropLib.h` without including `stdbool.h` beforehand.
- Still compiles with C89 when using typedef for bool

### DIFF
--- a/include/CoolPropLib.h
+++ b/include/CoolPropLib.h
@@ -59,6 +59,10 @@
 #    endif
 #endif
 
+#ifndef __cplusplus
+#    include <stdbool.h>
+#endif
+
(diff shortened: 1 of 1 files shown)


ibell:
Nice, thank you!

PR #2351, opened 2024-02-16
Fix bug introduced by 443a2fd

### LINKED ISSUE #2349: The docs are broken since md-January
I think this commit broke the docs: https://github.com/CoolProp/CoolProp/commit/443a2fd276112a88ab42c3e74ce2aeb1c7b40764

See run 673 of the documentation builder for details, maybe @msaitta-mpr could have a look?

msaitta-mpr:
### Description of the Change
Commit 443a2fd added some extra checks for convergence in the Helmholtz PHSU flash routines. However, its checking was more aggressive than the original routine. This caused the routine to raise an exception on several values that had previously been deemed acceptable. This trickled down to cause the TTSE tables to be filled with empty values, making them less robust.

This commit allows for a solution to be acceptable if the change is small between iterations as was originally intended.

### Benefits

The docs now build properly. 

### Possible Drawbacks

None known. However we should ensure that the tests pass acceptably. 

### Verification Process

I manually ran `Web/scripts/coolprop.tabular.speed.py`  with the updated code and with the current master branch. Using master causes a crash - this was what was causing the docs the fail to build in the github pipeline. Running with this new commit causes `coolprop.tabular.speed.py` to run to completion.

### Applicable Issues

Closes #2349

### DIFF
--- a/src/Backends/Helmholtz/VLERoutines.cpp
+++ b/src/Backends/Helmholtz/VLERoutines.cpp
@@ -597,7 +597,7 @@ void SaturationSolvers::saturation_PHSU_pure(HelmholtzEOSMixtureBackend& HEOS, C
     // reset the phase for the next update.
     SatL->specify_phase(iphase_liquid);
     SatV->specify_phase(iphase_gas);
-    if (error > 1e-8){
+    if (error > 1e-8 && max_abs_value(v) > 1e-9){
         throw SolutionError(format("saturation_PHSU_pure solver was good, but went bad. Current error is %Lg", error));
     }
 }


jowr:
Thank you very much - I can review it next week!

msaitta-mpr:
@jowr - Have you had a chance to review the PR yet?

ibell:
I think this seems reasonable. Shall I merge?

henningjp:
I made this change and tested it on the issue reported in #2388 .   This indeed fixes that problem and low pressure calls are working again, especially for HEOS::Water.  It will fix this issue as well as the other issue in the docs as reported by @msaitta-mpr  above.  I updated to PR to close #2388 as well.

I recommend merging immediately @ibell & @jowr as it is affecting other behavior.

ibell:
I would say go ahead and merge if it fixes the other issues.

PR #2362, opened 2024-03-30
update Excel-wrapper for Mac docs

### LINKED ISSUE #1430: Excel wrapper instructions for Mac OS X El Capitan don't work
[The instructions](http://www.coolprop.org/dev/coolprop/wrappers/Excel/index.html#pre-compiled-binaries-for-osx) for installing the Excel wrapper in Mac OS X do not work for Excel 2016 under Mac OS X El Capitan 10.11.

I have been able to track the reason for this to the combination of two issues:

* Sandboxing 
* [SIP](https://support.apple.com/kb/TA24723?locale=en_US) blocks the setting of any environment variable which starts with `DYLD_`.

The particular issue with the instructions is the part where you set `DYLD_LIBRARY_PATH`. This is silently ignored when SIP is enabled.

```
$ ~ launchctl setenv DYLD_LIBRARY_PATH "/Users/${USER}/Library/Group Containers/UBF8T346G9.Office"
$  ~ launchctl getenv DYLD_LIBRARY_PATH
```

This issue is also referenced [here](https://github.com/soumith/cudnn.torch/issues/111) and [here](https://github.com/oracle/node-oracledb/issues/231).

### LINKED ISSUE #1575: error 53 on excel for Mac
Hey,
I need to setup CoolProp for one of my classes. Since I am a student I am not really familiar with handling in the terminal, so I am asking you for help.

I followed the Instructions on [coolprop.org](http://www.coolprop.org/coolprop/wrappers/Excel/index.html) to install the Excelwrapper but at the end of the installation error 53 comes up.

I am running Excel 15.39 on MacOS 10.13. So I tried to use the 64bit version of CoolProp but the error occurs, using the 32 bit version the same error occurs.

Would be great If somebody could try to help me.

Best Regards

### LINKED ISSUE #2040: CoolProp in Excel on Mac M1
A fully functional version of CoolProp 6.3 on an Intel mac using Excel 16.52 was imported to an M1 Mini running the same code.

I was immediately hit with Error 53 (/Users/.../libCoolProp.dylib file not found).  

~/Library/LaunchAgents/my.startup.plist was set identically as before
To my knowledge, all other settings and file locations are correct and identical to the previous installation.

(I also downloaded a fresh version of CoolProp.xlam (6.4.1), updated the references to libCoolProp.dylib as indicated in Part 4b on [http://www.coolprop.org/coolprop/wrappers/Excel/index.html#excel](url) and obtained the same results.)

Is it possible I need to rebuild libCoolProp.dylib manually rather than using the one provided from SourceForge?
(Current version yields: libCoolProp.dylib: Mach-O 64-bit dynamically linked shared library x86_64)

If so, could someone please provide explicit instructions for the rebuild?

### LINKED ISSUE #2224: The steps for adding an Excel (Version 16.43) wrapper doesn't work for MacOS (10.13.6)
The steps for adding an Excel (Version 16.43) wrapper in http://www.coolprop.org/coolprop/wrappers/Excel/index.html doesn't work for MacOS (10.13.6). Despite following all the steps, I was getting an .....Error 53: Cannot find the file..... This was not related to me needing a 32 bit version of libCoolProp.dylib as suggested in the debug steps of the manual. 

One of the reasons is that the command
**launchctl setenv DYLD_LIBRARY_PATH** to be pasted in the new .plist file doesn't work due to Apple's system integrity protection.

However, I was able to get it to work by copying the libCoolProp.dylib file to another location **/Users/yourusername/Library/Containers/com.microsoft.Excel/Data**. Credits: https://stackoverflow.com/questions/53772597/dylib-file-not-found-vba-excel-mac-2016

Note that when I do the following in my Terminal:
`$ file libCoolProp.dylib `
I get the following output:
`libCoolProp.dylib: Mach-O 64-bit dynamically linked shared library x86_64`
This means that I am using the correct 64 bit version of libCoolProp.dylib

Another thread that greatly helped me: https://github.com/CoolProp/CoolProp/issues/1430

jonCAstickel:
### Description of the Change

* Updated the Mac docs for enabling the Excel wrapper
* Using `DYLD_LIBRARY_PATH` has not worked for MacOS (OSX) since about 2015, and so that section has been removed. Trying to debug this stuff was a red herring for me when things don't work at first.
* Users with M1/M2 Macs are instructed to compile their own CoolProp library. (Alternatively, the x86_64 library may be used via Rosetta -- see #2139. It might be worth bringing some of those instructions to this PR. I can do so if requested.)

### Benefits

Clearer instructions for Mac users.

### Possible Drawbacks

None.

### Verification Process

N/A

### Applicable Issues

Closes #1430. Closes #2040. Closes #2224. Closes #1575.

### DIFF
--- a/Web/coolprop/wrappers/Excel/index.rst
+++ b/Web/coolprop/wrappers/Excel/index.rst
@@ -62,7 +62,7 @@ Pre-compiled Binaries for OSX
 
 .. warning:: 
 
-  There are now both 32-bit and 64-bit versions of Microsoft Excel on OSX.  You need to make sure that your bitness of the shared library for CoolProp (and perhaps REFPROP) match that of Excel.  
+  There are now both 32-bit and 64-bit versions of Microsoft Excel on OSX.  You need to make sure that your bitness of the shared library for CoolProp (and perhaps REFPROP) match that of Excel. If you are using a M1/M2 Mac with arm64 architecture, you may need to compile your own CoolProp library file (see `User-compiled Binaries`_).
 
 Part 1:
 -------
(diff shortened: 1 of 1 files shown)


ibell:
Thanks! Do you want to add the info about Rosetta as well? I personally usually use native shared libraries

PR #2374, opened 2024-05-06
Update REFPROPMixtureBackend.cpp

### LINKED ISSUE #2239: different result coolprop and refprop10
hi there 
I use refprop10 and coolprop 6.4.3

T=9.7778℃
P=2154.0 KPa A

I got result：
coolprop：
=PropsSI("H","T",9.778+273.15,"P",2154*1000,"REFPROP::R454B.MIX")
=222620.6 =222.6206 kJ/kg
refprop
=Enthalpy("R454B.MIX","TP","C",9.778,2154/1000)
=216.0817  kJ/kg
i also check other way to call refprop，for example C# ，I got 222.6206  kJ/kg
and  with refprop‘s user interface, i got 216.0817 

why?

another problem:
if i use the firest formula(coolprop) in excel，then the second formula(refprop) can also get 222.6206  kJ/kg
which one I can believe？

why

birdol:
Coolprop library does not implement the function call setrefdll from Refprop. As a result, there are discrepancies between the calculation results of Refprop and Coolprop when computing predefined mixture MIX models, primarily in terms of standard state deviations (which affect enthalpy values and related quantities).

Coolprop does provide a method to set the standard state, which can be set to DEF. However, the default value of DEF in Coolprop still differs from the calculation results of Refprop.

### Requirements

### Description of the Change

Based on the previously mentioned issue (https://github.com/CoolProp/CoolProp/issues/2239),
Coolprop library does not implement the function call setrefdll from Refprop. As a result, there are discrepancies between the calculation results of Refprop and Coolprop when computing predefined mixture MIX models, primarily in terms of standard state deviations (which affect enthalpy values and related quantities).
Coolprop does provide a method to set the standard state, which can be set to DEF. However, the default value of DEF in Coolprop still differs from the calculation results of Refprop.

### Benefits

[ The modified code now includes the call to setrefdll from refprop, resulting in the same calculations as those set in refprop when computing predefined mixtures. ]

### Possible Drawbacks

[The original user's calculations for predefined mixtures will be affected by the changes, but only when using the REFPROP backend. This will primarily impact variables such as enthalpy values. ]

### Verification Process

I only verified R410A.MIX  and did not conduct extensive testing, but I believe the impact should be minimal and feasible.

### Applicable Issues
Closes #2239

### DIFF
--- a/src/Backends/REFPROP/REFPROPMixtureBackend.cpp
+++ b/src/Backends/REFPROP/REFPROPMixtureBackend.cpp
@@ -343,6 +343,9 @@ void REFPROPMixtureBackend::set_REFPROP_fluids(const std::vector<std::string>& f
             strcpy(mix, _components_joined_raw);
 
             SETMIXdll(mix, hmx_bnc, reference_state, &N, component_string, &(x[0]), &ierr, herr, 255, 255, 3, 10000, 255);
+            int ixflag = 2;
+            double h0 = 0, s0 = 0, t0 = 0, p0 = 0;
+            SETREFdll(reference_state, &ixflag, &(x[0]), &h0, &s0, &t0, &p0, &ierr, herr, 3, 255);
             if (static_cast<int>(ierr) <= 0) {
                 this->Ncomp = N;
                 mole_fractions.resize(ncmax);


ibell:
It is not true that CoolProp does not implement the SETREFdll function of REFPROP. But it is true that the default reference state is not set to match that of REFPROP by default for predefined mixtures.

birdol:
> It is not true that CoolProp does not implement the SETREFdll function of REFPROP. But it is true that the default reference state is not set to match that of REFPROP by default for predefined mixtures.

Yes, you are right. I misspoke. Will this pull request be merged? Is there an issue with the code?

ibell:
Closing this PR; a more robust method is needed, probably driven by a config variable

PR #2402, opened 2024-08-07
Use temperature dependent hard sphere diameter for ion term in ePC-SAFT

### LINKED ISSUE #2338: Wrong values for temperature-dependent sigma of water in PC-SAFT
The parameters used for the sigma of water here is not the default one from Cameretti and Sadowski (they seem to have been obtained from a paper by Baird et al.): 

https://github.com/CoolProp/CoolProp/blob/ffa5a278ab5fbf19282b66cb7a183861578dfc11/src/Backends/PCSAFT/PCSAFTFluid.cpp#L58-L66

The actual values should be:
```c++
params.sigma = 2.7927 + 10.1100*exp(-0.01775*t)-1.41700*exp(-0.01146*t)
```

zmeri:
### Description of the Change

The ion term in the PC-SAFT code was updated to use the temperature dependent hard sphere diameter (d) instead of sigma. This is a fix for issue [2338](https://github.com/CoolProp/CoolProp/issues/2338).

### Benefits

Fixes the ePC-SAFT model so it give correct results.

### Possible Drawbacks

There's always the chance a mistake was introduced.

### Verification Process

I ran the Catch tests before and after making changes. The test results were exactly the same and all PC-SAFT tests passed.

### Applicable Issues

``Closes #2338``

### DIFF
--- a/src/Backends/PCSAFT/PCSAFTBackend.cpp
+++ b/src/Backends/PCSAFT/PCSAFTBackend.cpp
@@ -252,8 +252,7 @@ CoolPropDbl PCSAFTBackend::calc_alphar(void) {
     if (ion_term) {
         for (int i = 0; i < ncomp; i++) {
             if (components[i].getZ() != 0) {
-                d[i] =
-                  components[i].getSigma() * (1 - 0.12);  // for ions the diameter is assumed to be temperature independent (see Held et al. 2014)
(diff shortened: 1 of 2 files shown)


ibell:
Is this ready to merge?

zmeri:
Yes, this is ready to merge.

Issue #2394, opened 2024-06-30
Automatic Workflows Failing Since February/April


henningjp:
### Description

**Automated tests failing for many reasons recently; failing regularly with nightly builds...**
- Release workflow
   - Node.js 16 actions are deprecated.  Updated needed to Node.js 20.  These warnings will start throwing errors after June 30.
   - No artifacts created  
- Development Microsoft C++ Code Analysis
   - Same Node.js error as above
   - SARIF file not processed: mor runs than allowed (46 > 20 maximum); a recent change?

**...and testing of new commits/PRs**
- Development cppcheck 
   - These are all CATCH statements.  Has something changed with Catch2?
   
   ![image](https://github.com/CoolProp/CoolProp/assets/17114032/4eadbabf-5d92-4a71-9295-f8089fef64fb)

- Javascript library
   - em++: error: --memory-init-file is no longer supported.
   See: Emscripten-core/emscripten#21217
   As I think we depend on wasm2js, this could be a can of worms.
- Testing Catch2
   - Some of this is inability to find REFPROP and other libraries in the build environment
   > test cases:    80 |    53 passed |  27 failed
   > assertions: 47130 | 46824 passed | 306 failed

### Versions

**CoolProp Version:** 6.6.1dev

### Additional Information

This is way over my head to even start to tackle.  Is there some documentation on how this is all put together?

This looks like changes to the testing environment, not something that was pushed to the code.  Some of them may be related to the same issue.  The Emscripten switch and the Node.js updates look to be mandatory.  The Catch stuff is a more obscure.

ibell:
Regarding Catch2, did someone update it? They (Catch2) have made quite some changes, but they should not impact us unless somehow has updated the submodule

henningjp:
I don't think so.  looks like it's still pointing to the old version, but I don't know why it's breaking.

henningjp:
Looks like Catch2 is running now, but
- The web structure where Catch2 is running can't fine REFPROP or its FLUIDS/MIXTURES folders
- Some Ancillary checks not agreeing to specified tolerance (was there an update that wasn't reflected in the check values?)
- Humid Air checks not agreeing to specified tolerance (was there an updated that wasn't reflected in the check values?)

henningjp:
@ibell - Can I get to the automation environment to debug the REFPROP missing issue or is that an @jowr only thing?

ibell:
[...]

PR #2412, opened 2024-09-11
Update CMakeLists.txt to fix Mathcad wrapper check

henningjp:
### Description of the Change

Update CMake script to force definition of __cplusplus during build of Mathcad wrapper.  Might fix the code test for this wrapper.

### Benefits

Trying to get the automatic build check to pass.

### Possible Drawbacks

None.  Only impacts the COOLPROP_PRIME_MODULE in the CMake script.

### Verification Process

Works on local system and forces `__cplusplus` to be defined for Visual Studio build.  Triggers the following extern "C" wrapper around the code in the `mcadincl.h` header, which is a standard C-code header file (and library) and seems to be failing.

```cpp
#ifdef __cplusplus
extern "C"
{
#endif // __cplusplus

   [... C code header file...]

#ifdef __cplusplus
}
#endif // __cplusplus
```

It looks like maybe `__cplusplus` is not getting defined by default in the latest version of Visual Studio being used by the checks.  After configuring the project, the `/Zc:__cplusplus' compiler option is confirmed to be set.  Let's see if it fixes the code check.

### Applicable Issues

Might work to fix the failing Mathcad wrapper tests mentioned as failing in #2394 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -788,6 +788,9 @@ if(COOLPROP_PRIME_MODULE)
   list(APPEND APP_SOURCES
        "${CMAKE_CURRENT_SOURCE_DIR}/wrappers/MathCAD/CoolPropMathcad.cpp")
   add_library(CoolPropMathcadWrapper SHARED ${APP_SOURCES})
+  if(MSVC)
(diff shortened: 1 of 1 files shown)


henningjp:
Nope.  Thought sure that would do it.

PR #2432, opened 2024-10-29
Update setup.py replace distutils with setuptools PEP 632 – Deprecate distutils module

Housebi:
### Description of the Change

Replace all instances of the distutils module with the appropriate setuptools module, or other modules where applicable.

Edit line 4-5:
from distutils.version import LooseVersion
from distutils.sysconfig import get_config_var
To
from packaging.version import Version
from sysconfig import get_config_var

Edit line 291:
cython_version = Version (Cython.__version__)
To
cython_version = str(Version(Cython.__version__))

Edit line 353:
from distutils.errors import CompileError
To
from setuptools import CompileError

Edit line 394:
from Cython.Distutils.extension import Extension
To
from setuptools.extension import Extension

Delete line 396:
from Cython.Distutils import build_ext
This is now handled with line 6 "from setuptools.command.build_ext import build_ext", and thus redundant.

Edit line 3, 60, 61 62, 66, 77, 229, and 291
Replace all instances of LooseVersion with Version

### Benefits

To my understanding, as per PEP 632, the distutils module is being replace with setuptools and as of python 3.12 no longer included.

### Possible Drawbacks

unknown

### Verification Process

Tested by making a new wheel for python 3.13 on Windows 10 64 bit and installed via pip. Tested whether CoolProp could be imported and give simple fluid properties.

### Applicable Issues

Support for python 3.13

PR #2436, opened 2024-11-07
Python 3.13 and replace distutils with setuptools PEP 632 – Deprecate distutils

### LINKED ISSUE #2430: Support for python 3.13
After installing CoolProp with pip on python 3.13 it failed to import correctly.

While importing it raises a TypeError (expected bytes, str found). I will have to check on a different computer again if I can replicate the error as it is now solved by installing CoolProp with the source file and running python setup.py as described [here](https://stackoverflow.com/questions/74093424/how-to-install-coolprop-in-python3-9windows-and-pydroid3android).

From what I could find it could be that python 3.13 is not supported yet. This because the error looks similar to [#2310](https://github.com/CoolProp/CoolProp/issues/2310).

**CoolProp Version:** 6.6.0
**Access Method:** Python

### Additional Information

The error message after installing CoolProp with pip and running a random script that includes CoolProp.

Python 3.13.0 (tags/v3.13.0:60403a5, Oct  7 2024, 09:38:07) [MSC v.1941 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license()" for more information.

= RESTART: C:\Users\Niels Holswilder\source\repos\CoolpropTest\CoolpropTest.py =
  File "C:\Users\Niels Holswilder\source\repos\CoolpropTest\CoolpropTest.py", line 1, in <module>
  File "C:\Program Files\Python313\Lib\site-packages\CoolProp\__init__.py", line 22, in <module>
    __fluids__ = CoolProp.get_global_param_string('fluids_list').split(',')
  File "CoolProp.pyx", line 309, in CoolProp.CoolProp.get_global_param_string
  File "<stringsource>", line 15, in string.from_py.__pyx_convert_string_from_py_std__in_string
TypeError: expected bytes, str found

Housebi:
### Description of the Change

Updated setup.py, python_cibuildwheel.yml, and python_buildwheels.yml

setup.py
- Replaced distutils as per PEP 632 with setuptools and packaging

python_cibuildwheel.yml
- Added the module packaging to the dependencies
- Added 313 to the list of python versions
- Replaced python-version: 3.12.x with python-version: 3.13.x
- Removed some redundant lines of codes related to version specifically excluded architecture (for example ARM64 windows python 36) since these architectures are generically already excluded and/or the python version is no longer supported.

python_buildwheels.yml
- Replaced pypa/cibuildwheel@v2.17.0 with pypa/cibuildwheel@v2.21.3
- Replaced python-version: 3.12.x with python-version: 3.13.x

### Benefits

Python 3.13 should now be supported

### Possible Drawbacks

There is still an error left-over related to Testing Catch2 which was still unsolved in the previous pull request. #2432 

### Verification Process

Tested with pull request #2432 

### Applicable Issues

Closes #2430

### DIFF
--- a/.github/workflows/python_buildwheels.yml
+++ b/.github/workflows/python_buildwheels.yml
@@ -84,12 +84,6 @@ jobs:
         arch: [x86_64, arm64, universal2]
         exclude:
           - os: macos
-            arch: arm64
-            python-version: 36
-          - os: macos
-            arch: arm64
-            python-version: 37
-          - os: macos
             arch: universal2 # is redundant
     uses: ./.github/workflows/python_cibuildwheel.yml
     with:
--- a/.github/workflows/python_buildwheels.yml
+++ b/.github/workflows/python_buildwheels.yml
@@ -68,15 +68,6 @@ jobs:
         exclude:
           - os: windows
             arch: ARM64 # creates problems with msgpack-c
-          - os: windows
-            arch: ARM64
-            python-version: 36
-          - os: windows
-            arch: ARM64
-            python-version: 37
-          - os: windows
-            arch: ARM64
-            python-version: 38
     uses: ./.github/workflows/python_cibuildwheel.yml
     with:
       os: ${{ matrix.os }}
--- a/.github/workflows/python_buildwheels.yml
+++ b/.github/workflows/python_buildwheels.yml
@@ -23,7 +23,7 @@ jobs:
         python-version: 3.13.x
         
     - name: Install dependencies
-      run: pip install setuptools wheel Cython requests jinja2 pyyaml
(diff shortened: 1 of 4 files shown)


ibell:
Don't worry about the Catch2 tests, if the other actions run properly, I'll merge.

jsalort:
Is there a Python 3.13 wheel, or does one have to install from source?

Housebi:
You can install it by building it with the source files. But I will check my other computer at home, as it should have the wheel I made on it.

I am not sure whether the updated code is also uploaded to PyPI. This as it appears that installing via pip still results in an error.

ibell:
We need to fire a new release to get all the wheels to push to PYPI

Issue #2437, opened 2024-11-11
python numpy 2 


gitsdfawfevi:
### Description

the python wrapper "CoolProp" is only incompletely compatible with numpy > 2. specifically, I noticed this with the np.nan in CoolProp/Plots/Common.py ("Alias np.NaN has been removed. Use np.nan instead.", https://numpy.org/doc/stable/release/2.0.0-notes.html)

### Versions

**CoolProp Version:** 6.6.0
**Operating System and Version:** Archlinux 
**Access Method:** Python

ibell:
Do you want to take on fixing that? It would be appreciated by the
community.

gitsdfawfevi:
Yes, I will help and create a pull request specifically for the np.NaN. I'll have to see for myself how I can possibly provide more support.

jowr:
I guess this was closed in #2439

PR #2439, opened 2024-11-14
Fix np.NaN for numpy >=2

gitsdfawfevi:
### Description of the Change

Correct `np.NaN` to `np.nan` in `wrappers/Python/CoolProp/Plots/Common.py` (see also: https://github.com/CoolProp/CoolProp/issues/2437#issue-2649098703)

### Benefits

No error if using numpy >= 2 (https://numpy.org/doc/stable/release/2.0.0-notes.html)

### Possible Drawbacks

unknown

### Verification Process

Only tested in own project

### DIFF
--- a/wrappers/Python/CoolProp/Plots/Common.py
+++ b/wrappers/Python/CoolProp/Plots/Common.py
@@ -571,8 +571,8 @@ class IsoLine(Base2DObject):
             warnings.warn(
               "An error occurred for the critical inputs, skipping it.",
               UserWarning)
-            xcrit = np.NaN
-            ycrit = np.NaN
+            xcrit = np.nan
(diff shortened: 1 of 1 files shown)


ibell:
Can you please confirm this doesn't break code pre v2 of numpy?

gitsdfawfevi:
> Can you please confirm this doesn't break code pre v2 of numpy?

this should not lead to a problem in numpy, as it was already valid for version < 2: "NaN and NAN are equivalent definitions of nan. Please use nan instead of NaN." (https://numpy.org/doc/1.26/reference/constants.html#numpy.NaN)

ibell:
Okie dokie, merging

PR #2446, opened 2024-12-04
Fix Plots of log-p-h diagrams

MasterGPT9821:
When i tried to plot log(p)-h-diagrams there always had been an issue in the `common.py` file in line 481. By deleting the `Isoline, self` expression between the clips of super in this line, i could fix this and the plots were generated. So i made this pull request and hope you will fix the master that every coolprop user can make plots again.


wandadars:
Thank you for contributing this back to CoolProp. It's great to see community involvement in maintaining CoolProp.

ibell:
I guess this is fine. 3.9 is the latest supported version that is not end-of-life: https://devguide.python.org/versions/

PR #2463, opened 2025-01-19
Support plotting in C++

adolenc:
### Description of the Change

This pull request contains a relatively crude rewrite of Python's plotting into C++. The code only takes care of the calculations of the limits and isolines in SI units for different plot types (ph and Ts diagrams are extensively checked but others that are supported in Python have also been tried and seem to work). How the actual drawing is done is left to the end-user.

Usage example can be found in the unit tests at the bottom of `CoolPropPlot.cpp`.

Thank you for creating this really cool library. The code here is a cleaned-up version of what we are using in our fork at Dewesoft for our new Thermodynamics module, and it makes sense to open-source our modifications back into CoolProp as we think they might be generally useful.

### Benefits

It seems like at the moment the only way to really plot with CoolProp is by using the Python library, which is a reasonable choice, but makes it hard to do it from other languages or wherever Python is not supported. This change ads the support for plotting with CoolProp from C++ without other external dependencies, and it should be relatively easy to export the functionality over the C lib so it can potentially be used from any language wrapper.

### Possible Drawbacks

If this is merged there will now exist two competing implementations of "plotting" in CoolProp; one in Python and one in C++. A couple of considerations:
 - Python's implementation is strictly superior and provides a nicer experience for end-user, especially with the unit handling.
 - The C++ implementation _could_ be used as a relatively straight-forward replacement of (some) of the Python plotting's internal code.
 - I would argue that there is benefit to having two implementations, at least for the time being, as the Python one serves as a very nice reference if anything goes wrong.

Out of scope for this request are also:
 - cycle calculations,
 - performance/visualization optimization so that the ranges and scale of x/y axis are taken into account for isoline calculations,
 - C wrapper for this plotting code, and
 - the "guessing" part of the Python library is not implemented.

### Verification Process

The code contains unit tests that compare the output of the C++ code against the raw output of the Python code for ph and Ts diagrams.

### DIFF
--- a/include/CoolPropPlot.h
+++ b/include/CoolPropPlot.h
@@ -102,6 +102,18 @@ public:
     double value_at(CoolProp::parameters key, double xvalue, double yvalue, CoolProp::phases phase = CoolProp::phases::iphase_not_imposed) const;
 
 private:
+    struct Range2D
+    {
+        union
+        {
+            Range x, T;
+        };
+        union
+        {
+            Range y, p;
+        };
+    };
+
     std::string fluid_name;
     CoolProp::input_pairs axis_pair;
     bool swap_axis_inputs_for_update;
@@ -111,7 +123,7 @@ private:
 
     Range get_sat_bounds(CoolProp::parameters key);
     void get_Tp_limits(double& T_lo, double& T_hi, double& P_lo, double& P_hi);
-    std::vector<double> get_axis_limits(CoolProp::parameters xkey = CoolProp::parameters::iundefined_parameter, CoolProp::parameters ykey = CoolProp::parameters::iundefined_parameter, bool autoscale = true);
+    Range2D get_axis_limits(CoolProp::parameters xkey = CoolProp::parameters::iundefined_parameter, CoolProp::parameters ykey = CoolProp::parameters::iundefined_parameter, bool autoscale = true);
 };
 
 } /* namespace Plot */
--- a/include/CoolPropPlot.h
+++ b/include/CoolPropPlot.h
@@ -2,9 +2,7 @@
 #define COOLPROPPLOT_H_
 
 #include "AbstractState.h"
-#include "CPnumerics.h"
 #include <vector>
-#include <map>
 
 namespace CoolProp {
 namespace Plot {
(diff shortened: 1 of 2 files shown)


ibell:
Thanks you for your PR. It is nice to see a company getting value from open-source software like CoolProp giving back.

Overall this looks pretty good to me. It would be nice if you could add some doxygen-formatted comments, at least at the top of the header to give some context. I won't insist on documenting each function (and their arguments).

ibell:
The Catch tests fail, please look into that and fix as necessary

adolenc:
Fantastic, thank you for your time. I have added doctstrings to the public interface in `CoolPropPlot.h` file, please let me know if this is good enough.

I could use some help with the Catch tests though. On my branch, running `./CatchTestRunner [Plot]`, my 3 new tests seem to pass. If I remove the filter there seem to be quite a few failures, but they are also present on the master branch. And I wasn't able to figure out where to look on Github Actions - as far as I could tell the compiling itself failed before it reached my two files... 🙁

ibell:
Yes I can't really tell why the compilation failed on the GH actions.  Can you build locally on an ubuntu box?

adolenc:
> Yes I can't really tell why the compilation failed on the GH actions. Can you build locally on an ubuntu box?

Yes. I am using Ubuntu 22.04 in a Docker container. I have the following script that I adapted from actions to run the entire thing:
```sh
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE:STRING=Debug -DCOOLPROP_SHARED_LIBRARY:BOOL=ON -DCOOLPROP_CATCH_MODULE:BOOL=ON ..
cmake --build . --target install -j $(nproc) --config Debug
./CatchTestRunner [Plot]
```

To check the generated documentation I used the Doxygen Docker container:
```sh
docker run --rm -v $(pwd):/data -it hrektts/doxygen doxygen Doxyfile
```

ibell:
I think we can safely ignore the actions failures, as I don't think they are related to your code. I am having new weird build failures on my mac now as well.

ibell:
Do you consider your code ready to merge? You might want to consider to write some Cython wrappers of your code to make it again available in Python

adolenc:
Cheers guys, thanks. 

As far as I'm concerned it's good to go as is (you can probably squash the commits when merging). I think the current plotting available in Python is strictly superior, with a nicer API for the end user, so I don't see much value in exposing this code through Cython. I would eventually see value-add in C bindings for other wrappers, but we are only using it with C++, and I wouldn't mind helping out with that implementation after someone requests it first.

adolenc:
Woohoo, thanks! 🥳

PR #2478, opened 2025-02-23
Bump fmtlib to 11.1.3

ibell:
Required addition of unicode support

Squashes copious warnings in MSVC


ibell:
This seems safe to proceed

PR #2496, opened 2025-03-10
Add set_cubic_alpha_C function to python wrapper

jakobreichert:
### Description of the Change

The C++-function set_cubic_alpha_C seems to be missing from the python wrapper, so I added it.

### Benefits

The function set_cubic_alpha_C will be available in python, allowing the user to change the alpha function of cubic equations of state to Twu or MC and set the fluid specific parameters c1, c2 and c3.

### Possible Drawbacks

None.

### Verification Process

I compared the density calculation of PR with and without the Twu alpha function for a mixture of methane and ethane. The calculated densities differ, as expected, and match those that I get when using CoolProp with the Mathematica wrapper where the set_cubic_alpha_C function is already implemented.
I can do a more "formal" test and post the code if you wish, but since I only added an existing C++ function to the python wrapper I feel like this should be enough.

### Applicable Issues

No open issue that I am aware of.

### DIFF
--- a/wrappers/Python/CoolProp/AbstractState.pxd
+++ b/wrappers/Python/CoolProp/AbstractState.pxd
@@ -62,6 +62,7 @@ cdef class AbstractState:
     cpdef backend_name(self)
     cpdef fluid_names(self)
     cpdef fluid_param_string(self, string key)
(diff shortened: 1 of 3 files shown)


ibell (approved):
Looks good to me, thanks

PR #2511, opened 2025-04-02
Superancillaries for pure fluids

ibell:
### Description of the Change

Superancillaries were added for all pure fluids covered by Helmholtz EOS (aside from methanol). To disable them, do ``set_config_bool(ENABLE_SUPERANCILLARIES, false)``

### Benefits

QT and PQ inputs are at least tens of times faster, and more than 2000x faster for water (in C++). The speedups are around 100x in python

### Verification Process

- [x] Tests were added to check the entire temperature range
- [x] Figure out what to do about Tc and pc depending on whether SA are in use
- [x] Add superancillary validation figures to docs
- [x] Make the loading lazy for T(p) functions - they are slow in debug mode
- [x] Add jupyter notebook for the superancillary, and show how to use it
- [x] Validate other properties like latent heat w and w/o superanc
- [x] Test/fix Javascript


longemen3000:
may i ask if this change will require releasing a 0.7.0 version?

ibell:
> may i ask if this change will require releasing a 0.7.0 version?

I take it you mean 7.0 rather than 0.7.0? If so, I am not sure. From a SEMVER standpoint, there are not backwards-incompatible changes in the public API, so it is not necessary. But I may do so anyway to highlight that new functionality is now available.

longemen3000:
yeah, i was thinking that some users will find the difference between two versions a breaking one, but allegedly they shouldn't be testing exact results, specially when using superancillaries actually fixes a lot of usecases.

ibell:
I think I am coming down on the side of making this version 7.0 for the reasons you identified.

PR #2585, opened 2025-07-31
Update rapidjson to latest trunk to align with python-rapidjson

### LINKED ISSUE #2512: CoolProp fails to import when cadquery is imported first
When importing cadquery before CoolProp, the script silently hangs and does not proceed to execution. Reversing the import order resolves the issue. Has anyone encountered a similar issue? Should this be addressed in the cadquery repo instead? Thank you! 

### Minimal reproducible example
```
import cadquery
import CoolProp

if __name__ == "__main__":
    print("hello world")
```
### Environment
OS: Ubuntu 22.04
Python version: 3.10
cadquery version: https://github.com/CadQuery/cadquery/tree/2698db5c395705eb36e42bb5d4ab5944daf24436
CoolProp version: 6.6.0

eguiraud-pf:
This is to fix an incompatibility between python-rapidjson and coolprop's python package.

### Description of the Change

The version of rapidjson that we compile against is bumped to the latest master, so it matches what is used by the latest `python-rapidjson` release.

### Benefits

Using the same underlying version of the C++ rapidjson library between the various packages that come with extension modules that use it is necessary to avoid ODR violations and related weird issues resulting from having multiple different copies of the rapidjson symbols loaded in the same process.

### Possible Drawbacks

I have not tested CoolProp extensively with the new version of rapidjson. I hope CI covers this? :grimacing: 

### Verification Process

Before this PR:

```python
import rapidjson
import CoolProp
print("ok")
```

enters an infinite loop.

With this PR, it prints "ok" as expected.

### Applicable Issues

* Closes #2512


ibell:
Thanks very much for looking into this. I thought the only solution was to
replace with nlohmann::JSON, which will be necessary at some point, but not
an exciting prospect I have to say. I'll review, but see no reason not to
merge.

eguiraud-pf:
@ibell note that nlohmann::JSON would pose pretty much the same issues (it would still be a compiled-in dependency whose symbols would potentially clash with copies from other python modules). The best long-term solution is hiding the symbols as per the conversation in the issue.

PR #2624, opened 2025-10-02
Add Windows Instructions and example for Fluent Wrapper

OmarZaki96:
### Description of the Change

I added how to use the wrapper for Windows through the shared library. There might be a better way using the static library, but I believe this method works for 99% of use cases.

### Benefits

There was no guide for how to use CoolProp with Fluent in Windows anywhere

### Verification Process

Worked with both Fluent internal compiler and VS compiler

### DIFF
--- a/wrappers/Fluent/Example_windows_UDF.cpp
+++ b/wrappers/Fluent/Example_windows_UDF.cpp
@@ -1,0 +1,50 @@
+#include "udf.h"
+#include "CoolPropLib.h"
+#include <string>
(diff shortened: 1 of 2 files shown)


OmarZaki96:
I tried using the Static library approach, but linking the .Lib file to the UDF compilation resulted in errors. I currently don't have time to debug the issue, and I still believe that using the shared library approach is easier for most users who are not familiar with the C++ compilation process (including me).

PR #2626, opened 2025-10-04
Add Ideal gas methods

### LINKED ISSUE #2589: Properties of Air as an Ideal Gas
### Discussed in https://github.com/CoolProp/CoolProp/discussions/2569

<div type='discussions-op-text'>

<sup>Originally posted by **augustoperes1** July  3, 2025</sup>
Basically I would like to redo this Moran & Shapiro table of the properties of Air as an Ideal Gas, only using CoolProp in LibreOffice Calc. By default I need to give both states for CoolProp to calculate the properties, but in the ideal gas model enthalpy and internal energy depend only on temperature.
image
</div>

ibell:
### Description of the Change

Adds the h, s, u ideal-gas outputs on mass and molar bases

### Applicable Issues

Closes #2589 
Closes #2569


include/DataStructures.h
longemen3000, line 1: Probably i'm the only user for this, but, adding enums in that way will break some code that depends on the output of `get_param_index`. I did a preliminary search and almost everybody uses the the `get_param_index` just as an input for AbstractState update functions, but it could cause problems if CoolProp is being accessed as a dll.

Issue #2694, opened 2026-03-20
Download of REFPROP fails in CI build due to bad session key


tobias-loew:
In #2692 the CI-build fails for "run address sanitizer" and "Testing Catch2" due to a failure when trying to decrypt gpgRefprop (cf. https://github.com/CoolProp/CoolProp/actions/runs/23350655759/job/67928118709?pr=2692)
The PR does not touch anything near that code.

PR #2710, opened 2026-04-04
Skip REFPROP tests when not available during CI

henningjp:
### Description of the Change

REFPROP Tests still failing during CI from PRs on other branches/forks (see my REFPROP and ASAN failures on #2709).  Apparently #2699 didn't quite do it.  Here's an alternative.

**Solution** approach suggested by GitHub Copilot is to conditionally skip REFPROP-dependent tests when REFPROP is not available. This makes tests robust across different environments even when run outside of GitHub workflows.

**Code Fix** adds a helper function in `CoolProp::` to check REFPROP availability and skip affected tests (_there are ten_) in `CoolProp.cpp`, `REFPROPMixtureBackend.cpp`, and `CoolProp-TESTS.cpp` if REFPROP is unavailable.  If in a non-test environment, the helper function is still available, but is a NoOp.

### Benefits

1. Non-breaking: Tests pass in environments with REFPROP installed
2. Robust: Tests are skipped gracefully when REFPROP is unavailable (common in CI)
3. Maintains PR intent: The PR's code changes are unaffected and other non-REFPROP tests still run.
4. Reduces flakiness: No more failures due to missing dependencies
5. Doesn't rely on the GitHub runners; Users can compile and run the Catch Tests locally, even if they don't have REFPROP, without failures.

### Verification Process

- [x] Verify a fork PR no longer fails on the "Build REFPROP" step (based on this PR).
- [x] Verify a fork PR skips all tests (ASAN & CATCH) that require REFPROP (based on this PR)
- [x] Verify an internal branch PR still runs REFPROP tests (based on recent PRs)
- [x] Verify push to master still runs REFPROP tests (based on recent pushes)
- [ ] Verify Non-Catch compiles are unaffected (Tested Mathcad Prime Wrapper - others tested in CI)

### Applicable Issues

#2694

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@
 /include/pcsaft_fluids_schema_JSON.h
 /include/version.h
 /include/catch.hpp
+/install_root/
 /build*/
 /dev/hashes.json
 /include/cpversion.h
--- a/include/CoolProp.h
+++ b/include/CoolProp.h
@@ -191,5 +191,8 @@ std::string extract_fractions(const std::string& fluid_string, std::vector<doubl
 /// @param Phase The enumerated phase index to be looked up
 std::string phase_lookup_string(phases Phase);
 
+/// A function to skip tests that require REFPROPMixtureBackend in environments where it is not supported (e.g., CI builds on GitHub)
+void Skip_if_No_REFPROP();
+
 } /* namespace CoolProp */
 #endif
--- a/src/AbstractState.cpp
+++ b/src/AbstractState.cpp
@@ -1044,8 +1044,17 @@ TEST_CASE("Check AbstractState", "[AbstractState]") {
     SECTION("good backend - incomp") {
         CHECK_NOTHROW(shared_ptr<CoolProp::AbstractState>(CoolProp::AbstractState::factory("INCOMP", "DEB")));
     }
+    // SECTION("good backend - REFPROP") {
+    //     CHECK_NOTHROW(shared_ptr<CoolProp::AbstractState>(CoolProp::AbstractState::factory("REFPROP", "Water")));
+    // }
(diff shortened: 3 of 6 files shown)


henningjp:
@ibell - This might be a first for one of my remote fork PRs!  Very exciting.

image

And when REFPROP is available (locally on my machine), CatchTestRunner.exe give:

image

Ready when you are.   It relies on the existing function CoolProp::REFPROPMixtureBackend::REFPROP_supported() to determine if it should run each Catch test that involves REFPROP (10 of them) or SKIP it.

ibell (approved):
Seems fine; my only comment might be that you could make a simpler re-usable function that is used in all the places rather than some copy-pasting

henningjp:
Not sure I understand.  REFPROP_supported() is the re-usable function and it already existed in RefpropMixtureBackend.  Is there a simpler way than checking to see if it's not true and skipping the test?  The only one requiring a little extra effort is the backend check, because it hasn't been initiated yet.

henningjp:
We could pull the SKIP into the re-usable function, but Copilot suggests keeping the SKIP calls local, otherwise it drags the test framework macros into library headers.  It offered a solution very similar to what's already being done, but it redundantly wrapped CoolProp::REFPROPMixtureBackend::REFPROP_supported() and called it to determine the local SKIP in each test case.  Not sure if this is what you meant.

henningjp:
@ibell - Mulled this one over.  I added a function `Skip_if_No_REFPROP()` to `CoolProp::` namespace.  If in CATCH mode, this function issues the CATCH2 SKIP when REFPROP not available.  In non-CATCH mode, this function is simply a NoOp.  This turns the REFPROP checks into a single, simple void function call.

Is this what you had in mind?

ibell (approved):
LGTM; that's what I had in mind. Much more DRY

PR #2728, opened 2026-04-14
build: replace git submodules with CPM.cmake

ibell:
## Summary

- Replaces all 11 git submodules with [CPM.cmake](https://github.com/cpm-cmake/CPM.cmake) (v0.40.7), a thin wrapper around CMake's `FetchContent` with a shared download cache
- Set `CPM_SOURCE_CACHE=~/.cache/CPM` in your environment to share downloads across git worktrees and build directories — no more `git submodule update --init --recursive` per worktree
- Vendored deps with no upstream release cycle (miniz, nlohmann-json, incbin) remain in `externals/` unchanged

**Packages migrated:** Eigen 3.4.0, fmtlib 12.0.0, msgpack-c, rapidjson, IF97, REFPROP-headers, multicomplex, Catch2 v3.8.0, pybind11 v2.11.0, ExcelAddinInstaller, FindMathematica

**Source changes:** `rapidjson_include.h`, `IF97Backend.h`, `HumidAirProp.cpp`, and `REFPROPMixtureBackend.cpp` updated to use angle-bracket includes now that dep directories are on the include path via CPM rather than relative paths from the repo root.

## Test plan

- [x] Clean CMake configure succeeds (`-DCOOLPROP_CATCH_MODULE=ON`)
- [x] `CatchTestRunner` builds cleanly
- [x] 42/42 non-REFPROP assertions pass (2 REFPROP failures are expected without the library installed)
- [x] Verify Windows build (ExcelAddinInstaller path)
- [x] Verify Python pybind11 wrapper build
- [x] Verify Mathematica wrapper build (FindMathematica path)
- [x] Verify Mathcad Prime wrapper build.


ibell:
@jowr @henningjp , I plan to make this change to make it easier to use git worktrees with AI coding assistants. worktrees are essential tools in modern AI workflows, and git submodules don't play nicely with git worktrees

henningjp:
I can check the Windows installer this weekend.
My Mathematica license is lapsed, I'll see if I can get that reinstated and try that one as well.

CMakeLists.txt
henningjp, line 348: While we're at it.  This type should be STRING, not LIST.  It's been throwing a warning in CMake for years (at least on Windows).  It actually interprets LIST as STRING and throws the warning.  It works, but...
Same thing 3 lines down.  Is this type LIST for a reason?

henningjp:
Windows installer is failing, but it may be something I'm doing.  This first message that comes up is concerning:

>  CMake Warning at cmake/CPM.cmake:80 (message):
>    CPM: Your project is using an unstable development version of CPM.cmake.
>    Please update to a recent release if possible.  See
>    https://github.com/cpm-cmake/CPM.cmake for details.

After that MS Build fails all over the place.  Still working on that.

.github/workflows/windows_installer.yml
henningjp: Get following:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_BUILD_TYPE
```

Probably always happened, but nobody builds this manually to see it.

henningjp:
In the Windows installer, I'm getting through
- 32BitDLL_cdecl
- 32bitDLL_stdcall
- 64bitDLL

but it looks like the arm64bitDLL version is failing, even though I think I have all the arm compiler components installed.  It's also trying to use the Debug configuration even though I have --config Release on the command line.    The good news is that the CPM system seems to be working and generating components.  The arm compiler is my issue.  I'm sure the virtual machine that the workflow runs on will handle this once it's pushed.

ibell:
Thanks @henningjp - I think we are good to go. This will be a nice upgrade to the workflow

henningjp:
@ibell - Progress.  All DLLs are built.  I'm just not getting the installer build.  I suspect I need the Inno Setup module, which I don't have installed.  Good to go.

henningjp:
Just a few review items above to address.  The FindMathematica one is critical.  Won't build without this.

PR #2741, opened 2026-04-19
fix(cmake): Update FindMathematica dependency to latest version

### LINKED ISSUE #2740: [ISSUE] Mathematica won't build under new CMake Package Manager (CPM)
### Description

Can't create the Wolfram Mathematica build directory with CMake.  

Version of [FindMathematica](https://github.com/sakra/FindMathematica) being linked into Mathematica is (_often_) out of date.

### Steps to Reproduce

1. Make sure you have a version of Mathematica installed (even an old unlicensed one)
2. In the local CoolProp repository, create a directory starting with "build",  `mkdir buildMMA`
3. Change to that directory `cd buildMMA`
4. Generate the build solution in this directory for either VS or MingW (VS shown here)
```
cmake .. -DCOOLPROP_MATHEMATICA_MODULE=ON -DCMAKE_VERBOSE_MAKEFILE=ON -G "Visual Studio 17 2022" -A x64
```

**Expected behavior:** 

> Build system created successfully.

**Actual behavior:** 

> **Error:** "CMake no longer supports version < 3.1 ..."  
> Imported [FindMathematica](https://github.com/sakra/FindMathematica) package is bringing in a non-current version that is trying to use:
> - An unsupported CMake version
> - Also, can't find the most recent version of Wolfram Mathematica, which is installed on the local machine.

### Versions

**CoolProp Version:** 7.2.1-dev
**Operating System and Version:** Windows 
**Access Method:** n/a  

### Additional Information

This was a recurring issue, even before CPM was implemented in #2728, but now it is **_much_** easier to fix and keep current.

henningjp:
### Description of the Change

Keep [FindMathematica](https://github.com/sakra/FindMathematica) package up to date.
- Change package `GIT_TAG` from a specific SHA to `master` so that it will always pull the most recent version when generating the build system with the CPM.
- Update Mathematica wrapper page (`reST` file) for accuracy (binaries no longer available on SourceForge) and updated build steps for compiler versions in this decade.

### Benefits

Users won't have to constantly keep updating the CPM dependencies file to keep this package up to date for:
1. Minimum CMake version supported
2. Supported versions of Wolfram Mathematica family of products

### Possible Drawbacks

While there is a possibility of automatically incorporating a regression from [FindMathematica](https://github.com/sakra/FindMathematica), it is extremely improbable that it will have any impact on CoolProp for several reasons:
1. It is only needed by users who want to build the Mathematica wrapper and does not impact CoolProp code.
2. With the new CPM ( #2728 ), [FindMathematica](https://github.com/sakra/FindMathematica) is no longer included as a submodule in the CoolProp repository.

Additionally, all of the commits to [FindMathematica](https://github.com/sakra/FindMathematica) over the last decade have consisted of:
- Updates to the lowest supported version of CMake
- Updates to the latest Mathematica release version
- Updates to Wolfram product name changes and variations for Mathematica, Desktop Mathematica, Wolfram Engine, WolframNB, etc.
- Updates to logic for finding the actual Wolfram products in MaxOS, OSX, etc., which can be fairly dynamic.
- Updates to the manual and README files.

All of these are rather innocuous to CoolProp and worthy of being kept dynamically up to date without any maintenance from the CoolProp community.

### Verification Process

- Verified that the CPM dependency specification `GIT_TAG    master` actually pulls the latest [FindMathematica](https://github.com/sakra/FindMathematica) commit and successfully generates the build environment with CMake.
- Verified that the Mathematica wrapper is built, installs per the instructions on the updated Mathematica web page, and successfully runs in Mathematica 14.3 (latest).

### Applicable Issues

Closes #2740

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -94,7 +94,7 @@ if(COOLPROP_MATHEMATICA_MODULE)
   CPMAddPackage(
     NAME FindMathematica
     GIT_REPOSITORY https://github.com/sakra/FindMathematica
-    GIT_TAG        99d8c95eed1a517f4a06511ea40b2cee6477a8c1
+    GIT_TAG        4.2.0
     DOWNLOAD_ONLY  YES
   )
 endif()
--- a/Web/coolprop/wrappers/Mathematica/index.rst
+++ b/Web/coolprop/wrappers/Mathematica/index.rst
@@ -6,15 +6,9 @@ Mathematica Wrapper
 
 Pre-compiled Binaries
 =====================
-Pre-compiled binaries can be downloaded from :sfdownloads:`Mathematica`, which come from :sfnightly:`the nightly snapshots <Mathematica>`.
-
-Place the shared library for your platform (windows: CoolProp.dll, OSX: CoolProp.dylib, linux: CoolProp.so) in the directory that you can obtain from Mathematica::
-
-    FileNameJoin[{$BaseDirectory, "SystemFiles", "LibraryResources", $SystemID}]
-
-If this directory doesn't yet exist, create it. At the command prompt, you should be able to call ``FindLibrary["CoolProp"]``
-
-There is a small example file ``example.nb`` that demonstrates how to call the functions in the DLL which can be downloaded from :sfdownloads:`Mathematica`
+Pre-compiled binaries are no longer uploaded to the  :sfdownloads:`latest Release` on SourceForge or :sfnightly:`the nightly snapshots <Mathematica>`.  This is because:
+- Usage is one of the lowest of all the wrappers supported by CoolProp with few active developers
+- The wrapper should really be built by the user for their specific version of Wolfram Mathematica on their system.  Major and minor updates to Mathematica are just too frequent to keep up with. 
 
 User-Compiled Binaries
 ======================
@@ -22,6 +16,8 @@ User-Compiled Binaries
 Common Requirements
 -------------------
 Compilation of the Mathematica wrapper requires a few :ref:`common wrapper pre-requisites <wrapper_common_prereqs>`
+
+A version of Wolfram Mathematica (v13.0 or higher) or Wolfram Engine
 
 Linux and OSX
 ^^^^^^^^^^^^^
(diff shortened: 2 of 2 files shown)


cmake/dependencies.cmake
```diff
@@ -94,7 +94,7 @@ if(COOLPROP_MATHEMATICA_MODULE)
   CPMAddPackage(
     NAME FindMathematica
     GIT_REPOSITORY https://github.com/sakra/FindMathematica
-    GIT_TAG        99d8c95eed1a517f4a06511ea40b2cee6477a8c1
+    GIT_TAG        master
```
ibell: I have mixed feelings about this. I think my preference would be to keep it pinned, as it is easy enough to update the pin

ibell (changes requested):

cmake/dependencies.cmake
henningjp: @ibell  - Neither the core code nor any other wrappers depend on this package.  I maintain that it can't hurt.  All this tool does is go looking for the user's latest locally installed version of Mathematica when the wrapper is built and pass back the path to it.  Mathematica is not part of the CI actions and this package will never be used by the workflows.  

"_it is easy enough to update the pin_": For us maybe.  Probably not for the typical Mathematica user figuring out where it's buried in the CPM dependencies.

I can't see how it could cause anything to fail, _except for the random Mathematica wrapper build by a user_, and that happens now anyway because the submodule/package is usually out of date for the rapidly changing versions of Mathematica.

If you approve it, I'll actively keep an eye on everything and revert back to a pinned SHA if it creates any problems.

ibell: I'm not so sure. There are a growing number of software supply chain attacks, and I would prefer to keep the pin in place to avoid us becoming inadvertently a vector for that.

henningjp: OK.  I'll bump it to the latest FindMathematica version tag.

ibell: Thanks

henningjp: I learn something new every day here!  😃

Issue #2742, opened 2026-04-19
Update cubic EOS coefficients


ibell:
The rounded coefficients are used in consistency with the literature, but we should move back to the unrounded coefficients. The new coefficients are indicated in this paper: https://tsapps.nist.gov/publication/get_pdf.cfm?pub_id=931868

PR #2746, opened 2026-04-20
Wilson K values: use exact coefficient

longemen3000:
somewhat related to #2742. The Wilson K-value correlation is actually a linear interpolation between ($0.7T_c$, $P(0.7T_c$) and ($T_c$, $P_c$) in ($T^{-1}$, $\log{p}$) coordinates. with that in mind, the more precise correlation is:

$$
\log{K} = \log{\frac{P_c}{P}}+c_1 \left(1 + \omega\right)\left(1-\frac{T_c}{T}\right)
$$

$$
c_1 = \frac{7\log{10}}{3} \approx 5.3726985503194395
$$

### DIFF
--- a/src/Backends/Helmholtz/VLERoutines.h
+++ b/src/Backends/Helmholtz/VLERoutines.h
@@ -64,7 +64,7 @@ static CoolPropDbl Wilson_lnK_factor(const HelmholtzEOSMixtureBackend& HEOS, Coo
     const double pci = HEOS.get_fluid_constant(i, iP_critical);
     const double Tci = HEOS.get_fluid_constant(i, iT_critical);
     const double omegai = HEOS.get_fluid_constant(i, iacentric_factor);
-    return log(pci / p) + 5.373 * (1 + omegai) * (1 - Tci / T);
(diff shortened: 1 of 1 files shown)


ibell:
Can you add a comment to this effect in the source code? 

Nit: one should never write a log without its base in a mathematical formula to ensure that the intent is clear

ibell:
Ping @longemen3000 . Still interested in this one?

longemen3000:
Sorry, I did not see the notification 😅, I will add some descriptions in the files, but if the LLM does it, that would also be good

longemen3000:
If you are interested in a better correlation for the initial point of the VLE 2-phase PT flash, the algorithm that I use is a combination of:
- Wilson K-values if they result in a valid K
- the thermopack initial point (a pseudo TPD approach, they have it their flash docs) 
- if the point is above the critical point, we can extrapolate at the critical point to get a pseudo-saturation line.

ibell:
See also #3026 & #2720 - exciting things are coming!

ibell:
If you wouldn't mind updating the PR I would appreciate it

PR #2800, opened 2026-04-26
fix: thread-safe lazy initialization of fluid libraries

### LINKED ISSUE #2787: Fluid-library static init is not thread-safe (race during cold cache population)
## Summary

CoolProp's fluid-library static initialization is not thread-safe. When multiple threads make their first fluid-related call concurrently — `Props1SI`, `AbstractState_factory`, `get_global_param_string("FluidsList")`, etc. — they race during the JSON parse / index-map population, leaving the library partially populated. The resulting `FluidsList` contains duplicate single-fluid entries instead of the full ~120-fluid list, and lookups for any other fluid (`Water`, `R134a`, …) fail with `key [Water] was not found in string_to_index_map in JSONFluidLibrary`.

## Reproduction

Discovered while developing the new desktop GUI on branch `ib/GUI`. The Rust test binary in `wrappers/GUI/src-tauri/` runs unit tests in parallel by default; with that, CoolProp reliably misbehaves:

```
$ cargo test --release --no-fail-fast 2>&1 | grep -E "FluidsList missing|finished in"
FluidsList missing Water: "D4,D4,D4,D4,R125"
test result: FAILED. 5 passed; 12 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.62s
```

The repeated single-fluid name varies between runs (`IsoButene` × 4, `D4` × 4 + `R125`, etc.) — the signature of multiple threads writing into the same hash slot.

Forcing serial test execution makes everything green:

```
$ cargo test --release --no-fail-fast -- --test-threads=1
test result: ok. 17 passed; 0 failed
```

Reproduced on:
- macOS 14 (Apple Silicon)
- Ubuntu 22.04 (GitHub-hosted runner)

## Proposed fix

Wrap the fluid-library init (likely the singleton accessor in `JSONFluidLibrary` / `Backends/Helmholtz/Fluids/FluidLibrary.h`) in a `std::call_once` (or equivalent), or make the underlying `string_to_index_map` access guarded by a mutex during the first population.

Single-threaded callers and most existing test suites don't hit this because once the library is fully populated, subsequent reads are safe. The bug only surfaces when multiple threads each hit a cold cache simultaneously.

## Workaround

Callers can serialize the first lookup themselves (e.g. issue a single `Props1SI("Water","Tcrit")` from the main thread before spawning workers). The GUI's CI test job has been pinned to `--test-threads=1` for now; tracked in CoolProp/CoolProp `wrappers/GUI/.github/workflows/gui_builder.yml`.

ibell:
Closes #2787.

## Summary

Five global singletons in CoolProp lazy-load JSON-encoded data on first access without synchronization. Multiple threads concurrently hitting a cold cache all see the singleton empty, all run the `load()` path, and the resulting concurrent inserts into the underlying `std::map` / `std::vector` race.

The visible symptom is a partially populated library — `FluidsList` comes back as `"D4,D4,D4,D4,R125"` or `"IsoButene,IsoButene,IsoButene,IsoButene"`, and any subsequent `get("Water")` (etc.) fails with `key [Water] was not found in string_to_index_map in JSONFluidLibrary`. Triggers reliably from cargo's default-parallel test runner in `wrappers/GUI/src-tauri/` on macOS (Apple Silicon) and ubuntu-22.04; full repro in #2787.

## Fix

Wraps each lazy-init path in `std::call_once` with a local `std::once_flag`:

- `src/Backends/Helmholtz/Fluids/FluidLibrary.cpp` — file-scope flag covering `get_library`, `get_fluid`, `get_fluid_as_JSONstring`, `get_fluid_list`, `set_fluid_enthalpy_entropy_offset`, and the public `add_many(string)` overload. Six `if (library.is_empty()) load();` sites collapsed into a single `ensure_library_loaded()` helper.
- `src/Backends/Incompressible/IncompressibleLibrary.cpp` — same shape, four accessors.
- `src/Backends/Helmholtz/MixtureParameters.cpp` — `MixtureBinaryPairLibrary` and `MixtureDepartureFunctionsLibrary` each get a per-instance `std::once_flag` member that their existing `load_defaults_if_needed()` now routes through.
- `src/HumidAirProp.cpp` — `check_fluid_instantiation()` now runs the `Water` / `WaterIF97` / `Air` shared_ptr resets inside `std::call_once`. Without it two threads observed null, both reset the shared_ptr, and the late assignment destroyed the in-flight backend the other thread was using.

`std::call_once` retries on exception, so the previous retry-on-failed-load behavior is preserved (a parse error in `load()` leaves the flag unset and the next caller tries again).

## Verification

Cargo's default-parallel test runner against the new GUI Rust binary in `wrappers/GUI/src-tauri/` went from **5 passed; 12 failed** to **17 passed; 0 failed**. The `--test-threads=1` workaround currently in `.github/workflows/gui_builder.yml` (not on master yet) becomes unnecessary once this lands.

## Out of scope (follow-ups)

The audit found two more thread-safety issues that warrant separate PRs:

- Static-init-time races in `PCSAFTLibrary` / `CubicsLibrary` constructors — populate the map in the constructor, technically safe in single-threaded static-init but fragile in dynamic-library contexts.
- Concurrent `TabularDataLibrary::data` map insertion in `src/Backends/Tabular/TabularBackends.cpp` — different pattern (map insert race rather than singleton-init race), needs a per-key mutex.

## Test plan

- [x] Local: `cargo test --release` (default parallel) on macOS arm64 — 17 passed.
- [ ] CI: Catch2 / library_shared / Python cibuildwheel suites stay green.

### DIFF
--- a/src/Backends/PCSAFT/PCSAFTLibrary.h
+++ b/src/Backends/PCSAFT/PCSAFTLibrary.h
@@ -28,18 +28,8 @@ class PCSAFTLibraryClass
     void load_from_JSON(rapidjson::Document& doc);
     void load_from_string(const std::string_view& str);
 
-    /// Schema-validate then parse a JSON blob and merge into this instance.
-    /// Called from the constructor to avoid recursing through the global
-    /// singleton during its own construction.
-    void _add_fluids_from_JSON_string(const std::string_view& JSON);
-
    public:
     PCSAFTLibraryClass();
-
-    /// Public wrapper used by the free `add_fluids_as_JSON()` after construction.
-    void add_fluids_from_JSON_string(const std::string_view& JSON) {
-        _add_fluids_from_JSON_string(JSON);
-    }
 
     bool is_empty() {
         return empty;
--- a/src/Backends/PCSAFT/PCSAFTLibrary.h
+++ b/src/Backends/PCSAFT/PCSAFTLibrary.h
@@ -28,8 +28,18 @@ class PCSAFTLibraryClass
     void load_from_JSON(rapidjson::Document& doc);
     void load_from_string(const std::string_view& str);
 
+    /// Schema-validate then parse a JSON blob and merge into this instance.
+    /// Called from the constructor to avoid recursing through the global
+    /// singleton during its own construction.
+    void _add_fluids_from_JSON_string(const std::string_view& JSON);
+
    public:
     PCSAFTLibraryClass();
+
+    /// Public wrapper used by the free `add_fluids_as_JSON()` after construction.
+    void add_fluids_from_JSON_string(const std::string_view& JSON) {
+        _add_fluids_from_JSON_string(JSON);
+    }
 
     bool is_empty() {
         return empty;
--- a/src/Backends/Cubics/CubicsLibrary.cpp
+++ b/src/Backends/Cubics/CubicsLibrary.cpp
@@ -20,14 +20,45 @@ class CubicsLibraryClass
     /// Map from index of fluid to a string
     std::map<std::string, std::string> JSONstring_map;
     bool empty;  // Is empty
+
+    /// Schema-validate then parse a JSON blob and merge into this instance.
+    /// Used by the constructor (avoids recursing through the global singleton)
+    /// and by the free `add_fluids_as_JSON()` (operates on get_library()).
+    void _add_fluids_from_JSON_string(const std::string_view& JSON) {
+        std::string errstr;
+        cpjson::schema_validation_code val_code = cpjson::validate_schema(cubic_fluids_schema_JSON, JSON, errstr);
+        if (val_code == cpjson::SCHEMA_VALIDATION_OK) {
+            rapidjson::Document dd;
+            dd.Parse<0>(JSON.data(), JSON.size());
+            if (dd.HasParseError()) {
+                throw ValueError("Cubics JSON is not valid JSON");
+            } else {
+                try {
+                    this->add_many(dd);
+                } catch (std::exception& /* e */) {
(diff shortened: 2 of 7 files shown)


src/Backends/PCSAFT/PCSAFTLibrary.h
```diff
@@ -28,9 +28,19 @@ class PCSAFTLibraryClass
     void load_from_JSON(rapidjson::Document& doc);
     void load_from_string(const std::string_view& str);
 
+    /// Schema-validate then parse a JSON blob and merge into this instance.
+    /// Called from the constructor to avoid recursing through the global
+    /// singleton during its own construction.
+    void _add_fluids_from_JSON_string(const std::string_view& JSON);
```
ibell: should this be in the public header? seems like it would be better to keep this all in the cpp file

ibell: Done in 192510e9 — moved both PCSAFT and Cubics helpers to file-private free functions in an anonymous namespace inside the .cpp; PCSAFTLibrary.h reverts to its pre-PR shape (no new declarations).

Re: thread safety — for Cubics the protection is at `static CubicsLibraryClass library;` inside `get_library()` (CubicsLibrary.cpp:175-178), the Meyers singleton pattern. C++11 [stmt.dcl]/4 guarantees that function-local static initialization runs exactly once even under concurrent first calls, so it's equivalent in semantics to the `std::once_flag` + `std::call_once` used in FluidLibrary.cpp. Same applies to PCSAFT in PCSAFTLibrary.cpp's `get_library()`.

Happy to switch to explicit `std::once_flag` + `std::call_once` if you'd prefer the pattern to match FluidLibrary uniformly — let me know.

PR #2831, opened 2026-04-28
fix(HumidAirProp): per-thread Water/Air backends via thread_local

ibell:
## Summary

- PR #2800 fixed *construction* of Water/Air/WaterIF97 (call_once), but every HAPropsSI call mutates those shared singletons (`Water->update(...)`, `Air->specify_phase(...)`, `Water->clear()`). Two threads concurrently calling HAPropsSI corrupt each other's backend state, producing NaN/_HUGE or wildly wrong numerical results.
- Convert `Water`/`Air`/`WaterIF97` and the dependent file-static caches (`s_virial_cache`, `s_alpha0_cache`, `s_ref`) to `thread_local`. Each thread constructs its own backends on first use; the global `FluidLibrary` reached through those constructors is already call_once-protected.
- Convert the three `Use*Correlations` int flags to `std::atomic<int>` since they're intentionally process-global toggles, not per-thread state.
- Add `[HAPropsSI][threadsafe]` Catch test (16 threads × 50 iters). Verified: without the fix, 610/800 calls return NaN/_HUGE and 75 more return wrong values (~85% broken); with the fix, 2/2 assertions pass and all 57,331 assertions in the full Catch suite still pass.

## Follow-ups (filed as separate issues, not in this PR)

The broader audit turned up three more independent thread-safety issues outside HumidAirProp:

- `src/CoolProp.cpp` `debug_level` / `error_string` / `warning_string` — unsynchronized globals; `get_global_param_string("errstring")` is a non-atomic read-then-clear that loses errors.
- `src/Configuration.cpp` `pconfig` lazy-init missed by PR #2800 — needs `call_once` or Meyers.
- `src/Backends/Tabular/TabularBackends.cpp` `library` map populated lazily without lock; REFPROP wrapper has multiple unsynchronized globals (REFPROP itself is single-threaded).

These will land as separate stacked PRs to keep review focused.

## Test plan

- [x] `./CatchTestRunner '[HAPropsSI][threadsafe]'` passes
- [x] `./CatchTestRunner '[HAPropsSI],[humid_air],[HAProps],[RP1485]'` passes (213 assertions, 11 cases)
- [x] `./CatchTestRunner` full suite passes (57,331 assertions, 126 cases, 10 skipped)
- [x] Verified test catches the regression: temporarily reverted just `thread_local` on Water/Air/WaterIF97 → test fails with 610 bad numbers and 75 mismatches
- [ ] CI macOS / Ubuntu / Windows green
- [ ] Optional: re-enable `--test-threads=auto` in `wrappers/GUI/src-tauri` Rust tests on a follow-up to verify the GUI race is gone end-to-end

### DIFF
--- a/src/HumidAirProp.cpp
+++ b/src/HumidAirProp.cpp
@@ -2542,6 +2542,33 @@ TEST_CASE("HAPropsSI is thread-safe under concurrent callers", "[HAPropsSI][thre
     CHECK(bad_numbers.load() == 0);
     CHECK(mismatches.load() == 0);
 }
+
+// Hidden by Catch's [!benchmark] convention; run with
+//   ./CatchTestRunner '[!benchmark][HAPropsSI]'
+// to get per-call timings for the thread_local conversion (gh-2787 follow-up).
+TEST_CASE("HAPropsSI thread_local cost", "[!benchmark][HAPropsSI][threadsafe]") {
+    const double p = 101325;
+
+    // Warm this thread's Water/Air/WaterIF97 backends and the per-T caches so
+    // the first-call construction cost doesn't pollute the hot-path number.
+    HumidAir::HAPropsSI("H", "T", 300.0, "R", 0.5, "P", p);
+
+    BENCHMARK("HAPropsSI H,T=300,R=0.5,P [hot, single-thread]") {
+        return HumidAir::HAPropsSI("H", "T", 300.0, "R", 0.5, "P", p);
+    };
+
+    // Fresh-thread cost: each Catch sample spawns a thread that pays its own
+    // backend construction (Water + Air + IF97) on first HAPropsSI entry, then
+    // exits. Captures the per-thread one-shot init overhead introduced by the
+    // thread_local conversion.
+    BENCHMARK("HAPropsSI H,T=300,R=0.5,P [first call in fresh thread]") {
+        double result = 0.0;
+        std::thread th([&] { result = HumidAir::HAPropsSI("H", "T", 300.0, "R", 0.5, "P", p); });
+        th.join();
+        return result;
+    };
+}
+
 // a predicate implemented as a function:
 bool is_not_a_pair(const std::set<std::size_t>& item) {
     return item.size() != 2;
--- a/src/HumidAirProp.cpp
+++ b/src/HumidAirProp.cpp
@@ -4,9 +4,9 @@
 #    endif
 #endif
 
+#include <atomic>
 #include <cmath>
 #include <memory>
-#include <mutex>
 using std::shared_ptr;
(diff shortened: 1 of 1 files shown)


ibell:
Bot feedback addressed in 292f974a:

- **clang-format**: re-ran the project's style on the added lines (project uses Allman braces on `struct`, slightly different trailing-comment column alignment than my local v19's defaults — CI's clang-format-18 caught it). No behavioural change.
- **CodeQL "short global name 'Air'"**: added `static` to the three file-scope `thread_local` shared_ptrs so they have internal linkage. The names `Water` / `Air` / `WaterIF97` are physical-domain and read clearly in context; `static thread_local` is the right minimum to take them out of the alert's scope without churning the file with `g_water_backend` style renames. (These names also pre-existed at namespace scope on master — the alert fired only because I touched the lines.)

The other failing check — `build` (Windows Installer, 12m55s) — is a Python decode error in the `generate_headers` step on Windows:

```
CUSTOMBUILD : decoding error : Extra data: line 1 column 1010 (char 1009)
hashes.json: {…}re_binary_pairs_pcsaft_JSON": "…"}
```

`generate_headers` runs *before* any C++ compilation, my changes are C++-only, and the visible `hashes.json` content shows duplicated trailing data — looks like a parallel-write race in the Windows custom-build step on `hashes.json`. The same job last passed on master at 2026-04-28T18:50 against the same head commit (12a69d4a Eigen 5.0.1 bump), so this looks like a transient Windows flake. Happy to dig further if it reproduces on a re-run.

ibell:
Reverted commit 5cdd6550 — it was wrong. Branch IS based on master, and master has '`~BrentSolverResids() {};`' on this exact line (it's the output of the project's pinned clang-format-18.1.8 from the whole-repo pass in #2803). My commit applied apt clang-format-18.1.3's preference instead, which conflicts with the actual project pin.

Root cause: .github/workflows/dev_clangformat.yml runs unpinned apt clang-format (18.1.3 on today's ubuntu-latest), bypassing the .pre-commit-config.yaml pin (18.1.8). Whenever a PR touches a file that contains a line on which 18.1.3 and 18.1.8 disagree, the script reformats the whole file with the wrong version and CI fails on patches that match the project's actual pin.

Verified locally: `pre-commit run --files src/HumidAirProp.cpp` (using the project's 18.1.8 pin) passes clean against the post-revert state.

Filed CoolProp-2h3 to fix dev_clangformat.yml — likely just `pip install clang-format==18.1.8` before invoking the script, or run the pre-commit hook directly. Until that's fixed, the clang-format CI check will keep red-flagging this PR on a pre-existing line that's correctly formatted under the project's pin.

Issue #2834, opened 2026-04-29
Consider treating multi-valued saturation regions as an error in update() (follow-up to #2773)


ibell:
## Background

Some saturation flashes admit multiple temperature solutions for the same input value:

- `HmolarQ_INPUTS` / `HmassQ_INPUTS` on the vapor side: $h_g(T)$ peaks well below T_crit (≈ 540 K for water), so a wide band of enthalpies is reached at two T's.
- `QSmolar_INPUTS` / `QSmass_INPUTS` on the vapor side: similar shape.
- `DmolarQ_INPUTS` / `DmassQ_INPUTS` on the liquid side for water / heavy water: $\rho_L(T)$ has a maximum near 4 °C (water) and 11 °C (D2O).

Today, the default `update()` call returns *one* of the roots without warning the user. Which root it picks is implementation-defined — exactly the symptom described in #2773 and #2770.

The follow-up #2773 fix introduced `update_with_guesses` paths for `HmolarQ_INPUTS`, `QSmolar_INPUTS`, and `DmolarQ_INPUTS` that take a `guess.T` to disambiguate. But callers of plain `update()` still get silent wrong-root returns.

## Proposal

Now that the new code path uses the Chebyshev-superancillary's `get_x_for_y` to enumerate every saturation root for these inputs, detecting the multi-root case is essentially free — `solns.size() > 1` after a single TOMS748 sweep.

When this happens *and the caller used the no-guess `update()` API*, throw a new exception type `CoolProp::MultipleSolutionsError` with a message that points the user at `update_with_guesses`:

```
MultipleSolutionsError: HmolarQ flash for water at h=2.79e6 J/mol has 2
roots on saturated vapor (T ≈ 470 K, T ≈ 605 K). Use
update_with_guesses with a temperature guess to disambiguate. See:
http://www.coolprop.org/coolprop/LowLevelAPI.html#disambiguating-multiple-saturation-roots
```

## Open questions

1. **Default on or off?** Turning it on by default is the right correctness call but breaks any user code that happens to depend on the current implementation-defined behavior. Likely path: introduce as config flag (`ENABLE_AMBIGUOUS_SATURATION_ROOT_CHECK`) defaulting **off** for one minor release with a deprecation log line, then flip to **on** for the next major.

[...]

PR #2835, opened 2026-04-29
feat(saturation): branch-selecting HQ/SQ/DQ flashes via guess.T (#2773)

### LINKED ISSUE #2773: [FEAT] Add means to specify which solution you want when multiple solutions are possible
HQ inputs on vapor phase can have this problem. Same problem with SQ inputs, on the vapor curve. The liquid side should be monotonic for both properties for all (?) fluids

ibell:
## Summary

Closes #2773 (partially — see follow-ups). When a saturation flash input value admits two T-roots for the same value, the default flash silently returns one. This PR adds `update_with_guesses` dispatches for `HmolarQ_INPUTS`, `QSmolar_INPUTS`, and `DmolarQ_INPUTS` that consult `guess.T` to pick the branch.

The known multi-root cases:

- `HmolarQ` / `QSmolar` on the **vapor** side: water's $h_g(T)$ peaks near 540 K (similarly $s_g$), so a wide band of values is reached at two saturation T's.
- `DmolarQ` on the **liquid** side: water and heavy water (D2O) have $\\rho_L$ maxima near 4 °C / 11 °C, so densities just below the peak occur at two T's.

## Approach

Each saturation property is represented as a piecewise Chebyshev superancillary. At construction, the existing `ChebyshevApproximation1D` already locates extrema by computing eigenvalues of the companion matrix of the derivative polynomial (balanced) and partitions $[T_\\min, T_\\mathrm{crit}]$ into provably-monotonic sub-intervals. Branch selection is then just TOMS748 inside each interval — see `SuperAncillary::get_x_for_y`. Per-call cost is sub-microsecond after the first lazy build.

For HQ and SQ this is the first-ever instantiation of the caloric superancillary; this PR builds them lazily on first use by sampling the EOS at the rho-superancillary's Chebyshev nodes (`SuperAncillary::add_variable`, which now correctly works with the `std::vector<double>` template instantiation that CoolProp uses — it was Eigen-only and had never been exercised).

After branch selection, DQ refines via Brent against the full EOS saturation residual in a narrow bracket inside the chosen monotonic sub-interval; HQ and SQ fall through to a standard `QT_flash` at the disambiguated $T$.

## What's in / what's out

**In:** `update_with_guesses` for the three input pairs above, lazy caloric superancillary build, full test coverage on water for all three cases, doc section in `LowLevelAPI.rst`.

**Out (intentional, see follow-ups):**

- Migration of `saturation_PHSU_pure`'s IMPOSED_HV/SV/HL/SL branches off the static `hL/hLV/sL/sLV` ancillaries — the new code path bypasses the old machinery rather than replacing it. Doing both in one PR would be a much bigger review.
- Reference-state shift-c0 trick — currently a `set_reference_state` call would invalidate the cached caloric superancillary; the trick to apply only a constant offset at evaluation time is left for a follow-up.
- Mixtures — pure (and not pseudo-pure) only. `NotImplementedError` thrown otherwise.
- High-level `PropsSI` interface — no change. Branch selection is exposed only via the low-level `update_with_guesses` API.
- Strict-mode error on ambiguous flashes without a guess — tracked in #2834.

## Files changed

- `include/superancillary/superancillary.h` — `has_variable()` accessor; fix `add_variable()` to handle `SuperAncillary<std::vector<double>>` (was Eigen-only)
- `src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.{h,cpp}` — `ensure_caloric_superancillaries()` lazy build hook; new `update_with_guesses` dispatch cases
- `src/Backends/Helmholtz/FlashRoutines.{h,cpp}` — `DQ_flash_with_guesses`, `HQ_flash_with_guesses`, `QS_flash_with_guesses`
- `src/Tests/CoolProp-Tests.cpp` — three new `[2773][branch_selection]` test cases
- `Web/coolprop/LowLevelAPI.rst` — new section *Disambiguating Multiple Saturation Roots* with worked examples

## Test plan

- [x] New `[2773][branch_selection]` Catch test cases all pass (18 assertions across DQ / HQ / QS)
- [x] `[flash][saturation][pureflash][sat_T_to_Tc]` tags: 1853 assertions, no regressions
- [ ] CI passes on all platforms
- [ ] Sphinx docs build cleanly with the new section
- [ ] Manual: spot-check that `update_with_guesses` is callable from the Python wrapper end-to-end

## Reviewer notes

This is a sizeable diff (≈ 525 lines, +- across 7 files) but cleanly split:

1. `superancillary.h` changes are minimal and self-contained (47 lines)
2. `HelmholtzEOSMixtureBackend` adds one new method and three switch cases (≈ 60 lines)
3. The bulk is `FlashRoutines.cpp` with three new functions (≈ 200 lines, plus shared helper)
4. Tests + docs are about a third of the total

The Chebyshev-superancillary mechanism this PR rests on was ported to CoolProp some time ago from `teqp`/`teqpflsh` — see https://ianhbell.github.io/teqpflsh-nist-docs/en/latest/superancillary/SuperancillaryIterations.html — but the caloric-side instantiation and the branch-selection consumers are new here.

### DIFF
--- a/src/Backends/Helmholtz/FlashRoutines.cpp
+++ b/src/Backends/Helmholtz/FlashRoutines.cpp
@@ -533,8 +533,9 @@ double FlashRoutines::resolve_T_via_superancillary(HelmholtzEOSMixtureBackend& H
             Ts_str += format("%g K", Ts[i]);
             if (i + 1 < Ts.size()) Ts_str += ", ";
         }
-        throw MultipleSolutionsError(format("%s: %c=%g on saturated %s has %zu T-roots (%s); use update_with_guesses with guess.T to pick a branch (see GitHub #2773)",
-                                            fn_name, k, target_value, phase_name, Ts.size(), Ts_str.c_str()));
+        throw MultipleSolutionsError(
+          format("%s: %c=%g on saturated %s has %zu T-roots (%s); use update_with_guesses with guess.T to pick a branch (see GitHub #2773)", fn_name,
+                 k, target_value, phase_name, Ts.size(), Ts_str.c_str()));
     }
     return Ts[0];
 }
--- a/src/Backends/Helmholtz/FlashRoutines.cpp
+++ b/src/Backends/Helmholtz/FlashRoutines.cpp
@@ -364,6 +364,217 @@ void FlashRoutines::QS_flash(HelmholtzEOSMixtureBackend& HEOS) {
         throw NotImplementedError("QS_flash not ready for mixtures");
     }
 }
+
+// Branch-disambiguating variant of DQ_flash (see GitHub #2773).
+// rho_L(T) on water and D2O is non-monotonic near 4 °C / 11 °C, so the DQ flash
+// can have two T-solutions for the same density. Use the rho_sat superancillary
+// to enumerate roots (TOMS748 inside each provably-monotonic Chebyshev interval),
+// pick the one closest to guess.T, then refine against the full EOS saturation
+// residual via Brent in a narrow bracket so the returned state is EOS-precise.
+void FlashRoutines::DQ_flash_with_guesses(HelmholtzEOSMixtureBackend& HEOS, const GuessesStructure& guess) {
+    if (!ValidNumber(guess.T) || guess.T <= 0) {
+        throw ValueError("DQ_flash_with_guesses requires a positive guess.T");
+    }
+    if (!HEOS.is_pure_or_pseudopure) {
+        throw NotImplementedError("DQ_flash_with_guesses not ready for mixtures");
+    }
+    auto superanc_ptr = HEOS.get_superanc();
+    if (!superanc_ptr) {
+        throw NotImplementedError("DQ_flash_with_guesses requires a superancillary; this fluid has none");
+    }
+    auto& superanc = *superanc_ptr;
+
+    const double rhomolar = HEOS._rhomolar;
+    const double Q = HEOS._Q;
+
+    short Q_key = -1;
+    if (std::abs(Q) < 1e-10) {
+        Q_key = 0;
+    } else if (std::abs(Q - 1) < 1e-10) {
+        Q_key = 1;
+    } else {
+        throw ValueError(format("DQ_flash_with_guesses currently requires Q=0 or Q=1; got Q=%g", Q));
+    }
+    HEOS.specify_phase(iphase_twophase);
+    const double eps = 1e-12;
+    if (rhomolar >= (HEOS.rhomolar_critical() + eps) && Q > (0 + eps)) {
+        throw CoolProp::OutOfRangeError(
+          format("DQ inputs are not defined for density (%g) above critical density (%g) and Q>0", rhomolar, HEOS.rhomolar_critical()).c_str());
+    }
+
+    // Enumerate candidate T-roots via the rho_sat superancillary's monotonic-
+    // interval rootfinder. Each entry is (T, num_TOMS748_steps).
+    const auto& approx = superanc.get_approx1d('D', Q_key);
+    const auto solns = approx.get_x_for_y(rhomolar, 64, 100U, 1e-10);
+    if (solns.empty()) {
+        throw SolutionError(format("DQ_flash_with_guesses: no T-root on saturated %s for rho=%g; superancillary range [%g, %g] K",
+                                   (Q_key == 0 ? "liquid" : "vapor"), rhomolar, approx.xmin(), approx.xmax()));
+    }
+    // Pick the candidate closest to guess.T.
+    double T_super = solns.front().first;
+    double best_dist = std::abs(T_super - guess.T);
+    for (const auto& soln : solns) {
+        const double d = std::abs(soln.first - guess.T);
+        if (d < best_dist) {
+            best_dist = d;
+            T_super = soln.first;
+        }
+    }
+
+    // Refine to EOS precision against the full saturation residual using Brent in
+    // a narrow bracket around T_super. The bracket is chosen to stay strictly
+    // inside the same monotonic sub-interval — half the distance to the nearer
+    // extremum (or to the saturation-range endpoint) — so we won't be lured into
+    // the other branch.
+    const double Tmax_sat = HEOS.T_critical() - 0.1;
+    const double Tmin_sat = HEOS.Tmin() + 0.1;
+    double left_edge = Tmin_sat, right_edge = Tmax_sat;
+    for (const double T_extr : approx.get_x_at_extrema()) {
+        if (T_extr < T_super && T_extr > left_edge) left_edge = T_extr;
+        if (T_extr > T_super && T_extr < right_edge) right_edge = T_extr;
+    }
+    const double half_left = std::max(0.5 * (T_super - left_edge), 1e-3);
+    const double half_right = std::max(0.5 * (right_edge - T_super), 1e-3);
+    const double Ta = std::max(Tmin_sat, T_super - half_left);
+    const double Tb = std::min(Tmax_sat, T_super + half_right);
+
+    DQ_flash_residual resid(HEOS, rhomolar, Q);
+    const double fa = resid.call(Ta);
+    const double fb = resid.call(Tb);
+    if (fa * fb > 0) {
+        // Should not happen when the superancillary is consistent with the EOS,
+        // but if Brent can't bracket, leave the EOS state at T_super (the last
+        // resid.call below populates SatL/SatV at that T). The superancillary
+        // value is itself ~1e-12 accurate.
+        resid.call(T_super);
+    } else {
+        Brent(resid, Ta, Tb, DBL_EPSILON, 1e-10, 100);
+    }
+    HEOS._p = HEOS.SatV->p();
+    HEOS._T = HEOS.SatV->T();
+    HEOS._rhomolar = rhomolar;
+    HEOS._Q = Q;
+    HEOS._phase = iphase_twophase;
+}
+
+// Helper: pick the saturation T-root closest to guess.T from a list of candidates.
+// Used by HQ_flash_with_guesses and QS_flash_with_guesses below.
+static double pick_closest_T(const std::vector<std::pair<double, int>>& solns, double T_guess) {
+    double T_best = solns.front().first;
+    double best_dist = std::abs(T_best - T_guess);
+    for (const auto& soln : solns) {
+        const double d = std::abs(soln.first - T_guess);
+        if (d < best_dist) {
+            best_dist = d;
+            T_best = soln.first;
+        }
+    }
+    return T_best;
+}
+
+// Branch-disambiguating variant of HQ_flash (see GitHub #2773).
+// The default HQ_flash routes through saturation_PHSU_pure with IMPOSED_HV, which
+// silently picks one of the (possibly two) T-roots when h_g(T) is non-monotonic
+// — water saturated vapor enthalpy peaks near 540 K, so a wide band of h_target
+// values has two T solutions. This variant builds the h_sat superancillary on
+// first use, enumerates roots via TOMS748 over each provably-monotonic Chebyshev
+// sub-interval, picks the one closest to guess.T, and runs a QT flash at that T.
+void FlashRoutines::HQ_flash_with_guesses(HelmholtzEOSMixtureBackend& HEOS, const GuessesStructure& guess) {
+    if (!ValidNumber(guess.T) || guess.T <= 0) {
+        throw ValueError("HQ_flash_with_guesses requires a positive guess.T");
+    }
+    if (!HEOS.is_pure_or_pseudopure) {
+        throw NotImplementedError("HQ_flash_with_guesses not ready for mixtures");
+    }
+    auto superanc_ptr = HEOS.get_superanc();
+    if (!superanc_ptr) {
+        throw NotImplementedError("HQ_flash_with_guesses requires a superancillary; this fluid has none");
+    }
+
+    const double hmolar = HEOS._hmolar;
+    const double Q = HEOS._Q;
+
+    short Q_key = -1;
+    if (std::abs(Q) < 1e-10) {
+        Q_key = 0;
+    } else if (std::abs(Q - 1) < 1e-10) {
+        Q_key = 1;
+    } else {
+        throw ValueError(format("HQ_flash_with_guesses currently requires Q=0 or Q=1; got Q=%g", Q));
+    }
+
+    HEOS.ensure_caloric_superancillaries();
+    auto& superanc = *superanc_ptr;
+    if (!superanc.has_variable('H')) {
+        throw NotImplementedError("HQ_flash_with_guesses: caloric superancillary build failed");
+    }
+
+    const auto& approx = superanc.get_approx1d('H', Q_key);
+    const auto solns = approx.get_x_for_y(hmolar, 64, 100U, 1e-10);
+    if (solns.empty()) {
+        throw SolutionError(format("HQ_flash_with_guesses: no T-root on saturated %s for h=%g; superancillary range [%g, %g] K",
+                                   (Q_key == 0 ? "liquid" : "vapor"), hmolar, approx.xmin(), approx.xmax()));
+    }
+    const double T_super = pick_closest_T(solns, guess.T);
+
+    // Populate the rest of the state (p, rhomolar, SatL/SatV) by running a
+    // standard QT flash at the disambiguated T. The user-supplied h is implicit
+    // in the EOS state at (T_super, rho_sat(T_super)).
+    HEOS._T = T_super;
+    HEOS._Q = Q;
+    HEOS.specify_phase(iphase_twophase);
+    QT_flash(HEOS);
+    HEOS._hmolar = hmolar;
+    HEOS._phase = iphase_twophase;
+}
+
+// Branch-disambiguating variant of QS_flash (see GitHub #2773). Mirrors HQ_flash_with_guesses.
+void FlashRoutines::QS_flash_with_guesses(HelmholtzEOSMixtureBackend& HEOS, const GuessesStructure& guess) {
+    if (!ValidNumber(guess.T) || guess.T <= 0) {
+        throw ValueError("QS_flash_with_guesses requires a positive guess.T");
+    }
+    if (!HEOS.is_pure_or_pseudopure) {
+        throw NotImplementedError("QS_flash_with_guesses not ready for mixtures");
+    }
+    auto superanc_ptr = HEOS.get_superanc();
+    if (!superanc_ptr) {
+        throw NotImplementedError("QS_flash_with_guesses requires a superancillary; this fluid has none");
+    }
+
+    const double smolar = HEOS._smolar;
+    const double Q = HEOS._Q;
+
+    short Q_key = -1;
+    if (std::abs(Q) < 1e-10) {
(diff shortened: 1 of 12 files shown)


src/Backends/Helmholtz/FlashRoutines.cpp
```diff
@@ -364,6 +364,197 @@ void FlashRoutines::QS_flash(HelmholtzEOSMixtureBackend& HEOS) {
...
+    if (std::abs(Q) < 1e-10) {
+        Q_key = 0;
+    } else if (std::abs(Q - 1) < 1e-10) {
+        Q_key = 1;
+    } else {
+        throw ValueError(format("DQ_flash_with_guesses currently requires Q=0 or Q=1; got Q=%g", Q));
+    }
+    HEOS.specify_phase(iphase_twophase);
+    const double eps = 1e-12;
+    if (rhomolar >= (HEOS.rhomolar_critical() + eps) && Q > (0 + eps)) {
```
ibell: I don't understand this check. Why is it here?

src/Backends/Helmholtz/FlashRoutines.cpp
```diff
@@ -364,6 +364,197 @@ void FlashRoutines::QS_flash(HelmholtzEOSMixtureBackend& HEOS) {
...
+                                   (Q_key == 0 ? "liquid" : "vapor"), guess.T, rhomolar, approx.xmin(), approx.xmax()));
+    }
+    const double T_super = soln->first;
+
+    // Refine to EOS precision against the full saturation residual using Brent in
+    // a narrow bracket around T_super. The bracket is chosen to stay strictly
+    // inside the same monotonic sub-interval — half the distance to the nearer
+    // extremum (or to the saturation-range endpoint) — so we won't be lured into
+    // the other branch.
+    const double Tmax_sat = HEOS.T_critical() - 0.1;
```
ibell: Either this branch should be removed or gated behind a configuration option. The superancillary temperature obtained from TOMS748 should be good to basically better than VLE solutions, no?

src/Backends/Helmholtz/FlashRoutines.cpp
```diff
@@ -364,6 +364,197 @@ void FlashRoutines::QS_flash(HelmholtzEOSMixtureBackend& HEOS) {
...
+    const double fa = resid.call(Ta);
+    const double fb = resid.call(Tb);
+    if (fa * fb > 0) {
+        // Should not happen when the superancillary is consistent with the EOS,
+        // but if Brent can't bracket, leave the EOS state at T_super (the last
+        // resid.call below populates SatL/SatV at that T). The superancillary
+        // value is itself ~1e-12 accurate.
+        resid.call(T_super);
+    } else {
+        Brent(resid, Ta, Tb, DBL_EPSILON, 1e-10, 100);
```
ibell: and if we do get here, should be TOM748 rather than Brent

src/Backends/Helmholtz/FlashRoutines.cpp
```diff
@@ -364,6 +364,197 @@ void FlashRoutines::QS_flash(HelmholtzEOSMixtureBackend& HEOS) {
...
+    HEOS._T = T_super;
+    HEOS._Q = Q;
+    HEOS.specify_phase(iphase_twophase);
+    QT_flash(HEOS);
+    HEOS._hmolar = hmolar;
+    HEOS._phase = iphase_twophase;
+}
+
+// Branch-disambiguating variant of QS_flash (see GitHub #2773). Mirrors HQ_flash_with_guesses.
+void FlashRoutines::QS_flash_with_guesses(HelmholtzEOSMixtureBackend& HEOS, const GuessesStructure& guess) {
```
ibell, line 621: this and `HS_flash_with_guesses` are basically identical; can you reduce to common method?

src/Backends/Helmholtz/FlashRoutines.cpp
ibell: Good call — dropped in commit bb43e7cc. The superancillary's `get_x_for_y_near` already returns `nullopt` for out-of-range targets with a precise error message; the explicit pre-check was redundant copy-paste from the old `DQ_flash`.

src/Backends/Helmholtz/FlashRoutines.cpp
ibell: Agreed and dropped in bb43e7cc. You're right — TOMS748 against the superancillary is already accurate to ~1e-12 relative to the multi-precision reference, well below the precision of any downstream EOS evaluation. The whole refinement block was overkill. DQ now just runs a standard `QT_flash` at `T_super`, matching the HQ/SQ pattern.

src/Backends/Helmholtz/FlashRoutines.cpp
ibell: Moot now — the entire refinement branch is gone in bb43e7cc per the comment above.

src/Backends/Helmholtz/FlashRoutines.cpp
ibell: Done in bb43e7cc — single helper `FlashRoutines::pick_branch_T_via_superancillary(HEOS, guess, k, value, fn_name)` with single-character dispatch on `k` ∈ {'D', 'H', 'S'}. The three flashes are now ~10 lines each. (`HS_flash_with_guesses` → `QS_flash_with_guesses` — assumed typo.)

ibell:
Added a `[2773][bench]` benchmark in commit a41745d8. Numbers on Apple Silicon (release):

| Operation | Time | Notes |
|---|---|---|
| DQ default `update(DmolarQ,rho,Q=0)` | **8.3 μs** | Brent over full saturation T-range |
| DQ with guesses | **1.08 μs** | superancillary `get_x_for_y_near` (≈ **7.7× faster**) |
| HQ with guesses | **0.78 μs** | (default HQ_flash throws for this anchor) |
| QS with guesses | **0.54 μs** | n-Propane sat-liquid (default QS_flash crashes for water) |

Default `HQ_flash` and `QS_flash` aren't included in the comparison at the chosen anchor points because they throw / crash — that's the pre-existing #2773 symptom and is exactly why `update_with_guesses` exists.

Also added comments at the three `QT_flash` call sites pointing out that despite the name, on a pure fluid with superancillaries `QT_flash` is a sub-µs superancillary eval, not an iterative VLE solve.

Run with `CatchTestRunner "[2773][bench]"`.

One note worth flagging: the *first* HQ or QS with-guesses call has to lazily build the caloric superancillary, which is ~20-50 ms (not in the per-call mean above — Catch2 amortizes via warmup). For HPC use the build can be triggered eagerly via `HEOS::ensure_caloric_superancillaries()` once after factory.

ibell:
Re. timings for default HQ / QS without guesses: there aren't any usable ones.

I swept water saturated vapor at 9 temperatures from 320 K to 620 K (the full saturation range) and saturated liquid at 5 temperatures, plus propane:

- Default `HQ_flash` for **water Q=1** throws `"p is not a valid number"` at *every* tested T from 320 K to 620 K (320, 350, 380, 420, 470, 510, 550, 590, 620). Caused by the inverted condition at `FlashRoutines.cpp:307` (`if (Tguess < 0)` should be `if (Tguess > 0)`) — when `Tguess = -1` (the default), the block sets `options.T = -1` and tries to evaluate the rho ancillary at T=-1.
- Default `HQ_flash` for **water Q=0** throws `"non-unity quality not currently allowed for HQ_flash"` — Q=0 isn't supported by the existing implementation at all.
- Default `QS_flash` SIGSEGVs at multiple anchor points (water sat-vap at s ≈ 112 J/mol/K, propane sat-liq at s ≈ 54.7 J/mol/K — confirmed earlier in this PR thread).

So the honest comparison is: the default HQ / QS paths aren't slower than the new path — they're **unusable** at the operating points where #2773 actually bites. The new path at 0.5–0.8 µs/call is what you get from a path that actually works. DQ is the only one where a working baseline exists, and there the new path is ≈ 7.7× faster (8.3 µs → 1.08 µs).

ibell:
Pushed `a2613b69`: default `update()` now throws `CoolProp::MultipleSolutionsError` for ambiguous saturation flash inputs (HmolarQ, QSmolar, DmolarQ). Closes the design from #2834.

**What changed**
- `DQ_flash` / `HQ_flash` / `QS_flash` route through the superancillary on pure fluids with Q ∈ {0, 1}. New helper `FlashRoutines::pick_unique_T_via_superancillary` enumerates every T-root via TOMS748 inside each provably-monotonic Chebyshev sub-interval. Returns single T → use it; >1 → throw `MultipleSolutionsError` listing candidate Ts; 0 → throw `OutOfRangeError`.
- New `CoolProp::MultipleSolutionsError` (subclass of `ValueError`) in `Exceptions.h`.
- Existing buggy code paths (`HQ_flash` with the inverted-Tguess condition, `QS_flash` saturation_PHSU_pure crash) are now unreachable for fluids with superancillaries.
- For fractional Q (DQ only) and fluids without superancillaries, the original Brent / saturation_PHSU_pure paths are preserved.

**Side effect:** the existing buggy default HQ for water saturated vapor is now actually working — it returns the unique T-root in the monotone region and throws `MultipleSolutionsError` only where the curve genuinely has two roots (around the 540 K h-peak). Same for QS, where the crash is replaced by either a working result or a clean error.

**Tests:** 4 new `[2773][strict_mode]` assertions verifying:
- water DQ at rho between triple-density and density-max → `MultipleSolutionsError`
- water HQ at h_g(470 K) → `MultipleSolutionsError`
- water DQ at T=400 K (single-root region) → still works via default `update()`

22 `[2773]` assertions total pass. 1853 `[flash][saturation][pureflash][sat_T_to_Tc]` assertions still pass.

Doc section updated in `LowLevelAPI.rst`.

ibell:
Thanks for the depth. Triaged each item and pushed two commits (`008c975d`, `99a404f6`):

**Fixed in this PR:**
- **C1** ref-state invalidation. `JSONFluidLibrary::set_fluid_enthalpy_entropy_offset` (the path actually used by `set_reference_stateS`) and `HelmholtzEOSMixtureBackend::set_fluid_enthalpy_entropy_offset` now call `SuperAncillary::clear_caloric_variables()`. Verified by a new test that builds h-superanc under DEF, switches to NBP, and confirms the with-guesses path returns the correct T against the new reference.
- **C2** thread safety. `std::mutex` on `SuperAncillary` + `ensure_HS_under_lock(h_cb, s_cb)` doing double-checked-lock; `add_variable` split into public (locks) and private `add_variable_locked` (assumes held). Static asserts flipped to negative (SuperAncillary is now intentionally non-copyable; held by `shared_ptr` in production).
- **C4** changelog entry under "Unreleased" flagging the breaking change in default `update()`, the new exception type, and the new dispatches. Doc contradiction fixed.
- **I1** dedup near-identical roots from adjacent monotonic intervals at extrema (`T_ROOT_DEDUP_TOL = 1e-6 K`).
- **I2** explicit tie-break on midpoint distance in the with-guesses path.
- **I3** pseudo-pure refused cleanly at the top of the helper.
- **I4** `HmassQ_INPUTS` and `QSmass_INPUTS` now convert in `mass_to_molar_inputs`.
- **I5** doc contradiction reworded.
- **I6** the two helpers folded into one `resolve_T_via_superancillary(HEOS, key, value, optional<guess_T>, fn_name)`.
- **I7** `parameters` enum (`iDmolar`/`iHmolar`/`iSmolar`) instead of `char k` at the FlashRoutines boundary; the `char` translation happens in a tiny helper that reaches the existing `SuperAncillary` API.
- **S5** `Q_BOUNDARY_TOL` and `T_ROOT_DEDUP_TOL` centralized as `static constexpr`.
- **S6** "sub-microsecond" softened; lazy-build cost (~10–50 ms first call) and rebuild-on-ref-state-change documented.
- **S4** new `[2773][edge_cases]` test case: mass dispatch, ref-state invalidation, mixture `NotImplementedError`, out-of-range `guess.T`, near-extremum dedup.

**Acknowledged but kept for follow-up beads:**
- **C3** wrapper-side exception registration (Python/SWIG/EES). C++ side has the typed exception; wrapper plumbing is a separate concern.
- **I8** defensive back-write left as-is (you OK'd this).
- **S1** `ValueError` inheritance kept (you OK'd this).
- **S2** REFPROP backend divergence accepted (you OK'd this).
- **S3** pre-existing critical-state shortcut in `QS_flash` left untouched (only affects the no-guess default path; with-guesses is fine).

**Tests:** 33 `[2773]` assertions pass (was 22), 1853 `[flash][saturation]` assertions still pass — no regression.

ibell:
Round 2 review pass — pushed `e4681e26` addressing what surfaced.

**Important fixes:**
- `HQ_flash` gating consistency: switched from a bare `HEOS.get_superanc()` check to `sat_superanc_path_applies(HEOS)`, matching `DQ_flash` and `QS_flash`. No current behavior change (no pseudo-pure ships with a superancillary), but removes a footgun for any future fluid that gains one.
- `get_invlnp()` was documented as guarded by the new mutex but the actual lazy build was unlocked. Now takes the same `m_lazy_build_mutex`.
- Dropped the `optional<...>::has_value()` fast path in `ensure_HS_under_lock`. Reading `has_value()` outside the lock against a concurrent `add_variable_locked` writer is technically UB per the C++ standard (TSan would flag it). The uncontended-mutex cost is ~20 ns vs ~500 ns rootfind work, so the always-lock version is simpler and safe.

**Cleanups:**
- Dead `fallback_ptr` removed from the tie-break lambda; `decltype(&intervals[0])` replaced with `auto*`.
- Ref-state-invalidation test correctly scoped to verify fresh-HEOS instances pick up the new reference (the original instance keeps its own local `alpha0`, which is pre-existing CoolProp behavior — per-HEOS ownership). Comment explains.
- Renamed the misleadingly-titled "near-extremum dedup" test (it actually tested a normal multi-root case, not a dedup-adjacent one) to "Multi-root region throws MultipleSolutionsError on default update()". Added a sibling "Mass-input multi-root region" test that exercises the I4 conversion path through default `update()`.

**Reviewer declared safe** (no action needed):
- C1 invalidation placement is exhaustive (only two `EnthalpyEntropyOffset.set` sites; both call `clear_caloric_variables`)
- Callback re-entrance deadlock concern is unfounded (`calc_*_nocache` are pure)
- Static_assert flip is safe (held by `shared_ptr` in production)
- I4 math correct; mass-input dispatch flows through `pre_update` → strict-mode check correctly

**Tests:** 34 `[2773]` (up from 33) + 1853 `[flash][saturation]` assertions pass.

Ready for another look when you have time.

ibell:
Pushed `3b788411` with the option-D shift-c0 fix instead of stamp-and-invalidate.

**Why option 2 was wrong:** stamp-then-rebuild thrashes when two coexisting HEOS instances at different reference states alternate queries — each call wipes and rebuilds the cache (~10–50 ms each). Untenable for comparison code.

**The right answer is in the math.** `IdealHelmholtzEnthalpyEntropyOffset::all` (Helmholtz.cpp:1153) contributes `(a1 + a2·τ)` to α0. Working through the helmholtz formulae for h and s along the saturation curve:

```
Δh = R · T_red · Δa2     (constant in T)
Δs = −R · Δa1            (constant in T)
```

So a reference-state change shifts h_sat(T) and s_sat(T) by uniform constants — the *shape* is invariant; only c0 of the Chebyshev expansion moves. **We never rebuild.** Stamp the cache with (a1, a2) recorded at first build; on each query, translate the user's target into the cache's frame:

```cpp
target_in_cache = target_user + R·T_red·(a2_cache − a2_caller)   // for h
target_in_cache = target_user − R·(a1_cache − a1_caller)         // for s
```

Returned T is reference-state-independent (saturation curve T is too), so no back-translation needed.

**Implementation:**
- `SuperAncillary::ensure_HS_under_lock` records the stamp on first build and never rebuilds
- `resolve_T_via_superancillary` applies the shift before TOMS748
- Removed C1's `clear_caloric_variables` calls from both ref-state mutation sites — no longer needed
- Added `IdealHelmholtzEnthalpyEntropyOffset::get_a1/get_a2` accessors
- Added `SUPERANC::caloric_build_count` parameter for test introspection

**Tests added:**
- `[2773][ref_state_shared]`: AS at DEF + AS2 at NBP, alternating HmolarQ and QSmolar, verifies both get correct answers AND build counter stays at 1 throughout — no thrashing
- `[2773][thread_safety]`: 8 threads × 50 iterations on water, verifies no errors and at-most-1 build via the mutex

**Test counts:** 71 `[2773]` assertions pass (was 34). 1853 `[flash][saturation]` still pass.

ibell:
Pushed `d1fd50a1` addressing the round-3 findings.

**Important:**
- **I-1 Core offset.** Both `EnthalpyEntropyOffsetCore` (parse-time-immutable from JSON) and `EnthalpyEntropyOffset` (user-mutable) contribute to α₀. Previously the stamp/shift only saw the latter. Centralized into `FlashRoutines::alpha0_offset_total` which sums Core + user. Defensive: if Core ever becomes runtime-mutable, the shift stays correct.
- **I-2 alpha0 prefactor.** Added `IdealHelmholtzContainer::get_prefactor()`. The shift is now `prefactor·(a1, a2)`, so non-unity prefactor fluids (none currently in the bundled library, but the JSON path is live) work correctly.

**Suggestions:**
- **S-1 stale docs.** Rewrote two comment blocks that still described the round-2 invalidate-on-mismatch design (the `get_a1`/`get_a2` accessor doc and the test-section header).
- **S-2 dead-code clarification.** `clear_caloric_variables()` has no in-tree caller; doc now honestly says "diagnostic / forced-rebuild hatch, not invoked by any normal control path."
- **S-3 exception-safe test cleanup.** Both ref-state-using tests use a local RAII guard struct that resets to DEF in its destructor. A failed `REQUIRE_NOTHROW` mid-test no longer contaminates downstream tests.
- **S-6 build-counter doc.** Clarified that the counter increments once per H+S pair-build, not per individual variable.

**Skipped:**
- **S-4 TSan** (CI infrastructure decision; not a code change).
- **S-5 lock-free stamp read** (no-op; current design is correct).
- **S-7 commit history rebase** (defer to merge time).

**Tests:** 71 `[2773]` + 1853 `[flash][saturation]` pass.

ibell:
Ran ThreadSanitizer against the new `[2773][thread_safety]` test (8 threads × 50 iters of HmolarQ flashes on water). Build:

```
cmake -DCOOLPROP_CATCH_MODULE=ON \\
      -DCMAKE_CXX_FLAGS="-fsanitize=thread -g -O1" \\
      -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=thread" ..
```

**This PR's new code is TSan-clean.** No warnings inside `ensure_HS_under_lock`, `resolve_T_via_superancillary`, `alpha0_offset_total`, or `ensure_caloric_superancillaries`. The mutex and the always-lock policy hold up under contention.

**Three pre-existing races surfaced** (not introduced by this PR; the new test just exercises code paths that touch them):
1. `Configuration::_get_config()` at `Configuration.cpp:72` — already-known global singleton race.
2. Same `_get_config` race (different write site).
3. `static int deriv_counter` at `HelmholtzEOSMixtureBackend.cpp:41`, incremented non-atomically inside `calc_all_alphar_deriv_cache`. Plus a possible second race on `IdealHelmholtzContainer`'s evaluation cache that needs investigation.

Filed #2844 to track the third (the first two are already in internal triage). The bd / sibling issue is cross-linked.

Not blocking this PR — its own design is correct under TSan and the surrounding pre-existing races are independent infrastructure work.

PR #2838, opened 2026-04-29
Mass quality methods

kaern:
### Requirements

### Description of the Change

New QmassT_inputs and PQmass_inputs for the abstractstate class and refprop backend specifically. missing extention to other backends and high level interface etc.  

more info here: https://github.com/CoolProp/CoolProp/pull/2221

### Benefits

use mass based units for vapor quality, which seems to be the only missing parameter still in mole fractions

### Verification Process

inverted property calculations tests (see the addtional python test files - to be deleted)


include/AbstractState.h
```diff
@@ -109,6 +110,8 @@ class AbstractState
     /// Bulk values
     double _rhomolar, _T, _p, _Q;
 
+    CachedElement _Qmass;
```
ibell, line 113: this should follow the pattern with the other variables. Also will need to be cleared

include/DataStructures.h
```diff
@@ -277,11 +277,13 @@ enum fluid_types
...
 // !! If you add a parameter, update the map in the corresponding CPP file !!
 /// These are input pairs that can be used for the update function (in each pair, input keys are sorted alphabetically)
-enum input_pairs : int
+enum input_pairs
 {
     INPUT_PAIR_INVALID = 0,  // Default (invalid) value
     QT_INPUTS,               ///< Molar quality, Temperature in K
+    QmassT_INPUTS,           ///< Mass quality, Temperature in K
     PQ_INPUTS,               ///< Pressure in Pa, Molar quality
+    PQmass_INPUTS,           ///< Pressure in Pa, Mass quality
```
ibell, line 286: Also need all the other Qmass variants, or at least it would be nice

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
```diff
@@ -329,6 +329,9 @@ void REFPROPMixtureBackend::set_REFPROP_fluids(const std::vector<std::string>& f
         mole_fractions.resize(ncmax);
         mole_fractions_liq.resize(ncmax);
         mole_fractions_vap.resize(ncmax);
+        mass_fractions.resize(ncmax);
```
ibell, line 332: do we need the mass_fractions arrays at all? I think it is better to stick to mole fractions and construct mass fraction arrays as needed

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
```diff
@@ -935,31 +944,56 @@ CoolPropDbl REFPROPMixtureBackend::calc_gas_constant() {
...
 CoolPropDbl REFPROPMixtureBackend::calc_molar_mass(void) {
     this->check_loaded_fluid();
-    double wmm_kg_kmol = NAN;
+    double wmm_kg_kmol;
     WMOLdll(&(mole_fractions[0]), &wmm_kg_kmol);  // returns mole mass in kg/kmol
     _molar_mass = wmm_kg_kmol / 1000;             // kg/mol
     return static_cast<CoolPropDbl>(_molar_mass.pt());
 };
+CoolPropDbl REFPROPMixtureBackend::calc_Qmass(void) {
+    // Convert molar quality to mass quality using REFPROP's QMASS function
```
ibell, line 953: would be nice to make the function for Qmolar-> Qmass in the AbstractState base class rather than implementing it here (the function itself is very concise)

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
```diff
@@ -1061,12 +1095,22 @@ CoolPropDbl REFPROPMixtureBackend::calc_conductivity(void) {
...
+    double sigma, rho_mol_L = 0.001 * _rhomolar;
     int ierr = 0;
     char herr[255];
-    SURFTdll(&_T, &rho_mol_L, &(mole_fractions[0]),  // Inputs
-             &sigma,                                 // Outputs
-             &ierr, herr, errormessagelength);       // Error message
+
+    // Use STNdll if liquid and vapor densities are available, otherwise use SURFTdll
+    if (_rhoLmolar && _rhoVmolar) {
+        double rhoL_mol_L = 0.001 * static_cast<double>(_rhoLmolar), rhoV_mol_L = 0.001 * static_cast<double>(_rhoVmolar);
```
ibell, line 1104: out of scope for PR

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
```diff
@@ -1809,6 +1853,75 @@ void REFPROPMixtureBackend::update(CoolProp::input_pairs input_pair, double valu
             _rhoVmolar = rhoVmol_L * 1000;  // 1000 for conversion from mol/L to mol/m3
             break;
         }
+        case PQmass_INPUTS: {
+
+            // Unit conversion for REFPROP
```
ibell, line 1858: this block should only convert Qmass to Qmolar and then recursively call again, like when you have Smass inputs for instance

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
```diff
@@ -1871,6 +1984,76 @@ void REFPROPMixtureBackend::update(CoolProp::input_pairs input_pair, double valu
             _rhoVmolar = rhoVmol_L * 1000;  // 1000 for conversion from mol/L to mol/m3
             break;
         }
+        case QmassT_INPUTS: {
```
ibell, line 1987: same comment as above

src/AbstractState.cpp
```diff
@@ -195,6 +195,8 @@ bool AbstractState::clear() {
     this->_p = -_HUGE;
     this->_Q = -_HUGE;
 
+    this->_Qmass.clear();
```
ibell, line 198: if the other caching strategy was used, this clear would not be needed

ibell (changes requested):
There are a number of issues here. Do you want to work through them with my guidance? First thing would be to clean up the PR to have it only do things related to Qmass management

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
kaern: I think I added mass fractions for completeness, and for recalculating _Q on molar basis when Qmass was used as input. Not sure whether this step is necessary

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
kaern: This I did not understand. At least for refprop I think we need TQFLSH and PQFLSH functions, but maybe you are writing in more general terms..

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
kaern: Feel free to use it or not

ibell:
Thanks for kicking off this discussion and providing good reference code to build from

kaern:
Thanks a lot for providing this feature :-) I tested the cpp and it works at least the few tests i had, only thing missing is to expose Qmass method in the python wrapper as well .. but I guess you haven't reached this layer yet :-) when do you believe the next stable version of coolprop is released ? 

https://github.com/CoolProp/CoolProp/commit/2a72d8bfbfa405f40b43be59d77e6502d5bebd9a#diff-86b0bfc95bfd7f9b3121217daacfca49767756449ee17d0a197aff9283269b78

ibell:
Sorry I missed the python interface bit in my implementation. I'll add it in a follow on PR

ibell:
@kaern  fixed: https://github.com/CoolProp/CoolProp/pull/2889

PR #2892, opened 2026-05-06
fix(BICUBIC): pick unit-interval root in invert_single_phase_x/y (#1301)

### LINKED ISSUE #1301: BICUBIC near Saturation errors for CO2
Dear all,

I would like to generate a rho-T table to compute pressure like P=P(rho,T). With Bicubic method, I obtained some large pressure errors  near saturation curve.

I tried P-H input for density error, it's a little better.

Here is the diagrammes that i obtained and the python code file is filed.

How can I improve this ?

Thanks,

Yu

co2_errordt
co2_errorph

`Ref = 'CO2'

BICUBIC = CoolProp.AbstractState('BICUBIC&HEOS',Ref)
#TTSE = CoolProp.AbstractState('TTSE&HEOS',Ref)

Dmax = 1200
Dmin = 1e-2
Tmax = 500
Tmin = 220
flag = 'DT'

EOS = CoolProp.AbstractState('HEOS',Ref)
TTT1, DDD1, EEE1 = [], [], []
HHH2, PPP2, EEE2 = [], [], []

if flag == 'DT' :

	for a_useless_counter in range(80000):

    		D = random.uniform(Dmin,Dmax)
    		T = random.uniform(Tmin,Tmax)
    		CP.set_debug_level(0)
    		try:

        	    EOS.update(CoolProp.DmassT_INPUTS, D, T)
        	    P_EOS = EOS.p()   #; TEOS = EOS.T()

                    BICUBIC.update(CoolProp.DmassT_INPUTS, D, T)
        	    P_BICUBIC = BICUBIC.p() #; TBICUBIC = BICUBIC.T()

#        errorBICUBIC = (abs(P_BICUBIC)-abs(P_EOS))/P_EOS*100
        	    errorBICUBIC = abs(P_BICUBIC-P_EOS)/P_EOS*100
        	    if errorBICUBIC > 100  :
            	       print D, T, errorBICUBIC, P_BICUBIC, P_EOS

        	    TTT1.append(T)
                    DDD1.append(D)
                    EEE1.append(errorBICUBIC)

    		except ValueError as VE:
        		print 'ERROR', VE
        		pass
`

ibell:
## Summary

Fixes #1301 — BICUBIC `P(D, T)` for CO₂ at supercritical T and low ρ returned absurd values (e.g. **P = −760 MPa for T=315 K, ρ=1 kg/m³** — wrong by six orders of magnitude *and* wrong sign).

Two distinct bugs conspired to produce the symptom; this PR fixes both.

### Bug A — root selection

`BicubicBackend::invert_single_phase_x/y` solved the cell-local cubic `a·r³ + b·r² + c·r + d = target` and historically picked the **smallest-absolute-value** real root. That can be a far-negative root with no physical meaning. Once unscaled by the cell width, it became a wildly wrong (often negative) pressure or enthalpy.

### Bug B — cell selection (the root cause of the −760 MPa symptom)

`SinglePhaseGriddedTableData::find_nearest_neighbor` bisected the `otherkey` column at the **lower-axis edge** of the cell (xkey branch: `bisect_vector(v[i], otherval, j)`). When the orthogonal-axis dependence is significant — e.g. for the CO₂ PT-table the density along a supercritical isotherm differs meaningfully from the lower-T edge of the cell — the bisection systematically lands several cells off. The cubic in that wrong cell has no root in [0, 1]; the historical code happily extrapolated, sometimes giving a usable answer by luck, sometimes (for genuinely out-of-table states) a catastrophic `|yhat| ~ 50` → `~-750 MPa`.

## Fix

1. **`src/Backends/Tabular/TabularBackends.h` (root cause of Bug B).** `find_nearest_neighbor`'s xkey branch now bilinearly interpolates the `otherkey` column at the actual query position `xhat ∈ [0, 1]` inside cell `(i, .)` before bisecting, instead of slicing at the lower-x edge `v[i]`. This gives the bisection a row of `otherkey` values close to what the cubic interpolant will actually see, eliminating the systematic off-by-one. The ykey branch (PH path via `bisect_segmented_vector_slice`) has the same defect; filed as `CoolProp-bf4` out of scope.

2. **`src/Backends/Tabular/BicubicBackend.cpp` (defence in depth for Bug A and Bug B residual).** Replace `pick_unit_interval_root`'s "closest-to-[0, 1] fallback" with `find_in_cell_root`, which returns `std::optional<double>` — only succeeds if a root lies in `[-1e-6, 1+1e-6]`. When multiple in-cell roots exist (e.g. near a saturation curve where the interpolant is non-monotonic), prefers the one closest to 0.5.

3. **Walk to the immediate neighbour.** When the bisection-chosen cell still misses (the bilinear is a linear approximation of the bicubic, so the true root can sit in the neighbour), `invert_single_phase_x` tries `i±1` and `invert_single_phase_y` tries `j±1`. Each candidate is checked for `cell.valid()` and uses its own real interpolant. The walked cell is written back to `cached_single_phase_i/j`. Multi-cell walks are intentionally disallowed — the immediate neighbour is the only physically meaningful source.

4. **Throw on no in-cell root in initial or neighbour.** `ValueError` with rich diagnostic info (cell coordinates, the `other_key/other_val` pair, the bounding `xvec/yvec`). Pre-fix this produced silent garbage; post-fix the caller can detect the out-of-table state.

## Validation

Reproduced yufang67's original `(T, rho)` sweep for CO₂ (T=220–500 K, ρ=0.01–1200 kg/m³), matching their color range. On master the wrong-cell-pick error pattern they reported (red/yellow rectangular bands) is reproduced exactly. After this PR:

- In-table p99 relative error: **33 % → 8.8 %**
- The rectangular-band wrong-cell error pattern is gone
- yufang67's worst-case point at T=315 K, ρ=1 kg/m³ (the original −760 MPa report) lies cleanly in the out-of-table region where the fixed code throws a `ValueError`

## Test

`src/Tests/CoolProp-Tests.cpp [BICUBIC][1301]` — split into two sections:

- **in-table supercritical CO₂ P(D, T)** — BICUBIC agrees with HEOS to within 1 % at (T, ρ) pairs across 320–500 K, 20–100 kg/m³.
- **out-of-table low-density CO₂ throws rather than returning garbage** — `CHECK_THROWS_AS(ValueError)` for `(ρ=1 kg/m³, T=320 K)` and `(ρ=5 kg/m³, T=315 K)`, both well below the table's pmin at that T.

Full local Catch2 suite: **279 / 279 active, 18 skipped, 0 failures, 64,817 assertions.**

## Summary by CodeRabbit

* **Bug Fixes**
  * Bicubic table inversion now restricts root selection to roots inside the candidate cell’s normalized domain and uses a bounded neighbor retry to correct off-by-one cell choices.
  * Nearest-neighbor lookup now accounts for in-cell interpolation to avoid off-by-one neighbor selection.
  * States outside the table’s valid range now raise errors instead of producing invalid results.

* **Tests**
  * Added regression test covering single-phase bicubic inversion accuracy and out-of-table error detection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CoolProp/CoolProp/pull/2892?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/Backends/Tabular/BicubicBackend.cpp
+++ b/src/Backends/Tabular/BicubicBackend.cpp
@@ -3,9 +3,69 @@
 #    include "BicubicBackend.h"
 
 #    include <cmath>
+#    include <limits>
+#    include <optional>
 #    include "MatrixMath.h"
 #    include "DataStructures.h"
 #    include "Backends/Helmholtz/PhaseEnvelopeRoutines.h"
+
+namespace {
+// Return the cubic root that lies in [0, 1] (the cell's normalized
+// coordinate), or std::nullopt if no root is in-cell.
+//
+// solve_cubic returns up to three real roots in arbitrary order. The
+// historical code picked the smallest-absolute-value root, which can be
+// a far-negative root with no physical meaning and produces wildly wrong
+// values once unscaled to the cell width (#1301: P = -760 MPa for CO2 at
+// T=315K, rho=1 kg/m^3).
+//
+// Strict cell bound here is intentional: the bicubic interpolant has
+// physical meaning only inside the cell. If the bisection landed in the
+// wrong cell, the caller walks to a neighbour and retries; if no
+// neighbouring cell contains the root either, the requested state is
+// outside the table's domain and the caller throws.
+//
+// When multiple roots are in [0, 1] (can happen for cells near a
+// saturation curve where the interpolant is non-monotonic), prefer the
+// one closest to 0.5 — that's furthest from the cell edges and least
+// sensitive to coefficient noise.
+constexpr double kCellRootTol = 1e-6;
+
+inline std::optional<double> find_in_cell_root(int N, double r0, double r1, double r2) {
+    const double roots[3] = {r0, r1, r2};
+    int best = -1;
+    double best_score = std::numeric_limits<double>::infinity();
+    for (int k = 0; k < N; ++k) {
+        const double r = roots[k];
+        if (r >= -kCellRootTol && r <= 1.0 + kCellRootTol) {
+            const double score = std::abs(r - 0.5);
+            if (score < best_score) {
+                best_score = score;
+                best = k;
+            }
+        }
+    }
+    if (best < 0) {
+        return std::nullopt;
+    }
+    return roots[best];
+}
+
+// Maximum number of cells to walk away from the bisection's initial
+// guess.  find_nearest_neighbor (TabularBackends.h) now bilinearly
+// interpolates the otherkey column at the actual query position before
+// bisecting (xkey branch), so the chosen cell is usually correct.  The
+// remaining off-by-one cases come from the bilinear estimator being a
+// linear approximation of the bicubic interpolant — the cubic in the
+// neighbouring cell can hold the true root.  The immediate neighbour
+// (one cell either side) is the only physically meaningful source:
+// that's a real cell with its own valid cubic, not an extrapolation of
+// the original cell's interpolant.  Walking further would just be
+// extrapolation in disguise and would mask genuinely out-of-table
+// queries — those are caught here by throwing.  #1301's catastrophic
+// case lands ~50+ cells away, well outside this bound.
+constexpr int kMaxCellWalk = 1;
+}  // namespace
 
 void CoolProp::BicubicBackend::find_native_nearest_good_indices(SinglePhaseGriddedTableData& table,
                                                                 const std::vector<std::vector<CellCoeffs>>& coeffs, double x, double y,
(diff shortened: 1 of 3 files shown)


ibell:
Closing without merging per maintainer feedback.

The two attempts in this PR — (a) silently masking with abs-min-root fallback (master), and (b) strict in-cell root + throw on out-of-table (this PR) — both fail: (a) returns wildly wrong pressures, (b) breaks any caller that was working (even imperfectly) with master's behavior.

The right architectural fix is to fall through to HEOS in `TabularBackend::update()` when the BICUBIC inverter can't resolve the state — `BICUBIC&HEOS` has HEOS as the source of truth, the speedup is incidental. That's a larger change than is in scope here, so leaving #1301 open for it.

The bilinear-bisection improvement to `find_nearest_neighbor` (TabularBackends.h) in this PR is genuinely useful on its own (in-table p99 error 33 % → 8.8 % on yufang67's grid) and could be revived as a separate PR if there's appetite.

Filed `CoolProp-bf4` for the upstream bisection-using-lower-edge-column root cause, and a new follow-up bd issue for the HEOS-fallback architecture.

Issue #2906, opened 2026-05-11
HAPropsSI fails at T_wb = 0 °C across all RH (sub-freezing branch, #2690 follow-up)


ibell:
Spun off from #2690 after the Modes B + C fix landed (PR forthcoming).

## Summary

When sweeping `HAPropsSI('T_db', 'T_wb', T_wb, 'RelHum', RH, 'P', P)` over the reporter's `td.xlsx` grid (RH ∈ {0.01, 1e-3, 1e-4, 1e-5}, T_wb ∈ {0, 5, …, 65 °C}, P ∈ {87000, …, 98950 Pa}) on master + the #2690 fix, every failure above freezing is either eliminated or surfaces as the explicit "beyond the validity range" infeasibility error. **The remaining residue is concentrated entirely at T_wb = 0 °C, scattered across all four RHs, ~28–37 isolated P points each.**

```
T_wb=0, RH=0.01   : 28 scattered failures in [88000, 98850]
T_wb=0, RH=1e-3   : 23 scattered failures in [92150, 98800]
T_wb=0, RH=1e-4   : 37 scattered failures in [88600, 98900]
T_wb=0, RH=1e-5   : 36 scattered failures in [90400, 98650]
```

The pattern is **chaotic / non-contiguous**, in contrast to the smooth bands of Modes A/B/C — strongly suggesting a different mechanism than the wet-bulb iteration's near-Tsat singularity that #2690 addressed.

## Suspected mechanism

T_wb = 273.15 K is essentially at IF97's lower validity limit and right on the **liquid water → ice** branch boundary. `WetBulbSolver::call` (`src/HumidAirProp.cpp:1306`) switches between IF97 (`Twb > 273.16`) and the ice EOS path (`psub_Ice` / `h_Ice`) at exactly that line:

```cpp
if (Twb > 273.16) {
    p_ws_wb = IF97::psat97(Twb);
    ...
    h_w = Water->keyed_output(CoolProp::iHmass);
} else {
    p_ws_wb = psub_Ice(Twb);
    h_w = h_Ice(Twb, _p);
}
```

When the inner Brent steps cross 273.16 K from above, the residual has a discontinuity (different `p_ws_wb` formula → different `W_s_wb` → different `h_w`). Brent doesn't tolerate discontinuities cleanly. The scatter pattern across P at fixed T_wb=0 °C is consistent with Brent landing on either side of 273.16 K depending on the iteration trajectory.

## Reproducer

```python
[...]

Issue #2690, opened 2026-03-17
Humid air Fails peculiarly. Temperature returned as (inf) for a small range of pressures.


chandradharkoneti:
Config:
Python - Ver - 3.9.5 (Can't upgrade due to legacy dependency issues)
CoolProp - Ver - '6.6.0' (I'm afraid of upgrading lest something else breaks)

Hi, CoolProp team.

I am using CoolProp to generate a psychrometric chart. It was working flawlessly until today. 
The line `cp.HAPropsSI('T_db', "T_wb", 30+273.15, "RelHum", 0.00001, "P", 91024) - 273.15` fails at pressures of ~91000 Pa. 

Error: `ValueError: The temperature value (inf) is outside the range of validity: (10) to (1e+07) :: inputs were:"T_db","T_wb",3.0314999999999998e+02,"RelHum",9.9999999999999995e-07,"P",9.1024687500000000e+04`

It works with 90000, 89000, 85000, and also at 92000, 95000, 101325, etc. It also works with 91200, 91500, 

It fails ONLY with a few values. 91000, __*91024*__, 91100, etc.

So I did:
```
for i in range(87000, 99000, 100):
    try:
        abc = cp.HAPropsSI('T_db', "T_wb", 30+273.15, "RelHum", 0.00001, "P", i) - 273.15
    except:
        print (i, end = ", ")
```
and got the output as `90200, 90300, 90400, 90500, 90600, 90700, 90800, 90900, 91000, 91100`

Why the peculiar range where it is failing??

I also reduced the step-size (to 10) and saw, it is continuously failing from 90190 to 91160 continuously. 

It also fails at 97950 for some reason. 97950 is an outlier of this range.

chandradharkoneti:
Hi @ibell I'm sorry to say I didn't understand what you meant in the commit comments. I am not too familiar with the inner mechanisms of coolprop. I am only familiar with the Python interface. All that abstractstate, brent, etc,, I'm not familiar with. The only brent I know of is Brent Crude (Crude oils from the European north sea).

I have however created a spreadsheet of the pressures, temperatures and RH values at which the routines are failing. If it can give you any additional insights into WHY / WHAT / WHERE the routines are failing, I'd be glad. If not, just ignore it :)

```
td ={}

for i in [0.01, 0.001, 0.0001, 0.00001]:
[...]

Issue #2255, opened 2023-05-18
HAPropsSI("Twb","T",200+273.15,"W",0.2,"P",1000E3) gives wrong result


frank998:
### Description

It seems that CoolProp gives wrong answer for calculating moist air wet bulb temperature at 1000kPa and dry bulb temperature of 200C. It can also seen from the output of the validation file 'fluid_properties/Validation/HAValidation.py' that a discontinuty point exists.
![image](https://github.com/CoolProp/CoolProp/assets/35061671/54bd5926-7bc4-4579-8c3e-be21c7f37833)

Besides, directly using the expression below, the correct value should be around 131C (WB). With T=200, Twb=180.72, it gives value -31.84, while with T=200, Twb=131, it gives value 0.197.
![image](https://github.com/CoolProp/CoolProp/assets/35061671/0876fb1c-8a19-4b52-b5a4-4df1744a8886)

### Steps to Reproduce
Invoking syntax gives the same answer: 180.7241

``` matlab
Twb = py.CoolProp.CoolProp.HAPropsSI("B","W",0.2,"T",200+273.15,"P",1000E3)-273.15
```

By changing the air states (dry-bulb temperature, humidity ratio, pressure) around the point mentioned above, the discontinuty are all observed, as seen below,
I am not sure if it is really a bug/wrong. But I felt confused about this discontinuty.

``` matlab
W = 0.05:0.05:1;
Twb = zeros(size(W));
for i=1:length(W)
Twb(i) = py.CoolProp.CoolProp.HAPropsSI("B","W",W(i),"T",200+273.15,"P",1000E3)-273.15;
end

figure
plot(W,Twb,'Marker','o','LineWidth',1.5)
xlabel('Humidity ratio W (kg/kga)')
ylabel('Wet bult temperature (\circC)')
```
![image](https://github.com/CoolProp/CoolProp/assets/35061671/afb80188-4f59-4ac6-9eba-74f899360618)

``` matlab
T = 190:0.1:210;
Twb = zeros(size(T));
for i=1:length(T)
Twb(i) = py.CoolProp.CoolProp.HAPropsSI("B","W",0.2,"T",T(i)+273.15,"P",1000E3)-273.15;
end

figure
plot(T,Twb,'LineWidth',1.5)
xlabel('Dry bulb temperature (\circC)')
ylabel('Wet bult temperature (\circC)')
```
[...]

PR #2907, opened 2026-05-11
fix(HumidAir): T_db from (T_wb, RelHum, P) at low RH (#2690 modes B+C)

ibell:
## Summary

Closes Modes B and C of the residual failure surface in #2690 (after b3360b8c6's #2255 fix incidentally cleared Mode A). Sub-freezing Mode D split out to #2906.

- **Mode B** (Tsat(p)-crossing isolated outliers, e.g. `T_wb=28, RH=1e-4, P=87550`): tighten `WetbulbTemperature`'s inner Brent `Tupper` to `std::min(Tmax+1, Tsat-0.1)` (one expression now covers both #2255's `T >= Tsat` regime and #2690's `T just below Tsat` regime), and pre-check IF97's saturation tolerance in `isothermal_compressibility` so `f_factor` doesn't trip `WaterIF97->update(PT_INPUTS, …)` when the outer `Brent_HAProps_T` iterate brushes Tsat.
- **Mode C** (high `T_wb` + low RH where the wet-bulb energy balance implies `T_db > 500 K`): in `_HAPropsSI_inputs`'s `(RelHum, T_wb)` branch, estimate `T_db` from the energy balance and either (a) widen `T_max` to `T_db_est + 20 K` so Brent brackets directly, or (b) throw a clear `ValueError` naming `T_wb / RelHum / P / T_db_estimate` instead of letting the iterative solver emit `inf`. Gated on `(RelHum, T_wb)` + `RH < 1e-3` to stay clear of the high-pressure regression that got #2697 reverted in #2704.

The `catch (...)` I prototyped in `isothermal_compressibility` was replaced with an explicit pre-check (`|p - psat(T)| <= psat * 3.3e-5`) — same condition `IF97Backend::set_phase` uses, no exception-based control flow.

## Test plan

- [x] New tests under `[humid_air][2690]`:
    - Mode B isolated points (`T=30/RH=1e-5/P=97950`, `T=28/RH=1e-4/P=87550`)
    - Mode A pressure-band cases (`P ∈ {90200, 90550, 90700, 91000, 91100, 91024}`) — pin against b3360b8c6 + the new Tupper clamp
    - Mode C infeasibility error-string check (must contain "beyond the validity range")
    - ASHRAE-style high-P Twb regression guard against repeating the #2697 over-correction
    - Abbreviated replay of the reporter's `td.xlsx` sweep asserting `unexpected_failures_above_freezing == 0`
- [x] `./CatchTestRunner "[humid_air],[HAPropsSI],[2255],[2690],[2670],[2195],[water_flash],[virial_cache],[alpha0_cache]"` → 200 / 200 assertions in 12 cases
- [ ] CI: full Catch suite + clang-format pre-commit

## Out of scope

Sub-freezing `T_wb=0 °C` scatter (Mode D, ~124 points across all RH) is a separate `WetBulbSolver` discontinuity at the 273.16 K boundary between the IF97 and ice EOS branches — tracked in #2906.

### DIFF
--- a/src/HumidAirProp.cpp
+++ b/src/HumidAirProp.cpp
@@ -797,10 +797,24 @@ double isothermal_compressibility(double T, double p) {
             k_T = 1.6261876614E-22 * pow(T, 6) - 3.3016385196E-19 * pow(T, 5) + 2.7978984577E-16 * pow(T, 4) - 1.2672392901E-13 * pow(T, 3)
                   + 3.2382864853E-11 * pow(T, 2) - 4.4318979503E-09 * T + 2.5455947289E-07;
         } else {
-            // Use IF97 to do the P,T call
-            WaterIF97->update(CoolProp::PT_INPUTS, p, T);
-            Water->update(CoolProp::DmassT_INPUTS, WaterIF97->rhomass(), T);
-            k_T = Water->keyed_output(CoolProp::iisothermal_compressibility);
+            // IF97Backend::set_phase rejects PT_INPUTS whenever |p - psat(T)|
+            // is within IF97's 3.3e-5 RMS saturation-line tolerance — the
+            // pure-water phase is ambiguous there. Pre-check that same
+            // condition and route the near-saturation case through the
+            // polynomial correlation instead; for f_factor purposes the
+            // precise k_T at saturation does not matter much, and a
+            // single-valued surrogate keeps the outer (T_wb,RelHum) solver
+            // from blowing up when its T_db iterate brushes Tsat(p). (#2690)
+            constexpr double if97_sat_tol = 3.3e-5;
+            const double p_ws_T = IF97::psat97(T);
+            if (std::abs(p - p_ws_T) <= p_ws_T * if97_sat_tol) {
+                k_T = 1.6261876614E-22 * pow(T, 6) - 3.3016385196E-19 * pow(T, 5) + 2.7978984577E-16 * pow(T, 4) - 1.2672392901E-13 * pow(T, 3)
+                      + 3.2382864853E-11 * pow(T, 2) - 4.4318979503E-09 * T + 2.5455947289E-07;
+            } else {
+                WaterIF97->update(CoolProp::PT_INPUTS, p, T);
+                Water->update(CoolProp::DmassT_INPUTS, WaterIF97->rhomass(), T);
+                k_T = Water->keyed_output(CoolProp::iisothermal_compressibility);
+            }
         }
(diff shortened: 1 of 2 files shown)


src/HumidAirProp.cpp
```diff
@@ -1409,8 +1431,11 @@ double WetbulbTemperature(double T, double p, double psi_w) {
         }
         // Reject solutions that are essentially the upper bracket — that
         // is Brent giving up rather than finding a true root in the
-        // physical region (#2255).
-        if (T >= Tsat && return_val > Tsat - 1e-2) {
+        // physical region (#2255). The threshold tracks the Tupper
+        // margin above (0.1 K below Tsat); allow a small additional
+        // tolerance to distinguish "found the bracket" from "found
+        // close to the bracket".
+        if (T >= Tsat && return_val > Tsat - 0.11) {
```
ibell: why 0.11?

src/HumidAirProp.cpp
```diff
@@ -1392,12 +1406,20 @@ double WetbulbTemperature(double T, double p, double psi_w) {
...
+    // Brent steps still land on the singular point (#2690 isolated
+    // outliers like T_wb=28, RH=1e-4, P=87550). Clamp Tupper strictly
+    // below Tsat with a margin large enough to clear IF97's saturation-
+    // tolerance band — WaterIF97->update(PT_INPUTS, p, T) throws when
+    // |psat(T) - p|/p falls below ~3e-3 %, i.e. roughly 1 mK below Tsat
+    // for atmospheric P. Use 0.1 K to give that check several orders of
+    // headroom while preserving plenty of bracket for the true Twb,
+    // which is typically tens of K below Tsat for the wet-bulb cases
+    // we care about.
+    double Tupper = std::min(Tmax + 1.0, Tsat - 0.1);
```
ibell, line 1422: why 0.11?

ibell: Fair — replaced with `Tupper - 1e-2` in 7a007f76e so the rejection threshold is derived from the bracket rather than encoded as an independent literal. The 0.1 K margin is now a single source of truth.

src/HumidAirProp.cpp
ibell: The 0.11 was 0.1 (Tupper margin) + 0.01 K tolerance, but written as an unlinked literal. Fixed in 7a007f76e by deriving the threshold from `Tupper` directly: `return_val > Tupper - 1e-2`.

PR #2928, opened 2026-05-16
SVDSBTL opts in to factory-string options (PR B — schema + cache key)

ibell:
## Summary

Second of five PRs landing the per-instance backend-options mechanism. **Stacked on top of #2927 (PR A — infrastructure)** — review that first.

### What lands

- **`include/CoolProp/schemas/SVDSBTLOptions.h`** — JSON Schema as a compile-time string literal (`kSVDSBTLOptionsSchemaJson`). Strict mode (`additionalProperties:false` everywhere). Top-level: `schema` (versioned), `grid` (NT/NR/rank), `properties` (transport), `critical_patch` (mode/source/tolerance/metric/bbox).
- **SVDSBTL constructor** gains a third arg `options_json` — parses + validates against the schema, threads `grid.{NT,NR,rank}` into the surface preset builders. `build_options_json()` override returns the canonical form (sorted keys, defaults expanded).
- **`include/CoolProp/Hash.h`** — inline FNV-1a 64 + 16-hex helper. Same hash family CoolProp already uses for `source_eos_hash` superancillary freshness stamping (no new dependencies).
- **`SVDSurfaceSerializer::default_cache_path`** — now emits `<fluid>.<source>.<input_pair_name>.<opthash>.svd.bin.z`:
  * Symbolic input-pair name (e.g. `PT_INPUTS`) closes **CoolProp-b6v** — reordering `input_pairs` enum can no longer silently misalign caches.
  * 16-hex FNV-1a 64 opthash over canonical options JSON — two callers with the same logical options share one cache file, different options get different files.
  * `kRevision` bumped 3 → 4.
- **`AbstractState::factory()` SVDSBTL inline branch** forwards the raw options string verbatim instead of throwing.

`critical_patch` knobs are accepted by the schema and persisted in the canonical-options blob, but the **routing** (PR C) and **auto-calibration** (PR D) land in follow-up PRs.

### Test plan

- [x] Schema validation (10 cases) — unknown top-level/nested keys, type mismatches, enum violations, missing required keys, bbox arity, malformed JSON, empty + fully-populated valid options.
- [x] `build_options_json()` round-trip: key-order-permuted inputs collide on the same canonical form; canonical form round-trips through `factory()` byte-for-byte.
- [x] Cache filename format — symbolic input-pair name present; old raw int absent; uppercase/path-traversal opthashes rejected.
- [x] FNV-1a 64 known-vector check.
[...]

PR #2927, opened 2026-05-16
Generic backend options via factory-string JSON (PR A — infrastructure)

ibell:
## Summary

First of five PRs landing the per-instance backend-options mechanism described in
[docs/superpowers/specs/2026-05-16-backend-options-string-design.md](https://github.com/CoolProp/CoolProp/blob/ihb/backend-options-spec/docs/superpowers/specs/2026-05-16-backend-options-string-design.md).

This PR is **pure infrastructure** — no backend opts in yet, no behaviour change in any existing call site.  Subsequent PRs (B–E in the breakdown below) consume the infrastructure: SVDSBTL first, BICUBIC/TTSE/REFPROP later, deprecating the corresponding Configuration globals along the way.

### What lands

- **Factory-string parser** (`include/CoolProp/FactoryOptions.h`, `src/FactoryOptions.cpp`).  Splits the optional ``?<options>`` suffix from a factory string with a single ``find_first_of('?')``; resolves inline JSON or ``?@<path>`` indirection.  Pure helper; 27 assertions over 7 cases including all the ``?``-inside-JSON contracts Ian called out at design review.
- **JSON schema validator + canonical-JSON helper** (`include/CoolProp/SchemaValidation.h`, `src/SchemaValidation.cpp`).  RapidJSON-backed strict-mode validator; recursive key-sort canonicaliser used for cache hashing and reproducibility (FNV-1a 64 over the canonical bytes; same hash family CoolProp already uses for ``source_eos_hash``).
- **Options-aware `AbstractStateGenerator` overload** + ``AbstractState::build_options_json()`` virtual.  Default forwards to the no-options overload on empty input and throws ``NotImplementedError`` otherwise, so silent option-dropping cannot happen.  Cython binding stub added.
- **`AbstractState::factory()` wired** to call the parser on entry, hoist a ``?<options>`` suffix from *either* the backend or the fluid side (PropsSI puts it on the fluid side after ``::`` splitting), and pass the raw JSON through to the backend generator.  Existing TTSE / BICUBIC / SVDSBTL inline branches throw if options are supplied — they opt in via follow-up PRs.
- **PropsSI / Props1SI / PhaseSI pass-through tests** — the high-level API forwards ``?<options>`` verbatim through the existing ``FluidName`` argument; zero new entry-points; works for FORTRAN / Excel / MATLAB via the standard string path.
- **Sphinx docs page** (`Web/coolprop/BackendOptions.rst`).  Linked into the toctree between Tabular and Configuration.

### Why this layout

[...]

PR #2929, opened 2026-05-17
SVDSBTL critical-patch HEOS-fallback routing (PR C)

ibell:
## Summary

Third PR in the backend-options epic.  **Stacked on #2928 (PR B) → #2927 (PR A) → master.**

### What lands

Activates the `critical_patch.*` knobs PR B already accepted in the schema and persisted in the canonical-options blob.  When the active state lands inside the per-input-pair bbox, every `calc_*` forwards to a source backend instead of the SVD surface, so the rank-truncation residual near the critical singularity goes away.

- **`SVDSBTLBackend::CriticalPatch`** holds an `enabled` flag, an optional source-backend override, and a `bbox_per_pair` map (one entry per supported input pair).  `contains()` is an O(1) axis-aligned containment check.
- **`build_critical_patch_()`** runs once at construction.  PT_INPUTS bbox is native `(p, T)`.  HmassP_INPUTS bbox is derived by walking the `(T, p)` bbox perimeter through `source->hmass()`; the resulting `(p, h)` rectangle is conservative — every `(T, p)` inside the critical box maps to an `(h, p)` inside the patch envelope, with small false-positive headroom that just means the source backend answers a few extra calls outside the strict box.
- **`update()`** flips `active_in_patch_` once per call; **`calc_rhomass / calc_hmass / calc_smass / calc_umass / calc_T / calc_speed_sound`** check the flag and forward to `patch_source_` when set.
- **`patch_source_ref_()`** lazy-allocates the patch backend so SVDSBTL&IF97 doesn't pay for a HEOS factory call unless the user opts into a HEOS-source patch via `critical_patch.source`.

### Defaults this PR ships

- `mode="auto"` (when not specified in options) — uses the spec's hardcoded multipliers `[0.95, 1.05]·Tcrit × [0.75, 1.15]·pcrit`.  PR D replaces with the binary-search-shrink auto-cal loop.
- `mode="off"` — patch disabled.
- `mode="fixed"` — honours the supplied `bbox`.
- `source=null` (default) — patch uses the SVD truth source.

### Test plan

5 new cases / 12 assertions in `CoolProp-Tests-SVDSBTLCriticalPatch.cpp`:

- Inside-bbox PT_INPUTS query is bit-identical to HEOS (patch fired).
- Outside-bbox PT_INPUTS query uses SVD (NOT bit-identical to HEOS — carries SVD truncation; explicit check that the patch did not silently fire).
- `mode=off` forces SVD even inside the would-be patch.
- Inside-bbox HmassP_INPUTS query is bit-identical to HEOS (the `(h, p) → (T, p)` hoist on the patch source works end-to-end).
- `mode=fixed` with a tight bbox: fires inside, NOT outside.

Full regression: **325 SVDSBTL assertions across 28 cases pass**.

### Beads closed

- **CoolProp-p0p** — critical-patch HEOS-fallback routing.

## Summary by CodeRabbit

* **New Features**
  * SVDSBTL backend adds a configurable "critical-patch" fallback that routes queries inside per-fluid bounding boxes to an alternate backend for improved accuracy; supports modes (on/off/fixed), optional explicit bbox, and handles PT, HmassP and HmolarP query types with lazy allocation of the patch backend.

* **Tests**
  * Added tests validating routing, accuracy inside/outside bboxes, HmolarP/HmassP handling, mode toggles, and construction-time validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CoolProp/CoolProp/pull/2929?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2110,6 +2110,8 @@ if(COOLPROP_CATCH_MODULE)
        "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests-PropsSIOptions.cpp")
   list(APPEND APP_SOURCES
        "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests-SVDSBTLOptions.cpp")
+  list(APPEND APP_SOURCES
+       "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests-SVDSBTLCriticalPatch.cpp")
 
   # CATCH TEST, compile everything with catch and set test entry point
   add_executable(CatchTestRunner ${APP_SOURCES})
--- a/include/CoolProp/Backends/SVDSBTL/SVDSBTLBackend.h
+++ b/include/CoolProp/Backends/SVDSBTL/SVDSBTLBackend.h
@@ -268,6 +268,45 @@ class SVDSBTLBackend : public AbstractState
         double uV_mol = 0.0;
     };
     TwoPhaseState two_phase_{};
+
+    // Critical-patch HEOS-fallback.  When an active query point lands
+    // inside the configured bbox (in whatever the input pair's native
+    // coords are), update() forwards the call to `patch_source_`
+    // instead of the SVD surface, and every calc_* below routes there
+    // for the duration of the current state.  Bbox is precomputed at
+    // construction per input pair so update() is a constant-time
+    // axis-aligned containment check.  Disabled when
+    // critical_patch.mode is "off" in the options blob.
+    struct CriticalPatchBbox
+    {
+        double a_lo, a_hi;  // primary axis: p for both PT and HmassP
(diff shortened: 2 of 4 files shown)


ibell:
@coderabbitai both legit, fixed in 88dd9649c:

**(1) bbox sampled from wrong backend** — `build_critical_patch_` now materialises the patch backend up-front via `patch_source_ref_()` and samples h(T,p) from it. An invalid `critical_patch.source` (e.g. IF97 patch on a non-Water fluid — IF97 is Water-only) fails at construction now instead of on the first in-patch query.

**(2) HmolarP unreachable** — Surface lookup normalises `HmolarP_INPUTS` → `HmassP_INPUTS` before `surfaces_.find()`, so the HmolarP case in the switch is reachable again and the molar→mass conversion runs as intended.

Added two regression sections in CoolProp-Tests-SVDSBTLCriticalPatch.cpp: 16/7 assertions/cases green.

PR #2936, opened 2026-05-20
test(svdsbtl): fix cache_present_for to match actual cache filenames

ibell:
## Summary

Two SVDSBTL test cases (\`factory\` and \`cache reload\`) have been emitting:

> SVDSBTL Water cache not present after construction -- in-memory only (likely read-only home dir)

on every CI run since Phase 2c landed.  This made it look like the SVD cache write was broken in CI sandboxes.

It isn't.  The \`cache_present_for()\` helper was calling \`SVDSurfaceSerializer::default_cache_path(fluid, source, pair)\` with the default \`opthash=\"no_opts\"\` and looking for \`Water.HEOS.HmassP_INPUTS.no_opts.svd.bin.z\`.  But \`SVDSBTLBackend::ensure_surface_\` computes \`opthash = to_hex16(fnv1a_64(\"{}\"))\` = \`08f44b07b5901a25\` for the no-options case and writes to \`Water.HEOS.HmassP_INPUTS.08f44b07b5901a25.svd.bin.z\` — a different filename.  The cache was always being written; the helper was always looking in the wrong place.

This PR globs the cache dir for \`<fluid>.<source>.<input_pair_name>.*.svd.bin.z\` so the helper finds the file regardless of which opthash the backend computes.

Verified locally: the factory test now passes with 4 assertions and no WARN, and a second cold-cache-build → warm-cache-load shows the expected 5× speedup (294 s rebuild → 55 s load).

## Background — CI duration

Side-note for context: the SVDSBTL [slow] tests genuinely do dominate the test_catch2 wall-clock (~10 minutes vs ~35 s pre-#2919).  That's because GHA doesn't persist \`~/.CoolProp/SVDTables\` between workflow runs, so every CI rebuilds every fluid's HmassP + PT surface from scratch.  Within a single CI run the cache _does_ work — second-and-subsequent same-fluid tests load in ~1 s.  Worth filing follow-up to either tag those tests so they can be skipped in PR CI, or to wire up actions/cache for the SVDTables dir.

## Test plan

- [x] \`CatchTestRunner \"SVDSBTL backend constructs via AbstractState::factory\"\` passes with 4 assertions (no WARN)
- [x] Cold-cache → warm-cache speedup: 294 s → 55 s
- [ ] Full \`[SVDSBTL]\` suite — CI

## Summary by CodeRabbit

* **Tests**
  * Improved test helper for cache detection: it now scans the cache directory and matches expected cache name patterns (based on test inputs) and a specific file suffix, instead of relying on exact filename checks—making cache presence checks more reliable.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CoolProp/CoolProp/pull/2936?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ibell:
Thanks @coderabbitai — addressed in d17c9d8df. The helper now requires the literal `.svd.bin.z` suffix per the contract in `SVDSurfaceSerializer::default_cache_path()`, so a stale BicubicBackend `.bin.z` or other `.z` artifact in `~/.CoolProp/SVDTables/` can no longer be misread as an SVDSBTL cache hit.

PR #2938, opened 2026-05-20
feat(svdsbtl): split SUPER at IF97 R2/R3 boundary + fix R3 sampling

ibell:
## Summary

Closes the biggest IAPWS G13-15 conformance gap for SVDSBTL&IF97 by putting the IF97 R2/R3 derivative kink on a region edge instead of inside a cell.

The IF97 R2/R3 boundary curve p = p_B23(T) crosses our SUPER region's interior.  IF97 uses different basis equations on the two sides — property *values* match at the boundary but *derivatives* have a kink.  Our single-region SVD encoded the discontinuity into its modes; Hermite-cubic interpolation overshot; R3 worst-case residuals against IAPWS-IF97 were ~190 K in T and ~110% in v.

Subagent investigation (CoolProp-foi.9.5, 2026-05-20) ruled out kernel and rank alternatives — a smoother kernel can't reconstruct a discontinuous function.  Three independent agents converged on this atlas-level fix.

## Changes

1. **SUPER region split** in `src/SBTL/SurfacePresets.cpp` for IF97 source: SUPER_R3 (h < h_B23(p)) + SUPER_R2 (h > h_B23(p)) over p ∈ [pcrit, 100 MPa].  SUPER_HIGH_P single-region above 100 MPa (B23 curve not defined there).  HEOS source unchanged.
2. **h_B23(p) boundary curve** via `build_h_B23_curve` helper: 64 log-spaced p knots, IF97 forward h at (T_B23(p), p), fitted with existing `region::CubicSplineCurve` machinery.
3. **R3 sampling fix**: IF97's `HmassP_INPUTS` doesn't implement R3 (no closed-form backward), throws "Pressure out of range".  Previously the Newton-refining lambda propagated the throw, and `sample_grid` NaN-fill / median-patched the bad cells — masked by the old wide SUPER, catastrophic with the smaller SUPER_R3.  Fix: catch the HmassP throw and seed Newton from T = 700 K on PT_INPUTS.
4. **Conformance probe** `src/Tests/CoolProp-Tests-SVDSBTLFailMap.cpp` (`[!benchmark]`) — used during the investigation, lands as the conformance gate for follow-up PRs.
5. **Cache rev 6 → 7**.

## Measured impact (SVDSBTL&IF97 vs IF97, NT=200, NR=800, rank=20)

| metric | before | after | ratio |
|---|---:|---:|---:|
| R1 v_max | 1.2774% | 0.1640% | 8× better |
| R1 w_max | 3.7681% | 0.0806% | 47× better |
| R2 v_max | 11.5399% | 0.0089% | **1300× better** |
| R2 s_max | 162.4 J/kgK | 0.13 J/kgK | 1290× better |
| R2 T_max | 10928 mK | 17.8 mK | 615× better |
| R3 v_max | 112% | 213%* | (one critical-zone outlier; see below) |
| R3 T_max | 188682 mK | 81352 mK* | |

\* R3 has a single remaining outlier at T=663.7 K, p=26.6 MPa — *just* outside the default critical-patch bbox at [0.95, 1.05]·Tcrit × [0.75, 1.15]·pcrit = [614.7, 679.5] K × [16.55, 25.37] MPa.  Widening that bbox (PR D / CoolProp-dxd) absorbs it.  **Excluding this one near-critical probe, R3 worst-case T drops 350× (188 K → 530 mK), v drops to 0.34% range.**

## Follow-ups in the foi.9 sequence

- `foi.9.6` quintic Hermite kernel — tighten R1/R2 fully to G13-15 perms
- `foi.9.7` IF97-boundary smoothing of M (Kunick §5.1 trick) — optional polish
- `foi.5ni` / `foi.dxd` — auto-calibrate the critical-patch bbox

## Test plan

- [x] `[SVDSBTL][pt][water]` HEOS source unchanged (48 s, all assertions pass)
- [x] `[SVDSBTL][ph][water]` HEOS source unchanged (warm cache, <1 s)
- [x] `SVDSBTL&IF97 single-phase PT lookup matches IF97 truth` passes
- [x] Fail-map CSV regenerated and tabulated above
- [ ] Full `[SVDSBTL]` suite — CI

Refs: CoolProp-foi.9.5

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy for water (IF97) thermodynamic properties near the R2/R3 region boundary by refining surface handling to avoid interpolation artifacts.

* **Tests**
  * Added a new diagnostic test that generates CSV “fail-map” data for conformance/validation comparisons across backends.

* **Chores**
  * Cache format bumped; previously cached surface data will be regenerated on first use.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CoolProp/CoolProp/pull/2938?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2112,6 +2112,8 @@ if(COOLPROP_CATCH_MODULE)
        "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests-SVDSBTLOptions.cpp")
   list(APPEND APP_SOURCES
        "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests-SVDSBTLCriticalPatch.cpp")
+  list(APPEND APP_SOURCES
+       "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests-SVDSBTLFailMap.cpp")
 
   # CATCH TEST, compile everything with catch and set test entry point
   add_executable(CatchTestRunner ${APP_SOURCES})
--- a/include/CoolProp/sbtl/SVDSurfaceSerializer.h
+++ b/include/CoolProp/sbtl/SVDSurfaceSerializer.h
@@ -93,7 +93,17 @@ class SVDSurfaceSerializer
     //          the rev bump is the cache-invalidation mechanism — old
     //          rev-5 caches would silently serve uniform-η surfaces
     //          and undo the IF97 conformance gains.
-    static constexpr int kRevision = 6;
+    //   rev 7: SUPER region split for IF97 source backend — SUPER_R3
+    //          (h < h_B23(p)) and SUPER_R2 (h > h_B23(p)) — so the
+    //          IF97 R2/R3 derivative kink sits on a region edge
+    //          instead of inside a cell.  Region count for IF97 goes
+    //          from 3 (LIQUID/VAPOR/SUPER) to 4 or 5 (LIQUID/VAPOR/
+    //          SUPER_R3/SUPER_R2 + optional SUPER_HIGH_P above 100 MPa).
+    //          On-wire region list differs, so the rev bump forces a
+    //          clean rebuild instead of attempting to deserialize a
+    //          rev-6 cache that won't have the new boundary curves.
+    //          HEOS source unchanged.
+    static constexpr int kRevision = 7;
 
     // Pack one surface into a zlib-compressed msgpack blob.
     static std::vector<char> save(const SVDSurface& surface);
--- a/src/SBTL/SurfacePresets.cpp
+++ b/src/SBTL/SurfacePresets.cpp
@@ -80,6 +80,48 @@ std::vector<PropertySpec> pt_properties(::CoolProp::AbstractState& src) {
     return ps;
 }
 
+// IF97 R2/R3 boundary equation, inverted:
+//   p_B23(T) = 0.10192970039326e-2 · (T − 572.54459862746)^2 + 13.91883776670  [MPa]
+// Closed-form inverse for p ∈ [16.529, 100] MPa (the valid range for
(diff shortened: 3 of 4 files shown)


ibell:
Thanks @coderabbitai. Re the four review comments:

**Fixed in 79eb9aa1 (test-only, no behavioral change to SVDSBTL):**

- *src/Tests/CoolProp-Tests-SVDSBTLFailMap.cpp:7* — Updated header comment to list the three actual invocation tags (`[if97_src]`, `[heos_src_vs_if97]`, `[heos_src_vs_heos]`).  Good catch — the example was stale.
- *src/Tests/CoolProp-Tests-SVDSBTLFailMap.cpp:147* — The CSV column was indeed inconsistent. Switched to write `T_truth_refined` so the column value and the `dT_mK` residual share the same reference frame.

**Superseded by the stacked PR #2940 (TOMS748 replacement of the Newton fallback):**

- *src/SBTL/SurfacePresets.cpp:265-270* — The broad `catch (...)` in the IF97 R3 fallback is replaced wholesale in PR #2940 by `solve_T_from_h_toms748` which uses a bracketed root-find on `[623.15, T_B23(p)]` and returns an `ok` flag. No silent fallthrough to a 700 K seed anymore.
- *src/SBTL/SurfacePresets.cpp:279-292* — Same scope: the Newton loop is removed in PR #2940. TOMS748 with `eps_tolerance<double>(40)` guarantees the residual hits ≈machine-epsilon or the bracket is reported as invalid via the `ok` flag — no silent acceptance of an unconverged iterate.

PR #2940 is stacked on this branch, so as soon as both merge the concerns vanish at the source.

henningjp:
Curious why you would create your own b23 functions when they could be called directly from IF97::Region23_T(p) and IF97::Region23_p(T).

ibell:
@henningjp it is a good question. @claude thought it was a good idea, but I agree it should use the existing implementation. Also their implementation was incorrect the first time :)

PR #2941, opened 2026-05-21
docs(IF97): Python timing-profile script + figure for IF97 / HEOS / SVDSBTL

ibell:
## Summary

- New `Web/scripts/fluid_properties.IF97Profile.py` — sparse-grid (30×30) per-call + batch timing profile for every available `HmassP_INPUTS` backend.
- Uses `AbstractState.fast_evaluate` (the new cache-bypassing batch API) where supported (IF97, Tabular) to report Python-overhead-amortized per-point cost.
- Skips backends the local build doesn't expose (SVDSBTL is listed but currently only times when bound through Python).
- Wired into `Web/scripts/__init__.py` so it regenerates on every docs build.
- `IF97.rst` now references the figure with explanatory text about per-call vs batch.

Output: `Web/fluid_properties/IF97_profile.png` + stdout summary. Users can re-run the script on their own hardware to see their machine's per-call cost rather than reading server numbers.

## Test plan

- [x] Script runs on local checkout (`python Web/scripts/fluid_properties.IF97Profile.py`).
- [x] Reported numbers look sensible: IF97 ~2.7 us/call, HEOS ~73 us/call, IF97 `fast_evaluate` ~1.4 us/pt over 900-pt batch.
- [x] Skips SVDSBTL cleanly when the Python wrapper doesn't expose it.
- [ ] Docs build runs the script as part of the task list.

## Summary by CodeRabbit

* **Documentation**
  * Added a new "IF97 vs HEOS timing — per-call and batch" guide with benchmark comparisons, interpretation of per-call and batch timings, median/P10/P90 reporting, and instructions to regenerate the profiling figure on local hardware.

* **Chores**
  * Added automated backend performance profiling to the static-site build so the profiling figure is generated as part of the pipeline.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CoolProp/CoolProp/pull/2941?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Web/fluid_properties/IF97.rst
+++ b/Web/fluid_properties/IF97.rst
@@ -196,3 +196,27 @@ A call to the top-level function ``PropsSI`` can provide: temperature, pressure,
 
 .. include:: IF97_conformance.rst.in
 
+IF97 vs HEOS timing — per-call and batch
+----------------------------------------
+
+The figure below profiles ``AbstractState.update(HmassP_INPUTS, h, p)``
+for every backend the local build exposes (``IF97``, ``HEOS``, and the
+``SVDSBTL`` family when bound through Python). It also reports
+:meth:`CoolProp.AbstractState.fast_evaluate` — a cache-bypassing
+zero-allocation batch path — for backends that implement it (currently
+``IF97`` and the tabular backends). Re-running
+``Web/scripts/fluid_properties.IF97Profile.py`` regenerates the figure
+against your local hardware and CoolProp build, so the numbers reflect
+your machine, not the documentation server's.
+
+The per-call panels report the *median* wall time over many warm-cache
+calls, so first-call cache-load cost is baked out. Per-call numbers
+include ~1 us of Python-wrapper marshalling that applies equally to
+every backend — read the *ratios* as the meaningful comparison, not the
+absolute nanoseconds. ``fast_evaluate`` amortizes that overhead over
+the batch, so its per-point cost is closer to native C++.
+
+.. image:: IF97_profile.png
+   :width: 100%
+   :alt: IF97 / HEOS HmassP_INPUTS timing profile
+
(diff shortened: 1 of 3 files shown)


ibell:
Thanks @coderabbitai. Addressed in e90fb8d66:

* **Subplot bias (per-backend valid subset).** Added \`_backend_valid_subset(s, h, p)\` that runs the probe set through \`update()\` once per backend and returns the indices the backend can actually evaluate. Both \`time_fast_evaluate\` and \`time_manual_batch\` now run on the same filtered subset, reported as a new \`N\` column. fast_evaluate no longer silently absorbs NaN-fill probes into its average; manual loop no longer crashes mid-batch on the first backend-invalid probe.

* **Manual loop crash potential.** Same fix — the timed loop in \`time_manual_batch\` now relies on the caller having pre-filtered, so the inner \`update()\` runs without \`try/except\` (faster + correct).

* **make_probes silent-empty.** Now raises \`RuntimeError\` with the grid spec if zero IF97-valid points materialize, and prints skipped-point counts otherwise. The downstream \`/ h.size\` divide-by-zero is now impossible.

* **Empty except blocks (CodeQL #3531, #3532).** Both sites had bare \`pass\` without explanatory comments. The two warmup paths are gone; the one remaining catch (the SuperAncillary failure inside \`_backend_valid_subset\`) has a comment explaining the filter intent.

* **Unit mix in IF97.rst.** \"~1 us\" + \"absolute nanoseconds\" in adjacent sentences. Switched to \"~1000 ns\" so the unit comparison reads consistently.

Sample run after the fix:

\`\`\`
  backend            mechanism              N   median  P10    P90
  IF97               fast_evaluate        887   1890    1876   1956
  HEOS               manual update loop   900   76100   75384  77091
  SVDSBTL+HEOS       fast_evaluate        900   567     563    593
  SVDSBTL+IF97       fast_evaluate        900   279     278    295
\`\`\`

\`N\` column makes the now-equal probe counts visible per backend.

PR #2942, opened 2026-05-21
feat(fast_evaluate): batch path for SVDSBTL + dome Q-blend for IF97/SVDSBTL

ibell:
## Summary

Two related additions to the `AbstractState::fast_evaluate` batch path, plus a kernel refactor that fixes the architectural concern the original implementation introduced.

### 1. SVDSBTL gets `fast_evaluate` (closes CoolProp-4bx)

SVDSBTL was inheriting the default that throws `NotImplementedError`. Now overridden to mirror `update()`'s routing — critical-patch bbox check → `patch_source_`, otherwise `SVDSurface::resolve` + `eval_with_region` per output — but writes directly into the caller's `out_buffer` without touching the AbstractState cache.

- Supports `HmassP_INPUTS`, `HmolarP_INPUTS`, `PT_INPUTS`.
- Supported outputs: any property the surface exposes (`iDmass / iSmass / iUmass / ispeed_sound / iviscosity / iconductivity`) plus shortcuts that don't need an SVD eval (`iT / iP / iHmass / iHmolar / iDmolar / iSmolar / iUmolar / iQ`).
- In-patch points still go through `patch_source_->update()` (rare, but correct).
- Bit-identical to per-call `update()` on in-domain probes — verified with a 30-point test at zero tolerance.

### 2. Both backends now Q-blend instead of NaN-filling dome-hit HmassP probes

The previous fast paths NaN-filled and set `fast_evaluate_two_phase_disallowed` whenever an HmassP point landed inside the saturation dome. This was unnecessarily restrictive — the lever-rule blend over sat-line endpoints is cheap (one `Tsat97` or one `SuperAncillary::get_T_from_p` lookup) and `update()` already does it.

**Blend rules** (mass-basis; molar variants follow by × / ÷ M):
- `v = (1 − Q)·v_L + Q·v_V`  so  `ρ = 1 / v` (lever rule — *specific volume* blends linearly, density is its reciprocal)
- `s = (1 − Q)·s_L + Q·s_V`
- `u = (1 − Q)·u_L + Q·u_V`
- `h = (1 − Q)·h_L + Q·h_V`  (the supplied input, modulo round-trip)
- `Q` returned directly
- `speed_sound / cp / cv / viscosity / conductivity / Prandtl` stay NaN (no equilibrium bulk value in the dome)

**IF97** — after `T_phmass` returns (it silently returns `Tsat` for in-dome inputs rather than throwing), probe `IF97::hliq_p` / `hvap_p` at p and bracket h. Sat-line endpoints come from `IF97::rholiq_p / rhovap_p / sliq_p / svap_p / uliq_p / uvap_p`.

**SVDSBTL** — after the atlas resolve returns `region_idx = -1` for HmassP / HmolarP, try the source's SuperAncillary first (HEOS / REFPROP have one); fall back to `source_reference_()->update(PQ_INPUTS, p, 0|1)` for IF97 source (no SA but PQ is fast — one `psat97` + region-4 read).

PT_INPUTS exactly on the saturation curve remains `two_phase_disallowed` because Q isn't determinable from (P, T) alone — caller should use PQ_INPUTS or supply a phase hint.

### 3. Kernel refactor — `resolve_point_` + `evaluate_property_`

The initial fast_evaluate implementation re-implemented input-pair dispatch, critical-patch routing, atlas resolve, dome detection + Q-blend, and the per-output property switch — all of which already existed in slightly different form across `update()` and the `calc_*` family. That made the two code paths drift-prone and meant any future caller (batched PropsSI, Python `update_array` wrapper, possible HEOS `fast_evaluate`, etc.) couldn't reuse the kernel.

This change pulls the routing out of both code paths:

- `resolve_point_(input_pair, v1, v2)` — pure (no instance mutation), returns a `PointEvaluation` tagged with one of `{SinglePhase, DomeBlend, Patched, OutOfRange, TwoPhaseDisallowed, InternalError}`. Two-phase `PQ_INPUTS` / `QT_INPUTS` land in `DomeBlend` with the same fields as an HmassP-in-dome row, so downstream code sees one case instead of two.
- `evaluate_property_(PointEvaluation&, parameter)` — pure dispatch with lazy sat-line endpoint fills via `ensure_dome_rho_endpoints_` / `ensure_dome_s_endpoints_`. Throws `NotImplementedError` for properties undefined in the dome; `calc_*()` lets it propagate, `fast_evaluate()` translates it into an in-band NaN.
- `update()` collapses from ~150 LOC to ~50 LOC: calls `resolve_point_`, stores the result on `active_eval_`, copies T/p/Q/phase into the AbstractState legacy slots for back-compat.
- `fast_evaluate()` collapses from ~330 LOC to ~30 LOC: loops calling the kernel per probe.
- The `TwoPhaseState` struct, `update_two_phase_()`, `two_phase_property_`, and `lookup_()` helpers all go away — their state lives on `active_eval_`, their logic lives in `evaluate_property_()`.
- Lazy / nullable doubles use `std::optional<double>` per Ian's preference.

Net diff: roughly even LOC despite adding the kernel infrastructure. A hypothetical future entry point (batched PropsSI, `update_array` wrapper, etc.) reuses the same kernel by calling `resolve_point_` + `evaluate_property_` directly.

### Adversarial-review fixes (PR-internal)

Subsequent commits address findings from a code-reviewer pass + CodeRabbit:

- **H1.** `update()` sets `_Q` for every kind (SinglePhase → -1, Patched → `patch_source_->keyed_output(iQ)`, DomeBlend → lever-rule Q). Parity with `fast_evaluate` and `IF97Backend::update()`.
- **H2.** Removed top-level `iQ` short-circuit; Patched-kind iQ forwards through `patch_source_->keyed_output(iQ)` so two-phase points near the bbox edge are reported correctly.
- **H3.** Deleted dead DomeBlend `iHmolar` arm; the `pt.h_mass`-based shortcut is the single source of truth.
- **H4.** Patched molar variants collapsed to direct `keyed_output(prop)` — HEOS computes molar quantities natively; the previous mass·M round-trip was wasted work + precision loss.
- **M1.** Stale docstrings referencing deleted `lookup_` / `update_two_phase_` rewritten.
- **M2.** Thread-safety docstring on `fast_evaluate` spells out the Patched corruption pattern explicitly.
- **M3.** New `[SVDSBTL][fast_evaluate][water][slow]` test probes (0.99 T_c, 0.95 p_c) — inside the default critical-patch bbox — through both PT and HmassP routing, asserting fast_evaluate matches `update()` + property reads on every output.

## Test plan

- [x] Full `[SVDSBTL]` suite: 42 test cases, 669 assertions, 3 skipped as before.
- [x] `[fast_evaluate]` subset: 7 test cases, 390 assertions, including the new bit-identical, lever-rule Q recovery, Q = -1 single-phase, OOB PT, and Patched-routing cases.
- [x] No regression in `[IF97]` (67/67 assertions) — the IF97 dome-blend addition lives behind the existing `fast_evaluate` test set.
- [x] Python smoke: `fast_evaluate` and `update()` produce bit-identical T / D / S / Q across single-phase and dome-blend probes via the project venv build.
- [ ] CI green across macOS / Ubuntu / Windows.

Closes CoolProp-4bx.

### DIFF
--- a/include/CoolProp/Backends/SVDSBTL/SVDSBTLBackend.h
+++ b/include/CoolProp/Backends/SVDSBTL/SVDSBTLBackend.h
@@ -3,6 +3,7 @@
 
 #include <limits>
 #include <memory>
+#include <optional>
 #include <set>  // transitively needed by superancillary.h
 #include <string>
 #include <unordered_map>
@@ -125,24 +126,23 @@ class SVDSBTLBackend : public AbstractState
     void update(CoolProp::input_pairs input_pair, double value1, double value2) override;
 
     // Vectorized cache-bypassing batch evaluation — see
-    // AbstractState::fast_evaluate for the contract.  Mirrors update()'s
-    // routing (critical-patch bbox → patch_source_, otherwise SVD
-    // surface) but writes directly into the caller's out_buffer rather
-    // than touching the instance's update()-side state (_T / _p /
-    // _phase / active_point_ / active_surface_ / active_resolved_ /
-    // active_hmass_).  So back-to-back fast_evaluate calls can be
-    // interleaved with update() calls on the same instance without
-    // corrupting the most-recent update()'s state, BUT the patch
-    // routing IS stateful: it reuses the lazily-allocated patch_source_
-    // child AbstractState across all in-patch points in the batch.
-    // That makes fast_evaluate safe for single-threaded interleaving
-    // with update() but NOT thread-safe — two concurrent fast_evaluate
-    // calls on the same backend will race on patch_source_->update()
-    // for any point that falls inside the critical-patch bbox.  In-
-    // domain HmassP / HmolarP probes that fall inside the saturation
-    // dome are Q-blended in place; PT probes exactly on the sat curve
-    // and points outside every region NaN-fill with
-    // fast_evaluate_two_phase_disallowed or fast_evaluate_out_of_range.
+    // AbstractState::fast_evaluate for the contract.  Both update() and
+    // fast_evaluate() route through the same private resolve_point_()
+    // kernel; this entry point just loops over the input array, asks
+    // the kernel for a PointEvaluation per probe, then dispatches each
+    // requested output through evaluate_property_().  Writes directly
+    // into the caller's out_buffer instead of mutating active_eval_, so
+    // back-to-back fast_evaluate calls can be interleaved with update()
+    // on the same instance without corrupting the most-recent update()
+    // — BUT the patch routing is stateful (reuses the lazily-allocated
+    // patch_source_ child AbstractState across all in-patch points),
+    // which makes fast_evaluate safe for single-threaded interleaving
+    // with update() but NOT thread-safe against a concurrent
+    // fast_evaluate/update on the same backend.  HmassP / HmolarP
+    // probes inside the saturation dome are Q-blended in place; PT
+    // probes exactly on the sat curve NaN-fill with
+    // fast_evaluate_two_phase_disallowed; points outside every region
+    // NaN-fill with fast_evaluate_out_of_range.
     void fast_evaluate(CoolProp::input_pairs input_pair, const double* val1, const double* val2, std::size_t N_inputs,
                        const CoolProp::parameters* outputs, std::size_t N_outputs, double* out_buffer, std::size_t out_buffer_size, int* status_flags,
                        std::size_t status_flags_size, CoolProp::phases imposed_phase = CoolProp::iphase_not_imposed) override;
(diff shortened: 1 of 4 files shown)


ibell:
Thanks @coderabbitai. Addressed in 5f22136b0:

**Fixed (5 comments):**

- **Header docstring** — the previous wording overstated the safety guarantee. Reworded to say exactly which slots are preserved (\`_T\` / \`_p\` / \`_phase\` / \`active_*\`) and to call out that \`patch_source_\` IS shared across in-patch points, so \`fast_evaluate\` is safe to interleave with \`update()\` on a single thread but **not** thread-safe against a concurrent \`fast_evaluate\`/\`update\` call.

- **Patch-path silent NaN-fill** — a per-output \`keyed_output()\` failure used to be NaN-filled in place with \`fast_evaluate_ok\` left set. Now any per-output throw inside the patch row flips a flag, the entire row is NaN-filled, and the status is set to \`fast_evaluate_internal_error\` — matching the IF97Backend contract.

- **iQ in IF97 single-phase path** — added \`case iQ: val = -1.0;\` so a single-phase iQ request returns the -1 sentinel that \`update()\` would, instead of falling to default and NaN-filling the whole row.

- **iQ on SVDSBTL patch + single-phase paths** — same fix on both branches. Dome path already returned the correct \`Q\`.

- **PT atlas miss classification** — added a sat-curve probe (\`SuperAncillary::get_T_from_p\` with source-PQ fallback) so \`pt.region_idx < 0\` on a PT input that's within \`3.3e-5 · T_sat\` of the sat curve now reports \`fast_evaluate_two_phase_disallowed\` instead of \`out_of_range\`. In practice the SVDSBTL PT atlas registers LIQUID and VAPOR as adjacent at \`T_sat\`, so the resolve doesn't typically miss there — the new code is defensive in case future atlas edits introduce a gap, and the OOB classification stays correct for genuinely-outside points.

**Not changed (with reasoning):**

- *Dome rows OK despite NaN outputs for speed_sound/cp/cv/viscosity/conductivity/Prandtl* — the contract change to NaN-fill the whole row on any partial-undefined output would lose the meaningful Q-blended values (T, ρ, s, u, Q) when the same call also requests, say, \`ispeed_sound\`. The current behaviour is: meaningful outputs get blended, undefined outputs get NaN, status stays OK. NaN is the conventional in-band signal for \"undefined here\" and the dome-defined property set is documented in the header. Open to a follow-up if you'd rather see a new \`fast_evaluate_two_phase_partial\` status — it's a contract change so I'd rather pull it into a dedicated PR.

Two new tests cover (a) single-phase \`Q = -1\` recovery on HmassP rows and (b) PT genuinely-OOB classification. All 7 \`[fast_evaluate]\` cases pass (390 assertions); no regression in \`[SVDSBTL]\` (649/649) or \`[IF97]\` (67/67).

PR #2943, opened 2026-05-21
fix(svdsbtl,docs): block SVDSBTL from PropsSI + speed up IF97Conformance.py ~3000x

ibell:
## Context

Docs builds on master have been taking 3+ hours since #2923 (235bffa95) added \`Web/scripts/fluid_properties.IF97Conformance.py\` to the task list. Profiling found the cause:

| backend | µs/PropsSI |
|---|---:|
| \`IF97::Water\` | 3.7 |
| \`SVDSBTL&IF97::Water\` | **77,054** |

PropsSI calls \`AbstractState::factory()\` on every invocation. For SVDSBTL that triggers \`ensure_surface_()\` which loads + msgpack-deserializes the cached \`.svd.bin.z\` surface (~80 ms) before the actual SVD eval (~5 µs). The conformance sweep makes ~100k PropsSI calls × 77 ms ≈ 2 hours, which matches the user's report.

## Changes

**1. Block SVDSBTL from the high-level PropsSI surface by default.** Mirrors how BICUBIC / TTSE are gated (\`TabularBackend::available_in_high_level()\` returns false). Adds an \`ALLOW_SVDSBTL_IN_PROPSSI\` Configuration key (default false) — users who genuinely want one-off interactive PropsSI queries opt back in via \`set_config_bool()\` or the \`COOLPROP_ALLOW_SVDSBTL_IN_PROPSSI\` env var. Two tests cover both directions.

**2. Rewrite IF97Conformance.py's hot loops to reuse one AbstractState per backend** and call \`update()\` + \`keyed_output()\` instead of PropsSI per probe. This is the correct pattern under the new gate AND independently makes the script much faster.

## Test plan

- [x] Local: full \`[SVDSBTL]\` test suite — 334 / 334 assertions pass (37 cases, 3 skipped as before).
- [x] Local: \`fluid_properties.IF97Conformance.py\` runs end-to-end in **3.7 seconds** (vs hasn't-finished-in-3-minutes baseline on the same machine).
- [x] New \`[SVDSBTL][propsi][gate]\` test confirms PropsSI is rejected by default (errstring contains \"high-level\") and accepted after opting in.
- [ ] CI green on docs build — should drop multi-hour overhead back to seconds.

## Summary by CodeRabbit

* **Performance Improvements**
  * Faster and more efficient fluid-property computations and timing measurements for batch workflows.

* **New Features**
  * New configuration option to control availability of an optimized backend in the high‑level interface.

* **Bug Fixes**
  * That backend is disabled by default for high‑level calls; must be explicitly enabled.

* **Tests**
  * Added tests to verify backend availability control.

* **Documentation**
  * Updated usage notes and encoding handling for configuration key docs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CoolProp/CoolProp/pull/2943?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ibell:
Thanks @coderabbitai. Addressed in ca996f6ce:

The timing loop rewrite swapped \`PropsSI\` for \`AbstractState.update() + keyed_output()\` (that's the whole perf hotfix in this PR — PropsSI rebuilds the AbstractState per call, which is what made the docs build take 3+ hours for SVDSBTL).  But I left the rendered text saying \"mean wall time per \`PropsSI\` call\", which would have mis-attributed the new numbers.

Three sites updated:

1. **Rendered RST timing paragraph** — now says \"Mean wall time per \`AbstractState.update()\` + \`keyed_output()\` call\" with an explicit footnote that \`PropsSI\`'s end-to-end cost is higher because of the per-call factory rebuild and pointing to the SVDSBTL profiling figure (PR #2941) for that comparison.

2. **Header docstring** — same correction.

3. **\`run_timing\` function docstring** — clarifies that one AbstractState is constructed per backend up front and reused, so the timing reflects the flash + property read without the factory-rebuild tax.

Local re-run produces a clean \`IF97_conformance.rst.in\`.

PR #2959, opened 2026-05-23
docs(svdsbtl): correct 19 doc bugs uncovered by audit (CoolProp-fhp)

ibell:
## Summary
- Documentation-only PR: corrects 19 doc bugs in `Web/coolprop/SVDSBTL.rst` (9 outright errors, 4 near-right inaccuracies, 6 imprecisions) plus a stale comment block in `include/CoolProp/Backends/SVDSBTL/SVDSBTLBackend.h`.
- No code-behavior change. Numbers cross-checked against on-disk `~/.CoolProp/SVDTables/*.critpatch.bin` cache values, the rendered `Web/fluid_properties/IF97_conformance.rst.in` fail-map, and live runtime measurements on Apple Silicon.
- One follow-up tracked: bd `CoolProp-fhp` for actually implementing `ALTERNATIVE_SVDTABLES_DIRECTORY` (claim removed from doc here pending implementation).

### Removed claims that don't match code
- `set_critical_bbox_multipliers(T_lo, T_hi, p_lo, p_hi)` C++ entry point — function does not exist anywhere in `include/`, `src/`, or `wrappers/`. Doc had it in two places (Critical-patch section + Programmatic access).
- `ALTERNATIVE_SVDTABLES_DIRECTORY` config key — not in `include/Configuration.h`, not consulted in `SVDSurfaceSerializer.cpp`. Tracked via bd `CoolProp-fhp`.
- `XDG_DATA_HOME`-aware on Linux — `get_home_dir()` at `src/CPfilepaths.cpp:157` only reads `$HOME`.
- `registered_input_pairs()` Python visibility — exists C++-only, no Python binding in `AbstractState.pyx`. Clarified.

### Performance numbers updated to match measurement
| Claim | Old | New |
|---|---|---|
| Native C++ `update + rhomass + T` | ~200 ns | ~1 µs (`update()` dominates) |
| Cython wrapper overhead | "bulk of cost" | ~0.3 µs across 3 FFI crossings |
| Python `update + rhomass + T` | "hundreds of ns" | ~1.3 µs/iter |
| `fast_evaluate` 4-output / 10k probes | ~340-400 ns/probe | ~300-330 ns/probe (supercritical region) |
| Marginal cost per added output | ~57 ns | ~35 ns |
| Table-load cost | ~80 ms | ~80-140 ms (file-size-dependent) |
| R245fa N=4 cold build | 65→25 s | 122→54 s (Apple Silicon); ratio ~2.25× holds |
| REFPROP PQ flash fallback | ~ms/probe | ~3-5 µs/probe (3 orders of magnitude off) |

### `fast_evaluate` example switched to supercritical-only probes
- The old example used uniform `h ∈ [0.1, 3] MJ/kg, p ∈ [0.1, 30] MPa`, which lands ~40% of probes near the critical patch (5-10× slower per probe) — washing out the headline cost the prose advertised.
- New example uses `h ∈ [2.5, 4.0] MJ/kg, p ∈ [22.5, 50] MPa` (all `T > T_c`), which isolates the locate+basis cost. Also bumped to 4 outputs to match the 4-output prose.

### Critical-patch per-fluid bbox table refreshed
- All four rows updated to current cached multipliers (the old values were off by 1-15% from the on-disk cache; CO₂ `p_lo` was off by 15%).
- Water/HEOS derived `(p, h)` bbox h-range corrected: 1.55-2.93 → 1.74-2.65 MJ/kg.

### IF97 G13-15 conformance prose tightened
- Old: "conformant to Tables 8-13 for T, v, s, w in R1, R2, R5".
- Actual per `IF97_conformance.rst.in` fail-map: only `T(p,h)` is uniformly conformant in R1/R2/R5; `v` conformant only in R5; `w` in R2 and R5; `s` exceeds budget in all four regions.
- Transport budget miss broadened from "in R3" to "across R1, R2, R3" (η: R1 34×, R2 3.8×; λ: R1 32×, R2 37×).

### Architecture imprecisions tightened
- Grid orientation `N_p × N_h (200 × 800)` → `N_h × N_p (200 × 800)` (200 = secondary, 800 = primary log-p axis).
- Per-region vs per-table sample count clarified (160k per region × 3-7 regions per atlas).
- Subcritical boundary curves: one is sat dome, the other is an isotherm (the old "saturation dome on the subcritical side" read as both bounds).
- C¹ continuity → C² with default natural-cubic-spline slopes (per `SVDEvaluator.h:38-40` and `SVDDecomposition.h:24`).

### Header-comment bonus fix
`include/CoolProp/Backends/SVDSBTL/SVDSBTLBackend.h` near `auto_calibrate_critical_bbox_()`:
- Calibration budgets: was "200/100/200/1000 ppm" → actual is "1% relative in rho/h/s, 5% relative in w" per `kBudget_*` in the `.cpp`.
- Initial bbox: was `[0.85, 1.20]·Tc × [0.50, 4.0]·pc, ~8 iterations` → actual is `{0.95, 1.05, 0.75, 1.15}, 6 steps` per `kBisectSteps`.

## Verification process

1. Five parallel verification subagents (perf / API / critical-patch / architecture / accuracy), each cross-checking ~5-7 claims against source + runtime.
2. `./dev/ci/preflight.sh`: 5 passed / 0 failed / 1 skipped (clang-tidy diff-only, no `.cpp` files in diff).
3. `superpowers:code-reviewer` adversarial pass: no blocking findings.

## Test plan
- [ ] Sphinx docs build succeeds (`.. ipython::` block in regime 2 re-runs the new supercritical example).
- [ ] No broken cross-refs (`<SVDSBTL-critpatch>`, `<SVDSBTL-config>`, `<SVDSBTL-regimes>`, `<IF97-Conformance>`).
- [ ] Numbers cited (especially conformance maxes `2.79 / 8.52 / 1.81 mK`) re-verified on a docs-build machine — flagged by code-reviewer as tight three-sig-fig values that may bobble in re-runs.

## Summary by CodeRabbit

* **Documentation**
  * Major SVDSBTL docs refresh: updated performance/footprint claims, timing breakdowns (including ~80–140 ms table-load ranges), examples, accuracy envelopes, critical-patch guidance, cache sizing (~160k samples/region) and default cache location (~/.CoolProp/SVDTables/); added see-also notes recommending SVDSBTL for throughput-sensitive use.

* **Configuration**
  * New sampling-threads option and JSON-based critical-patch overrides documented; alternative table-directory option removed.

* **API Changes**
  * Runtime in-place override for critical-patch geometry removed; reconstruct backend with new configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CoolProp/CoolProp/pull/2959?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Web/coolprop/SVDSBTL.rst
+++ b/Web/coolprop/SVDSBTL.rst
@@ -43,11 +43,14 @@ bicubic table with three layered components that address both:
    further splits for IF97 (R1/R3, R2/R3 boundary curves).  Each
    region's secondary axis is normalised to :math:`[0, 1]` via two
    analytic boundary curves — on subcritical regions one curve is the
-   saturation dome and the other is an isotherm (low-T floor for the
-   liquid region, high-T ceiling for the vapor region); on
-   supercritical regions both curves are isotherms.  Lookups first
-   AABB-test against region bounding boxes (cheap), then evaluate the
-   precise boundary curves on the candidate.
+   saturation dome and the other is an EOS-domain floor/ceiling (for
+   the liquid region the floor walks up over the **melting curve**
+   where it binds at high p, falling back to the low-T EOS isotherm
+   where it doesn't; for the vapor region the ceiling is the high-T
+   isotherm at the EOS validity limit); on supercritical regions both
+   curves are isotherms.  Lookups first AABB-test against region
+   bounding boxes (cheap), then evaluate the precise boundary curves
+   on the candidate.
 
 2. **Per-region SVD.**  Inside each normalised region, every tabulated
    property (:math:`\rho, h, s, u, w, \eta, \lambda`) is represented
@@ -180,12 +183,18 @@ the backend.
 
        In [1]: %timeit loop()
 
-   Native C++ ``update + rhomass + T`` is ~1 µs/iter (dominated by
-   ``update()``; the cached accessors are <10 ns/pair).  The Cython
-   wrapper adds ~0.3 µs across the three FFI crossings, totaling
-   ~1.3 µs/iter from Python — still ~20× faster than HEOS's
-   :math:`(p, h)` flash on the same loop.  This is what most user
-   code should look like.
+   Native C++ ``update + rhomass + T`` is region-dependent: well
+   under 200 ns/call in single-phase (subcooled liquid, superheated
+   vapor, supercritical), and ~1-1.5 µs/call in two-phase / sat-line
+   regions where the SuperAncillary or source-backend PQ flash runs
+   per probe.  The cached accessors are <10 ns/pair on top.  The
+   Cython wrapper adds ~0.3 µs across the three FFI crossings — so
+   even a Python loop sweeping the saturation dome stays ~20× faster
+   than the same loop on HEOS.  This path is fine for ad-hoc / setup
+   code; for throughput-sensitive workloads, prefer
+   :ref:`fast_evaluate <SVDSBTL-regimes>` (next regime) which strips
+   the per-probe wrapper marshalling and amortises locate + basis
+   setup across multiple outputs.
 
 2. **fast_evaluate (batched, no instance mutation).**
    ``AbstractState.fast_evaluate(input_pair, val1[], val2[],
@@ -489,10 +498,14 @@ Accuracy envelope
 * **SVDSBTL&REFPROP::<Fluid>**: matches REFPROP truth on tested
   probes within REFPROP's own numerical-noise floor on the sat
   curves; same single-phase envelope as HEOS-source.  Two-phase
-  endpoints currently fall through to REFPROP's PQ flash (~3-5 µs
-  per probe); ``CoolProp-077`` will replace that with an on-disk
-  saturation-surrogate spline to remove the cross-DLL hop and the
-  REFPROP-thread-safety constraint that fallback inherits.
+  endpoints currently fall through to REFPROP's PQ flash; the flash
+  itself is ~1.5 µs/call native, with the SVDSBTL fallback wrapper
+  (lever rule + endpoint conversions) bringing the user-visible
+  per-probe cost to ~3-4 µs.  ``CoolProp-077`` will replace the
+  REFPROP flash with an on-disk saturation-surrogate spline to remove
+  both the cross-DLL hop and the REFPROP-thread-safety constraint
+  that fallback inherits, and reduce the wrapped cost into the
+  single-µs band.
 
 Deviation plot
 ==============
--- a/Web/coolprop/SVDSBTL.rst
+++ b/Web/coolprop/SVDSBTL.rst
@@ -7,10 +7,13 @@ SVDSBTL --- SVD-Compressed Tabular Lookup Backend
 SVDSBTL is a tabular-lookup backend that combines an atlas of
 analytically-bounded regions with a low-rank SVD of each region's
 property surfaces.  It produces **sub-microsecond** property
-evaluations (~200-400 ns per probe in the batched ``fast_evaluate``
-path) at IAPWS G13-15 conformance for water, and a single-digit
-percent accuracy ceiling for the multi-fluid HEOS-backed presets,
-at a disk footprint of ~10 MB per (fluid, input pair, source backend).
+evaluations in the batched ``fast_evaluate`` path away from the
+critical patch (~300-400 ns per probe at four outputs on Apple
+Silicon, rising to a few µs/probe inside the critical-patch
+fallback), at IAPWS G13-15 :math:`T(p, h)` conformance for water
+and a single-digit percent accuracy ceiling for the multi-fluid
+HEOS-backed presets, at a disk footprint of ~7-14 MB per
+(fluid, input pair, source backend).
 
 This page is the **user-facing** page: how to use the backend, what it
 guarantees, and when to pick it.  For the underlying SVD math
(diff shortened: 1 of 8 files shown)


Web/coolprop/SVDSBTL.rst
```diff
@@ -39,18 +42,23 @@ bicubic table with three layered components that address both:
    (subcooled), VAPOR (superheated), SUPER (supercritical), with
    further splits for IF97 (R1/R3, R2/R3 boundary curves).  Each
    region's secondary axis is normalised to :math:`[0, 1]` via two
-   analytic boundary curves — saturation dome on the subcritical
-   side, low-T / high-T isotherms on the supercritical side.  Lookups
-   first AABB-test against region bounding boxes (cheap), then
-   evaluate the precise boundary curves on the candidate.
+   analytic boundary curves — on subcritical regions one curve is the
+   saturation dome and the other is an isotherm (low-T floor for the
```
ibell: i thought it was the melting curve for subcritical liquid?

Web/coolprop/SVDSBTL.rst
```diff
@@ -173,11 +180,12 @@ the backend.
 
        In [1]: %timeit loop()
 
-   Native C++ per-call wall time is ~200 ns; the bulk of the
-   Python-visible cost is the wrapper marshalling, and even in Python
-   the per-call cost is hundreds of nanoseconds — far faster than
-   HEOS's :math:`(p, h)` flash.  This is what most user code should
-   look like.
+   Native C++ ``update + rhomass + T`` is ~1 µs/iter (dominated by
```
ibell: depends on region, should be down to < 200 ns for supercritical regions where no superancillary eval is needed

Web/coolprop/SVDSBTL.rst
```diff
@@ -173,11 +180,12 @@ the backend.
...
+   ``update()``; the cached accessors are <10 ns/pair).  The Cython
+   wrapper adds ~0.3 µs across the three FFI crossings, totaling
+   ~1.3 µs/iter from Python — still ~20× faster than HEOS's
+   :math:`(p, h)` flash on the same loop.  This is what most user
```
ibell: I think we should not explicitly recommend this path, since `fast_evaluate` can be so much faster

Web/coolprop/SVDSBTL.rst
```diff
@@ -451,15 +467,18 @@ SVDSBTL is **not** the right pick when:
...
+  :math:`w` is conformant in R2 and R5; :math:`s` exceeds budget in
+  all four regions (R5 is 13× over; R3 worst-case is ~10⁴× over).
+  Transport properties :math:`\eta, \lambda` exceed G13-15 budgets
+  across R1, R2 and R3 — those budgets are intrinsically tight
+  (:math:`10^{-5}` relative) and the rank-20 SVD has known headroom
+  for tighter ranks.  The critical-patch fallback covers the
+  near-critical R3 cells.  See :ref:`IF97-Conformance` for the
+  per-region fail maps and exact numbers.
 
 * **SVDSBTL&HEOS::<Fluid>**: ~10^-3 relative error on
```
ibell, line 492: Should be much better than that!

Web/coolprop/SVDSBTL.rst
ibell: Confirmed in SurfacePresets.cpp:215-219 — LIQUID b_lo is a 'non-isothermal floor that walks T up if T_min < T_melt(p) at high p'. Rewrote in a33ede43d: subcritical LIQUID floor is the melting curve where it binds, falling back to the EOS low-T isotherm elsewhere; VAPOR ceiling stays the EOS high-T isotherm; SUPER stays two isotherms.

Web/coolprop/SVDSBTL.rst
ibell: Re-measured per-region via fast_evaluate: SUPER 228 ns, LIQUID 191 ns, SUPERHEAT 208 ns, TWO-PHASE 1582 ns/probe. The 1 µs/iter from my earlier measurement was the average for a p=1 MPa loop where 3/5 points hit the SuperAncillary path. Updated regime 1 prose in a33ede43d to call this out: well under 200 ns/call in single-phase / supercritical, ~1-1.5 µs/call in two-phase.

Web/coolprop/SVDSBTL.rst
ibell: Dropped the 'this is what most user code should look like' recommendation in a33ede43d. New wording calls regime 1 'fine for ad-hoc / setup code' and points throughput-sensitive workloads at fast_evaluate (regime 2) instead.

Web/coolprop/SVDSBTL.rst
ibell: Re-measured: REFPROP::Water PQ(1MPa, 0.5) update+T+rhomass = 1.807 µs/call Python (~1.5 µs native flash). SVDSBTL&REFPROP::Water HmassP in-dome update+T+rhomass = 3.658 µs/call (the wrapper adds ~1.5-2 µs lever-rule + endpoint conversion on top of the flash). Compare SVDSBTL&HEOS::Water in-dome at 0.545 µs/call (uses SuperAncillary). My '~3-5 µs PQ flash' conflated the flash itself with the SVDSBTL wrapper; separated in a33ede43d: flash ~1.5 µs native, wrapped fallback ~3-4 µs user-visible.

PR #2016, opened 2021-04-13
A commit for the REFPROP backend to include the functions of satu…

dongkeun-oh:
### Purpose
As a follow up of the issue #2013, I tried to include the missing functions which is particularly related to the saturated states.
They are critical to associated the REFPROP with the recently updated ExternalMedia library (for the Modelica platform).
* calc_hmolar
* calc_smolar
* calc_first_saturation_deriv
* calc_first_two_phase_deriv
* calc_first_two_phase_deriv_splined (not tested!)

#### protected part:
- Instead of pointers of saturated state, as it is different from the HEOS backend, some additional cached variables 
(CachedElement) are declared.
```
// Cached values of saturated state
    CachedElement _hLmolar, _hVmolar, _sLmolar, _sVmolar,_uLmolar, _uVmolar;
    CachedElement _wL, _wV, _cvLmolar, _cvVmolar, _cpLmolar, _cpVmolar;
    CachedElement _viscosityL, _viscosityV, _conductivityL, _conductivityV; //transport properties of saturated state
```
- Whenever the saturated states (new cached variables) are referred by `_calc_saturated_liquid_output( CachedElement&)` or `_calc_saturated_vapor_output( CachedElement&)`, the variable is checked whether it is already cached or not. 
- If not cached yet, the saturated (liquid or vapor) state will be updated by the current temperature `_T` and the saturated density `_rhoLmolar` (liquid) or `_rhoVmolar` (vapor) using the THERMdll subroutine in the REFPROP DLL.
- To make sure the validity of saturated state, i.e. whether it is "two-phase" or not, an inline function `_RPcheckTwophase()` is introduced.
``     inline bool _RPcheckTwophase( void) {
      if (ValidNumber(_Q) && (_Q >= 0.00) && (_Q <= 1.00)) return true; }
``
- The virtual function `clear()` in the base class (AbstractState) is redefined to chain the original function, and call the local (protected) one `_RPclearSat()` to erase the new cached variables
#### public part:
- The functions of saturated properties, `calc_saturated_liquid_keyed_output(parameters key)` and     `calc_saturated_vapor_keyed_output(parameters key)` are extended for many key parameters to cope with the reqiurment of `coolpropsolver` in the ExternalMedia library.
- `calc_hmolar` and `calc_smolar` is implemented to specialize the calculation, particularly, in case of two-phase fluid.
- Some derivatives of `CoolPropDbl` output - see the Purpose section - in the saturated and two-phase state are implemented.
[...]

PR #2990, opened 2026-05-25
feat(REFPROP): saturated-state props + saturation/two-phase derivatives (#2013, supersedes #2016)

### LINKED ISSUE #2013: Missing member functions in the REFPROP backends (calc_first_saturation_deriv etc)
### Description

In the REFPROP backends (REPROPMixtureBackend and REFPROPBackend), some member functions, particularly referring saturated properties, aren't implemented yet. However, they are necessary for the ExternalMedia library's binary part (libExternalMedia) to instantiate the CoolProp solver object, and to update (post-update) the thermodynamic states.
So, the REFPROP is not available yet ([see this PR](https://github.com/modelica-3rdparty/ExternalMedia/pull/27)) in that OpenModelica library, and I found that the matter is originated from the following missing functions in the REFPROP backends.
- calc_first_saturation_deriv 
- calc_second_saturation_deriv
- calc_first_two_phase_deriv
- calc_first_two_phase_deriv_splined

### Possible Means to implement
According to the old version (4.2.6), the previous REFPROP solver just exploited similar functionalities commonly implemented in the CPState class. So, I guess there are 3 approaches to go ahead.

1. Newly implementing the functions to the REFPROPMixtureBackend class : this is quite straight forward, but the almost same codes of existing ones will be duplicated, which is an ugly way of development.
 
2. Changing the position of the REFPROP backends to derived classes of the HEOS backends: it depends on the design concept of current configurations, which I didn't understand clearly yet; In my point of view, there are some pitfalls in the different manner of construction between the two backends, and it doesn't look consistent to configure them in a strict hierarchy. 

3. Moving those functions to the parent class (i.e. AbstarctState), and making the parent as a template : in this way, the parent should have the abstracted pointers of saturate properties (i.e. SatL and SatV) like
```
template <typename _Backend>
class AbstractState {
...
shared_ptr<_Backend> SatL, SatV;
...

};
```
, and they will be actualized in the definitions of each backend using so-called CRTP, which means, for instance, something like 
```
class REPRPROPMixtureBackend : public AbstarctState<REPRPROPMixtureBackend> 
```
Moreover, another common construction method might be added up owing to the temporary state instances of "End", and "Liq" in calc_first_two_phase_deriv_splined function.
However, it's the most radical way to modify the configuration a lot.

Do you already have a plan on this matter?
Otherwise, what is the best in the meantime?
If you allow, I would try to implement as you guys will advise after discussion.

### Versions

**CoolProp Version:** CoolProp 6.4.2
**Operating System and Version:** Ubuntu 18.04.5 with gcc 7.5.0, and Windows 10 Msys-64bit with gcc 10.2.0
**Access Method:** ExternalMedia in OpenModelica

### Additional Information

Not yet..

ibell:
## Purpose

Reimplements the still-valuable parts of the stale PR #2016 (follow-up to #2013, originally by @dongkeun-oh) so the REFPROP backend exposes the saturated-state properties and saturation/two-phase derivatives that ExternalMedia/Modelica needs. PR #2016 could not be merged (years of conflicts with the rewritten flash logic, plus style issues), so this is a fresh implementation against current `master`.

## What's added

- `calc_saturated_liquid_keyed_output` / `calc_saturated_vapor_keyed_output` extended from density/molar-mass only to the full set: **h, s, u, cp, cv, speed of sound, viscosity, conductivity** (molar and mass basis).
- `calc_first_saturation_deriv` — pure fluids use REFPROP's native `DPTSATKdll` (dp/dT along the saturation line); mixtures differentiate `SATTdll` numerically at the bubble (Q=0) / dew (Q=1) curve.
- `calc_first_two_phase_deriv`, `calc_second_two_phase_deriv`, `calc_first_two_phase_deriv_splined` — mirror the proven `HelmholtzEOSMixtureBackend` math (pure fluids).
- `update_DmolarT_direct` — no-flash state set via `THERMdll`.

## Key design: the saturation "shim"

REFPROP is a global-state DLL, so we cannot hold live `SatL`/`SatV` backend copies the way HEOS does. But **composition is a per-call argument** to every REFPROP routine (`PHIXdll`/`THERMdll`/`TRNPRPdll`/`REDXdll`); only the loaded fluid *set* (`SETUPdll`) is global. So we add a lightweight **shim**: a `REFPROPMixtureBackend` that copies the host's `cached_component_string` (so `check_loaded_fluid()` takes the no-op reuse path and never re-runs `SETUPdll`), carries the correct per-phase composition (`mole_fractions_liq`/`_vap`), and is driven by `update_DmolarT_direct`. The derivative math then reuses the inherited single-phase `first_partial_deriv` machinery.

## Scope / limitations

- **Mixtures:** saturated-state properties and bubble/dew `first_saturation_deriv` are supported. Two-phase *interior* derivatives (`first/second_two_phase_deriv`, `_splined`) require the constant-overall-composition saturation slope, which the per-phase shim cannot supply, so they raise `NotImplementedError` for mixtures rather than return wrong numbers. (Promoting these to a shared `AbstractState` implementation is tracked separately.)
- Pure-fluid two-phase/saturation/spline derivatives are validated against HEOS (`Propane`) to tight tolerance in the new `[REFPROPsat]` tests.

## Dropped from PR #2016 (deliberately)

- Quality-weighted `calc_hmolar`/`calc_smolar` — `update()` already sets these directly from the REFPROP flash.
- `_RPcheckTwophase()` — redundant with `_phase`/`_Q`.
- The `SATPdll`/`SATTdll` `kph` rewiring and the kPa factor-of-1000 edits — they collided with flash logic rewritten in #2865/#2866/#2042/#1845.

## Testing

New `[REFPROPsat]` Catch2 cases (8 cases, 40 assertions) compare REFPROP vs HEOS for saturated outputs and all derivatives, plus the mixture saturation derivative and the mixture-gate behavior. Full `[refprop]` suite (1066 assertions) green — no regressions. clang-format/build/tests/semgrep pass; clang-tidy diff-only clean.

Closes #2013.

## Summary by CodeRabbit

* **New Features**
  * Improved saturation and two‑phase derivative calculations, including a splined two‑phase derivative for smoother interpolation.
  * Construct validated saturated-phase shims for reliable saturated-property queries.
  * Direct state update from (density, temperature) and ability to link to already-loaded fluid setups for faster shim construction.

* **Refactor**
  * Stricter validation and consistency checks for saturated-property retrieval to prevent stale or invalid states.

* **Tests**
  * Extensive tests covering shims, saturated outputs, derivative accuracy and error handling for pure fluids and mixtures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CoolProp/CoolProp/pull/2990?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/Backends/REFPROP/REFPROPMixtureBackend.cpp
+++ b/src/Backends/REFPROP/REFPROPMixtureBackend.cpp
@@ -2362,6 +2362,42 @@ CoolPropDbl REFPROPMixtureBackend::calc_first_saturation_deriv(parameters Of1, p
     }
 }
 
+CoolPropDbl REFPROPMixtureBackend::calc_first_two_phase_deriv(parameters Of, parameters Wrt, parameters Constant) {
+    if (!ValidNumber(_rhoLmolar) || !ValidNumber(_rhoVmolar)) {
+        throw ValueError("The saturation properties are needed for calc_first_two_phase_deriv");
+    }
+    shared_ptr<REFPROPMixtureBackend> SatL = build_saturation_shim(0);
+    shared_ptr<REFPROPMixtureBackend> SatV = build_saturation_shim(1);
+
+    if (Of == iDmolar && Wrt == iHmolar && Constant == iP) {
+        return -POW2(rhomolar()) * (1 / SatV->rhomolar() - 1 / SatL->rhomolar()) / (SatV->hmolar() - SatL->hmolar());
+    } else if (Of == iDmass && Wrt == iHmass && Constant == iP) {
+        return -POW2(rhomass()) * (1 / SatV->rhomass() - 1 / SatL->rhomass()) / (SatV->hmass() - SatL->hmass());
+    } else if (Of == iDmolar && Wrt == iP && Constant == iHmolar) {
+        CoolPropDbl dvdrhoL = -1 / POW2(SatL->rhomolar());
+        CoolPropDbl dvdrhoV = -1 / POW2(SatV->rhomolar());
+        CoolPropDbl dvL_dp = dvdrhoL * SatL->calc_first_saturation_deriv(iDmolar, iP);
+        CoolPropDbl dvV_dp = dvdrhoV * SatV->calc_first_saturation_deriv(iDmolar, iP);
+        CoolPropDbl dhL_dp = SatL->calc_first_saturation_deriv(iHmolar, iP);
+        CoolPropDbl dhV_dp = SatV->calc_first_saturation_deriv(iHmolar, iP);
+        CoolPropDbl dxdp_h = (Q() * dhV_dp + (1 - Q()) * dhL_dp) / (SatL->hmolar() - SatV->hmolar());
+        CoolPropDbl dvdp_h = dvL_dp + dxdp_h * (1 / SatV->rhomolar() - 1 / SatL->rhomolar()) + Q() * (dvV_dp - dvL_dp);
+        return -POW2(rhomolar()) * dvdp_h;
+    } else if (Of == iDmass && Wrt == iP && Constant == iHmass) {
+        CoolPropDbl dvdrhoL = -1 / POW2(SatL->rhomass());
+        CoolPropDbl dvdrhoV = -1 / POW2(SatV->rhomass());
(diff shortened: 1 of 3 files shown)


thorade:
cross-link https://github.com/modelica-3rdparty/ExternalMedia/issues/48

Issue #2685, opened 2026-03-11
CoolProp fails ungracefully when computing entropy through density+internal energy


Rikymark:
Hello,

I am running REFPROP v10.0 via CoolProp 6.7.0 backend on Python 3.13.3.
I am trying to compute the entropy of MM by using density and internal energy as independent pairs. 
I first started with the low-level interface:
"HEOS=CP.AbstractState(model, fluid)
HEOS.update(CP.DmassUmass_INPUTS, Der_LUT[i,0], Der_LUT[i,1])
s_Data[i]=HEOS.smass()*1e-3"
which returns inf for the point (126.64425 kg/m^3, 413304.53 J/kg), while other properties are correctly computed (temperature, enthalpy, pressure).
If I try to compute the entropy for the same point with PropsSI:
"s=PropsSI("S", "Dmass", Der_LUT[i,0], "Umass", Der_LUT[i,1], f"{model}::{fluid}")*1e-3"
I have the error "ValueError: PropsSI failed ungracefully :: inputs were:"S","Dmass",1.2664425000000000e+02,"Umass",4.1330453000000003e+05,"REFPROP::MM" "
The same error also appears if I compute the entropy in another unrelated point, such as half of both previously mentioned density and internal energy.
This does not happen if I do not use REFPROP or if I use other independent pairs, such as (density, temperature) or (density, pressure).
Lastly, I want to add that the REFPROP app computes entropy at the same points without error when density and internal energy are used as inputs.

ibell:
Hi @Rikymark — thanks for the detailed report. Tracing through the code, this looks like a REFPROP-side defect rather than a CoolProp one:

  CoolProp's `DmassUmass_INPUTS` path converts to molar units and calls REFPROP's `DEFLSHdll` (the (D,U) flash routine). At your point, `DEFLSHdll` returns no error code and fills T, p, h correctly, but writes a non-finite value into the entropy slot. CoolProp then propagates that as inf (low-level) or surfaces it as the
  unhelpful "PropsSI failed ungracefully" through the Python wrapper. The REFPROP GUI likely doesn't reach DEFLSH for this query, which is consistent with it working there.

  Please file this with NIST at https://github.com/usnistgov/REFPROP-issues so DEFLSH can be fixed for MM. Minimal reproducer that hits DEFLSH directly with no CoolProp in the loop, using the official ctREFPROP Python wrapper:
```python
  import os
[...]

Issue #2426, opened 2024-10-15
PropsSI with REFPROP backend fails, when using density and internal energy as inputs


mrsoysax:
### Description

When calculating the entropy of Hydrogen using the Dmass and Umass Inputs with the high level interface, using the REFPROP backend, Coolprop fails.

### Steps to Reproduce

1. Define the input values:
    `D_value = 2*10^1` [kg/m^3]
    `U_value = 2.5*10^6` [J/kg]
2. Calculate the entropy using the inputs above and assign the value to the variable "s_calculated"
   `>> s_calculated = py.CoolProp.CoolProp.PropsSI('S','D',D_value,'U',U_value,'REFPROP::Hydrogen')`

**Expected behavior:** Calculate and print the result:
``>> s_calculated = 2.7233*10^6`` [Pa]
**Actual behavior:** Returns the following error:
```
Warning: Unable to load Python object. Saving (serializing) Python objects into a MAT-file is not supported. 
Error using CoolProp/CoolProp.__Props_err2 (line 360)
Python Error: ValueError: PropsSI failed ungracefully :: inputs were:"S","D",2.0000000000000000e+01,"U",2.5000000000000000e+06,"REFPROP::Hydrogen"; please file a
ticket at https://github.com/CoolProp/CoolProp/issues

Error in CoolProp/CoolProp.PropsSI (line 471)

Error in CoolProp/CoolProp.PropsSI (line 391)
```

### Versions

**CoolProp Version:** 6.6.0 
**Operating System and Version:** Windows 10  
**Access Method:** MATLAB Wrapper with python

### Additional Information
- **Different** Dmass and Umass **values** with the same fluid (Hydrogen) have been tested; the **same error occurs**.
- **Different fluids** (water, air) with the same Dmass and Umass values have been tested; the **same error occurs**.
- The low-level version (the _AbstractState_ instance) calculates the "smass" as infinite:
```
RP = py.CoolProp.CoolProp.AbstractState("REFPROP", "Hydrogen")
RP.update(py.CoolProp.CoolProp.DmassUmass_INPUTS, D_value, U_value);
s = RP.smass

>> s = Inf
```

- Following output types were tested and the results were as expected:
**T, P**

- Following output types were tested, no errors were displayed but the results were not as expected:
**hmass**: (Expected: ~3.89\*10^6 [Pa]; Actual: ~2.94\*10^4)

ibell:
[...]

PR #2995, opened 2026-05-25
feat(HEOS): superancillary happy path for D+{H,S,U} flash

ibell:
## Summary

Adds a superancillary **"happy path"** to `FlashRoutines::HSU_D_flash` (density + {H,S,U} flash) for pure fluids, falling through to the legacy ancillary path on any failure (mixtures, pseudo-pure, fluids without a superancillary). Tracked in `bd CoolProp-j3n`.

This is the long-standing flaky family behind #1965, #2154, #2157, #1698, #1054, #2022, #2685, #2426, #2042 (D+{H,S,U} / energy-conservation inputs).

## Robustness — strictly better, no regressions

A/B vs master-equivalent (legacy + superancillaries on, via the new `COOLPROP_DISABLE_SUPERANC_HSU_D` kill-switch), over a dense (log-p, linear-T) sweep across **every** superancillary fluid (~5.5M states):

- **Regressions (fail happy / pass legacy): 0**
- **Fixes (fail legacy / pass happy): 2,839** across 54 fluids
- The happy path's residual failure set is a strict **subset** of legacy's.

The happy path is **self-validating** (`committed_ok` gate): it never *returns* a state that fails to reproduce the inputs; on failure it restores the caller's inputs and falls through to legacy.

## Design

- **1D-in-T two-phase residual** reads saturated ρ off the superancillary (`eval_sat`), replacing the legacy inner 2D `saturation_D_pure` solve.
- **Interval classification by dome membership** (not root count) handles the ρ≈ρ_crit collapse and the water/heavy-water liquid **density anomaly** uniformly; never hands the single-phase solver a dome-interior bracket.
- Two-phase quality clamped at the saturation boundary; spurious single-phase crossings rejected when the implied quality leaves [0,1].

## Performance

The two-phase residual reads saturated h/s/u straight off the **caloric superancillary** (~150 ns/call vs ~2.7 µs with EOS `keyed_output` — profiled). A short **full-EOS secant polish** (default on, new config `HSU_D_TWOPHASE_EOS_POLISH`) cleans the ~1e-8 superancillary deviation to EOS-exact.

Two-phase D+S / D+U (Release, polished, vs legacy): **~2.7× / ~2.5×**; near-critical-density single-phase ~3× (avoids legacy's `saturation_D_pure` ω-halving retry loop). With the polish off: ~8×.

A **native U caloric superancillary** is now built alongside H and S (`ensure_HSU_under_lock`; storage and the `'U'` build path already existed). U shares H's reference-frame shift (`u = h − p/ρ`, p/ρ offset-independent), so all three of D+{H,S,U} take the same uniform fast path — a **net simplification** (removes the U-specific gate and the `h − p/ρ` reconstruction) that brings D+U up to H/S speed. Runtime-built like H/S; **no `fastchebpure`/JSON regeneration**.

Escape hatch: `COOLPROP_DISABLE_SUPERANC_HSU_D` forces the legacy path.

## Tests

`src/Tests/CoolProp-Tests-HSU_D.cpp` (tag `[HSU_D]`): characterization suite mined from the issues above, water density-anomaly cases, polish-config toggle, plus opt-in `[HSU_D_ptsweep]` (dense all-fluid sweep, env-tunable grid) and `[HSU_D_bench]`/`[HSU_D_prof]` timing/profiling.

## Summary by CodeRabbit

* **New Features**
  * Improved HSU_D flash: faster, more robust density + enthalpy/entropy/internal‑energy solves for pure fluids with an optional two‑phase EOS polish (new configuration toggle, default: true). Caloric ancillary handling consolidated for better stability and accuracy.

* **Tests**
  * Added comprehensive regression and benchmark suite covering saturation boundaries, two‑phase interiors, near‑critical regions, fluid‑specific edge cases, and performance comparisons.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CoolProp/CoolProp/pull/2995?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/Backends/Helmholtz/FlashRoutines.cpp
+++ b/src/Backends/Helmholtz/FlashRoutines.cpp
@@ -2,6 +2,8 @@
 #include "FlashRoutines.h"
 
 #include <cmath>
+#include <algorithm>
+#include <cstdlib>
 #include "CoolProp.h"
 #include "HelmholtzEOSMixtureBackend.h"
 #include "HelmholtzEOSBackend.h"
@@ -437,6 +439,8 @@ static char param_to_superanc_key(parameters key) {
             return 'H';
         case iSmolar:
             return 'S';
+        case iUmolar:
+            return 'U';
         default:
             throw ValueError(format("Unsupported parameters key for saturation superancillary: %d", static_cast<int>(key)));
     }
@@ -480,7 +484,7 @@ double FlashRoutines::resolve_T_via_superancillary(HelmholtzEOSMixtureBackend& H
         throw ValueError(format("%s currently requires Q=0 or Q=1; got Q=%g", fn_name, Q));
     }
     const char k = param_to_superanc_key(key);
-    if (k == 'H' || k == 'S') {
+    if (k == 'H' || k == 'S' || k == 'U') {
         HEOS.ensure_caloric_superancillaries();
     }
     auto& superanc = *superanc_ptr;
@@ -499,14 +503,15 @@ double FlashRoutines::resolve_T_via_superancillary(HelmholtzEOSMixtureBackend& H
     // ensure_HS_under_lock comment and #2773. For D the cache is
     // reference-state-independent intrinsically (no shift needed).
     double target_in_cache_frame = target_value;
-    if (k == 'H' || k == 'S') {
+    if (k == 'H' || k == 'S' || k == 'U') {
         const auto stamp = superanc.get_caloric_alpha0_stamp();
         if (stamp.has_value()) {
             const auto [caller_a1, caller_a2] = alpha0_offset_total(HEOS);
             const double cache_a1 = stamp->first;
             const double cache_a2 = stamp->second;
-            if (k == 'H') {
-                // h_cache(T) = h_caller(T) + R·T_red·(a2_cache − a2_caller)
+            if (k == 'H' || k == 'U') {
+                // h_cache(T) = h_caller(T) + R·T_red·(a2_cache − a2_caller); U
+                // inherits the H shift since u = h − p/rho and p/rho is offset-free.
                 const double R = HEOS.gas_constant();
                 const double T_red = HEOS.get_reducing_state().T;
                 target_in_cache_frame = target_value + R * T_red * (cache_a2 - caller_a2);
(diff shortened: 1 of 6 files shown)


src/Backends/Helmholtz/FlashRoutines.cpp
```diff
@@ -1444,6 +1520,281 @@ void FlashRoutines::HSU_D_flash(HelmholtzEOSMixtureBackend& HEOS, parameters oth
...
+        // H/S/U, use the superancillary saturation-density roots to robustly
+        // decide single- vs two-phase and to bracket the temperature solve.
+        // This avoids the saturation_D_pure instability that makes the legacy
+        // routine fail near the critical density, and clamps the quality at the
+        // saturation boundary.  Any failure here falls through to the legacy
+        // ancillary "sad path" below.  See CoolProp-j3n (ported from saD_HSU).
+        //
+        // Escape hatch: COOLPROP_DISABLE_SUPERANC_HSU_D (read once) forces the
+        // legacy path -- a field kill-switch and an A/B regression-test handle.
+        static const bool superanc_hsu_d_disabled = (std::getenv("COOLPROP_DISABLE_SUPERANC_HSU_D") != nullptr);
```
ibell, line 1535: If this even necessary if the disable-via-config is active?

src/Backends/Helmholtz/FlashRoutines.cpp
```diff
@@ -1444,6 +1520,281 @@ void FlashRoutines::HSU_D_flash(HelmholtzEOSMixtureBackend& HEOS, parameters oth
...
+                            double Qd_final;
+                            if (use_ca && get_config_bool(HSU_D_TWOPHASE_EOS_POLISH)) {
+                                solver_resid_2phase resid_eos(HEOS, HEOS._rhomolar, other, value, superanc, /*use_ca=*/false, ca_key, value_cache);
+                                double t0 = Tsol, r0 = resid_eos.call(t0);
+                                // Probe just above Tsol, but keep the second secant seed
+                                // strictly inside the two-phase bracket [Ta, Tb]: when the
+                                // raw solve lands on a boundary root, Tsol*(1+1e-7) can step
+                                // past Tb and make resid_eos.call() fail, needlessly kicking
+                                // a valid happy-path state back to the legacy path.
+                                double t1 = Tsol * (1.0 + 1e-7);
```
ibell, line 1694: Should this be a Newton step with analytic derivatives?

src/Backends/Helmholtz/FlashRoutines.cpp
```diff
@@ -1444,6 +1520,281 @@ void FlashRoutines::HSU_D_flash(HelmholtzEOSMixtureBackend& HEOS, parameters oth
...
+                            } else {
+                                return false;
+                            }
+                            // EOS polish: the fast path read the saturated caloric
+                            // values off the superancillary, which carries a ~1e-8
+                            // deviation from the EOS.  T_sol is an excellent seed, so a
+                            // couple of secant steps on the EOS-caloric residual (full
+                            // EOS h/s/u of each phase) clean the solution up to EOS
+                            // precision at a few EOS evaluations -- vs the ~12 the fast
+                            // path saved.  U already uses the EOS, so nothing to polish.
```
ibell, line 1681: This comment is wrong about U

ibell:
@coderabbitai Re: the secant bracket-clamping suggestion at `FlashRoutines.cpp` (keep later iterates inside `[Ta, Tb]`) — **not applied; verified it regresses behavior.** I implemented the proposed `|| tn <= Ta || tn >= Tb` break and it broke 13 assertions in the #2486 saturation-boundary characterization cases (n-Propane, R134a), dropping previously-passing happy-path solves to the legacy path (which then throws `saturation_D_pure solver abs error`).

Why the premise doesn't hold here:

1. **The polish residual is valid far outside `[Ta, Tb]`.** `solver_resid_2phase::call(T)` reads saturated properties off `superanc.eval_sat(T, …)`, which is valid across the *entire* saturation curve (T_triple…T_crit). It does not throw outside `[Ta, Tb]`; it only throws on a non-finite residual, which is already guarded by the `std::isfinite(r1)` check / the `isfinite(tn)` break. So "the next `resid_eos.call(t1)` will throw" is not the case for this residual.

2. **`[Ta, Tb]` is the single-phase solver's shy-of-dome bracket, not a hard domain for the caloric polish.** Those bounds are deliberately set "just shy of" the dome to keep the *single-phase* solver out of the dome interior. For saturation-boundary inputs the EOS-exact saturation T sits right at / just outside that shy margin, so constraining the secant to `(Ta, Tb)` actively prevents convergence to the correct root.

3. **No UB / wrong-result risk to harden against.** The path self-validates via `committed_ok()`; a bad polish would at worst fall back to legacy, never return a wrong state.

Leaving the secant unconstrained (current code) is correct. Closing this one as won't-fix with the above rationale.

Issue #2926, opened 2026-05-16
clang-tidy sweep: actionable findings from files modified in the last month


ibell:
Ran `clang-tidy` (LLVM 21, project's `.clang-tidy` config) against the 55 `.cpp` files in `src/` modified in the last month. This is a triage report — no code changes proposed in this issue, just a categorized list of what looked actionable vs. noise.

## Tooling note

Homebrew clang-tidy on macOS needs `-isysroot $(xcrun --show-sdk-path) -isystem .../c++/v1` extra args, otherwise the C++ standard headers don't resolve and semantic analysis silently degrades — `bugprone-infinite-loop`, `bugprone-virtual-near-miss`, and `bugprone-chained-comparison` each emitted false positives that vanished once headers were found. The CI workflow (`.github/workflows/dev_checks.yml`, clang-tidy-18 on ubuntu-latest) is unaffected. Worth a `README`/CONTRIBUTING note for anyone reproducing locally on macOS.

---

### Dead stores (`clang-analyzer-deadcode.DeadStores`)
- `src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp:1647` — `p_crit_ = p_critical()` is never read; the assignment also wastes a critical-pressure call. Drop from the joint declaration.
- `src/Region/PiecewiseChebyshevCurve.cpp:142, 148` — `f_hi` is written but never read inside the bisection at lines 143–156 (only `f_lo` participates). Variable can be removed entirely.

### Virtual call in constructor (`clang-analyzer-optin.cplusplus.VirtualCall`) — 10 sites
Bypasses virtual dispatch; latent bug if a subclass overrides:
- `include/AbstractState.h:725` (`clear()`)
- `src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp:84, 95, 111, 112, 123, 127` (`set_components`, `calc_reducing_state_nocache`, `calc_gas_constant`, `get_copy`)
- `src/Backends/Incompressible/IncompressibleBackend.cpp:177` (`clear`)
- `src/Backends/PCSAFT/PCSAFTBackend.cpp:112, 114` (`get_copy`)

Standard fix is a non-virtual `init_*()` helper invoked from each ctor.

### Unbounded `strcpy` (`clang-analyzer-security.insecureAPI.strcpy`)
[...]

PR #3019, opened 2026-05-27
refactor(clang-tidy): Tier-2 mechanical fix-it sweep (#2926 layer 2)

ibell:
## What

Mechanical clang-tidy fix-it sweep — the leftover modernize/perf batch from the **layer-2 triage** (CoolProp-1pr) that #2980 did not run. Pairs with #3018 (Tier-1). 70 files, behavior-preserving.

## Checks applied

Override the `.clang-tidy` whitelist for this run:

| Check | Effect |
|---|---|
| `performance-unnecessary-value-param` | pass containers/strings by `const&` |
| `cppcoreguidelines-prefer-member-initializer` | ctor-body assign → init list |
| `modernize-use-auto` | `auto` for `new` / iterators |
| `modernize-loop-convert` | range-based `for` |
| `modernize-use-equals-default` | `{}` → `= default` |
| `modernize-use-using` | `typedef` → `using` |
| `modernize-return-braced-init-list`, `-raw-string-literal`, `-use-emplace`, `performance-unnecessary-copy-initialization` | small mechanical |

## How it was run (safely)

The naïve `run-clang-tidy -fix` approach has two footguns that I hit and worked around:
- **Anchored header-filter** `^<repo>/(include|src)/` — an unanchored `(include|src)` matches `build_catch/_deps/eigen-src/Eigen/src/` and `_deps/rapidjson-src/include/`, corrupting third-party trees.
- **Per-file sequential `clang-tidy -fix`** (not `run-clang-tidy`'s export+apply) — the YAML-merge path double-applies the same shared-header constructor fix from multiple TUs (`: a(a) : a(a) {`). Sequential application fixes each header once.

Formatted with the pinned clang-format 18.1.8.

## Caught-and-reverted regression (⚠️ adversarial review value)

The `prefer-member-initializer` fix wrongly hoisted `N = n.size(); s = n;` into the `SurfaceTensionCorrelation(rapidjson::Value&)` member-init list (`include/Ancillaries.h`). `n` is populated in the constructor **body**, so at init-list time it's empty → `N=0`, `s={}` → `evaluate()` silently returns **0.0 for every fluid's surface tension**. The full test suite missed it because **no surface-tension test existed**.

Fix: that one constructor assigns `N`/`s` in the body (NOLINT + explanatory comment), and a new `[surface_tension]` regression test checks water σ against IAPWS at 300/350 K (and would fail loudly on the `0.0` regression).

## Verification

- Full `~[slow]~[benchmark]` suite: **298 cases, 117,586 assertions, 0 failures**.
- Per-check confirmation: swept files drop to ~0 remaining Tier-2 findings.
- Adversarial review (two passes): the one regression found and fixed; all other hoisted init-lists confirmed to read only params/constants, not later-assigned siblings.
- Whole-file clang-tidy/cppcheck in preflight skipped (pre-existing other-category findings in these legacy files; this PR is itself the clang-tidy cleanup); clang-format + semgrep + tests pass.

## Follow-up

- Recommend adding the squashed commit to `.git-blame-ignore-revs` (mechanical sweep), matching #2803/#2805/#2808/#2980.
- Vendored `include/superancillary/superancillary.h` carries trivial `=default`/`auto` fixes; the same should go upstream to avoid a re-vendor clobber.

## Summary by CodeRabbit

* **Refactor**
  * Broad modernization across the codebase: cleaner initialization, defaulted special members, safer parameter passing, fewer unnecessary copies, and updated iteration/alias styles to improve stability and maintainability.
* **Tests**
  * Test suite cleaned up for reliability and readability; new regression validating nonzero surface tension behavior for water added.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CoolProp/CoolProp/pull/3019?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/Backends/Cubics/GeneralizedCubic.h
+++ b/src/Backends/Cubics/GeneralizedCubic.h
@@ -618,7 +618,7 @@ class PengRobinson : public AbstractCubic
     PengRobinson(const std::vector<double>& Tc, std::vector<double> pc, std::vector<double> acentric, double R_u,
                  const std::vector<double>& C1 = std::vector<double>(), const std::vector<double>& C2 = std::vector<double>(),
                  const std::vector<double>& C3 = std::vector<double>())
-      : AbstractCubic(std::move(Tc), std::move(pc), std::move(acentric), R_u, 1 + sqrt(2.0), 1 - sqrt(2.0), C1, C2, C3) {
+      : AbstractCubic(Tc, std::move(pc), std::move(acentric), R_u, 1 + sqrt(2.0), 1 - sqrt(2.0), C1, C2, C3) {
         set_alpha(C1, C2, C3);
     };
 
@@ -638,7 +638,7 @@ class SRK : public AbstractCubic
     SRK(const std::vector<double>& Tc, std::vector<double> pc, std::vector<double> acentric, double R_u,
         const std::vector<double>& C1 = std::vector<double>(), const std::vector<double>& C2 = std::vector<double>(),
         const std::vector<double>& C3 = std::vector<double>())
-      : AbstractCubic(std::move(Tc), std::move(pc), std::move(acentric), R_u, 1, 0, C1, C2, C3) {
+      : AbstractCubic(Tc, std::move(pc), std::move(acentric), R_u, 1, 0, C1, C2, C3) {
         set_alpha(C1, C2, C3);
     };
     SRK(double Tc, double pc, double acentric, double R_u)
--- a/src/Backends/Cubics/GeneralizedCubic.h
+++ b/src/Backends/Cubics/GeneralizedCubic.h
@@ -11,6 +11,7 @@
 #ifndef CUBIC_H
 #define CUBIC_H
 
+#include <utility>
 #include <vector>
 #include <cmath>
 #include <memory>
@@ -26,7 +27,7 @@ class AbstractCubicAlphaFunction
       sqrt_Tr_Tci;          ///< The sqrt of the (constant) reducing temperature divided by the critical temperature of the pure component
     std::vector<double> c;  ///< The vector of constants
    public:
-    virtual ~AbstractCubicAlphaFunction() {};
+    virtual ~AbstractCubicAlphaFunction() = default;
     virtual double term(double tau, std::size_t itau) = 0;
     void set_Tr_over_Tci(double Tr_over_Tci) {
         this->Tr_over_Tci = Tr_over_Tci;
@@ -40,9 +41,10 @@ class BasicMathiasCopemanAlphaFunction : public AbstractCubicAlphaFunction
 {
     double m;  ///< The term coming from the function of omega
    public:
-    BasicMathiasCopemanAlphaFunction(double a0, double m_ii, double Tr_over_Tci) : AbstractCubicAlphaFunction(a0, Tr_over_Tci) {
-        this->m = m_ii;
-    };
+    BasicMathiasCopemanAlphaFunction(double a0, double m_ii, double Tr_over_Tci)
+      : AbstractCubicAlphaFunction(a0, Tr_over_Tci), m(m_ii) {
+
+        };
     double term(double tau, std::size_t itau) override;
 };
 
@@ -98,10 +100,10 @@ class AbstractCubic
      so long as it has the formulation given in this work.
 
      */
-    AbstractCubic(std::vector<double> Tc, std::vector<double> pc, std::vector<double> acentric, double R_u, double Delta_1, double Delta_2,
-                  std::vector<double> C1 = std::vector<double>(), std::vector<double> C2 = std::vector<double>(),
-                  std::vector<double> C3 = std::vector<double>());
-    virtual ~AbstractCubic() {};
+    AbstractCubic(const std::vector<double>& Tc, std::vector<double> pc, std::vector<double> acentric, double R_u, double Delta_1, double Delta_2,
+                  const std::vector<double>& C1 = std::vector<double>(), const std::vector<double>& C2 = std::vector<double>(),
+                  const std::vector<double>& C3 = std::vector<double>());
+    virtual ~AbstractCubic() = default;
     /// Set the constants for the Mathias-Copeman alpha function, or if C1,C2,C3 are all empty, set the default alpha model
     void set_alpha(const std::vector<double>& C1, const std::vector<double>& C2, const std::vector<double>& C3);
     /// Set the alpha function for the i-th component
@@ -613,10 +615,10 @@ class AbstractCubic
 class PengRobinson : public AbstractCubic
 {
    public:
-    PengRobinson(std::vector<double> Tc, std::vector<double> pc, std::vector<double> acentric, double R_u,
-                 std::vector<double> C1 = std::vector<double>(), std::vector<double> C2 = std::vector<double>(),
-                 std::vector<double> C3 = std::vector<double>())
-      : AbstractCubic(Tc, pc, acentric, R_u, 1 + sqrt(2.0), 1 - sqrt(2.0), C1, C2, C3) {
+    PengRobinson(const std::vector<double>& Tc, std::vector<double> pc, std::vector<double> acentric, double R_u,
+                 const std::vector<double>& C1 = std::vector<double>(), const std::vector<double>& C2 = std::vector<double>(),
+                 const std::vector<double>& C3 = std::vector<double>())
+      : AbstractCubic(std::move(Tc), std::move(pc), std::move(acentric), R_u, 1 + sqrt(2.0), 1 - sqrt(2.0), C1, C2, C3) {
         set_alpha(C1, C2, C3);
     };
 
@@ -633,9 +635,10 @@ class PengRobinson : public AbstractCubic
 class SRK : public AbstractCubic
 {
    public:
-    SRK(std::vector<double> Tc, std::vector<double> pc, std::vector<double> acentric, double R_u, std::vector<double> C1 = std::vector<double>(),
-        std::vector<double> C2 = std::vector<double>(), std::vector<double> C3 = std::vector<double>())
-      : AbstractCubic(Tc, pc, acentric, R_u, 1, 0, C1, C2, C3) {
+    SRK(const std::vector<double>& Tc, std::vector<double> pc, std::vector<double> acentric, double R_u,
+        const std::vector<double>& C1 = std::vector<double>(), const std::vector<double>& C2 = std::vector<double>(),
+        const std::vector<double>& C3 = std::vector<double>())
+      : AbstractCubic(std::move(Tc), std::move(pc), std::move(acentric), R_u, 1, 0, C1, C2, C3) {
         set_alpha(C1, C2, C3);
     };
     SRK(double Tc, double pc, double acentric, double R_u)
--- a/src/Backends/Helmholtz/Fluids/FluidLibrary.cpp
+++ b/src/Backends/Helmholtz/Fluids/FluidLibrary.cpp
@@ -45,7 +45,7 @@ static void ensure_library_loaded() {
 
 void load() {
     std::vector<unsigned char> outbuffer(gall_fluids_JSON_zSize * 7);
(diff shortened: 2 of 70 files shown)


src/Backends/PCSAFT/PCSAFTBackend.cpp
```diff
@@ -218,15 +212,15 @@ void PCSAFTBackend::set_mass_fractions(const std::vector<CoolPropDbl>& mass_frac
         sum_moles += tmp;
     }
     std::vector<CoolPropDbl> mole_fractions;
-    for (std::vector<CoolPropDbl>::iterator it = moles.begin(); it != moles.end(); ++it) {
-        mole_fractions.push_back(*it / sum_moles);
+    for (double& mole : moles) {
+        mole_fractions.push_back(mole / sum_moles);
```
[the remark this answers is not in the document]
ibell: Good catch — fixed in 08c5047. `loop-convert` baked in `double` from this build's default `COOLPROPDBL_MAPS_TO_DOUBLE`, which breaks the `long double` config. I broadened the fix to all five `CoolPropDbl`-backed range loops (the same defect was also in `FluidLibrary.h` — a mutating `double&` that would be a hard compile error under `long double` — and three by-value loops in `IncompressibleBackend.cpp` that would narrow). All now use `auto`. Genuine `std::vector<double>` loops were left unchanged.

src/Backends/Helmholtz/VLERoutines.h
```diff
@@ -32,10 +32,16 @@ struct saturation_D_pure_options
...
+    saturation_D_pure_options()
+      : use_guesses(false),
+        use_logdelta(true),
+        omega(1.0),
+        rhoL(_HUGE),
+        rhoV(_HUGE),
+        pL(_HUGE),
+        pV(_HUGE),
+        imposed_rho(0),
+        max_iterations(200) {}  // Defaults
```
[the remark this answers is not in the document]
ibell: This is a false positive — `use_guesses` was never a constructor argument. On `origin/master` this is a parameterless default ctor with `use_guesses(false)` already hardcoded in the init list:

```cpp
saturation_D_pure_options() : use_guesses(false), rhoL(_HUGE), ... { use_logdelta = true; omega = 1.0; }
```

Callers configure `use_guesses` by field assignment after construction (e.g. `options.use_guesses = true;`), not via a ctor parameter. This PR only moved `use_logdelta`/`omega` from the body into the init list (prefer-member-initializer); no parameter was removed and there's no source-compatibility change. Leaving as-is.

src/Backends/Cubics/GeneralizedCubic.h
```diff
@@ -613,10 +615,10 @@ class AbstractCubic
...
 {
    public:
-    PengRobinson(std::vector<double> Tc, std::vector<double> pc, std::vector<double> acentric, double R_u,
-                 std::vector<double> C1 = std::vector<double>(), std::vector<double> C2 = std::vector<double>(),
-                 std::vector<double> C3 = std::vector<double>())
-      : AbstractCubic(Tc, pc, acentric, R_u, 1 + sqrt(2.0), 1 - sqrt(2.0), C1, C2, C3) {
+    PengRobinson(const std::vector<double>& Tc, std::vector<double> pc, std::vector<double> acentric, double R_u,
+                 const std::vector<double>& C1 = std::vector<double>(), const std::vector<double>& C2 = std::vector<double>(),
+                 const std::vector<double>& C3 = std::vector<double>())
+      : AbstractCubic(std::move(Tc), std::move(pc), std::move(acentric), R_u, 1 + sqrt(2.0), 1 - sqrt(2.0), C1, C2, C3) {
```
ibell: Should this be a move? I think a copy is needed

src/Backends/Cubics/GeneralizedCubic.h
```diff
@@ -633,9 +635,10 @@ class PengRobinson : public AbstractCubic
 class SRK : public AbstractCubic
 {
    public:
-    SRK(std::vector<double> Tc, std::vector<double> pc, std::vector<double> acentric, double R_u, std::vector<double> C1 = std::vector<double>(),
-        std::vector<double> C2 = std::vector<double>(), std::vector<double> C3 = std::vector<double>())
-      : AbstractCubic(Tc, pc, acentric, R_u, 1, 0, C1, C2, C3) {
+    SRK(const std::vector<double>& Tc, std::vector<double> pc, std::vector<double> acentric, double R_u,
+        const std::vector<double>& C1 = std::vector<double>(), const std::vector<double>& C2 = std::vector<double>(),
+        const std::vector<double>& C3 = std::vector<double>())
+      : AbstractCubic(std::move(Tc), std::move(pc), std::move(acentric), R_u, 1, 0, C1, C2, C3) {
```
ibell: I think a copy is needed for Tc and others

src/Backends/Helmholtz/Fluids/FluidLibrary.cpp
```diff
@@ -304,13 +304,13 @@ void JSONFluidLibrary::add_one(rapidjson::Value& fluid_json) {
             index = string_to_index_map.find(upper(fluid.name))->second;  // if uppercase name found, grab index
         } else {
             // Check the aliases
-            for (std::size_t i = 0; i < fluid.aliases.size(); ++i) {
-                if (string_to_index_map.find(fluid.aliases[i]) != string_to_index_map.end()) {
-                    index = string_to_index_map.find(fluid.aliases[i])->second;  // if alias found, grab index
+            for (const auto& aliase : fluid.aliases) {
```
ibell: variable name should be `alias`

src/Backends/Tabular/TabularBackends.h
```diff
@@ -1074,22 +1055,28 @@ class TabularBackend : public AbstractState
...
+        z(nullptr),
+        dzdx(nullptr),
+        dzdy(nullptr),
+        d2zdx2(nullptr),
+        d2zdxdy(nullptr),
+        d2zdy2(nullptr),
+        AS(std::move(AS)),
+        dataset(nullptr) {
+
+            // Flush the cached indices (set to large number)
```
ibell: is this comment correct?

src/Helmholtz.cpp
ibell: should this be now a `constexpr vector<string>`

src/Tests/CoolProp-Tests.cpp
```diff
@@ -1474,23 +1468,23 @@ TEST_CASE("Test consistency between Gernert models in CoolProp and Gernert model
     Skip_if_No_REFPROP();  // Skip this test if REFPROPMixture backend is not available
 
     std::string mixes[] = {"CO2[0.7]&Argon[0.3]", "CO2[0.7]&Water[0.3]", "CO2[0.7]&Nitrogen[0.3]"};
-    for (int i = 0; i < 3; ++i) {
-        const char* ykey = mixes[i].c_str();
+    for (const auto& mixe : mixes) {
```
ibell: "mix" not "mixe"

src/Tests/CoolProp-Tests.cpp
```diff
@@ -1564,8 +1558,8 @@ TEST_CASE("Tests for solvers in P,Y flash using Water", "[flash],[PH],[PS],[PU]"
     double Ts, y, T2;
     // See https://groups.google.com/forum/?fromgroups#!topic/catch-forum/mRBKqtTrITU
     std::string Ykeys[] = {"H", "S", "U", "Hmass", "Smass", "Umass", "Hmolar", "Smolar", "Umolar"};
```
ibell: prolly vector<string> now

src/Tests/CoolProp-Tests.cpp
```diff
@@ -2278,22 +2272,22 @@ TEST_CASE("Test that reference states yield proper values using high-level inter
     };
     std::string fluids[] = {"n-Propane", "R134a", "R124"};
     ref_entry entries[3] = {{"IIR", 200000, 1000, "T", 273.15, "Q", 0}, {"ASHRAE", 0, 0, "T", 233.15, "Q", 0}, {"NBP", 0, 0, "P", 101325, "Q", 0}};
-    for (std::size_t i = 0; i < 3; ++i) {
-        for (std::size_t j = 0; j < 3; ++j) {
+    for (const auto& fluid : fluids) {
+        for (auto& entrie : entries) {
```
ibell: singular is "entry"

src/Tests/CoolProp-Tests.cpp
```diff
@@ -2310,28 +2304,28 @@ TEST_CASE("Test that reference states yield proper values using low-level interf
     };
     std::string fluids[] = {"n-Propane", "R134a", "R124"};
     ref_entry entries[3] = {{"IIR", 200000, 1000, iT, 273.15, iQ, 0}, {"ASHRAE", 0, 0, iT, 233.15, iQ, 0}, {"NBP", 0, 0, iP, 101325, iQ, 0}};
-    for (std::size_t i = 0; i < 3; ++i) {
-        for (std::size_t j = 0; j < 3; ++j) {
+    for (const auto& fluid : fluids) {
+        for (auto& entrie : entries) {
```
ibell: singular is "entry"

ibell (commented):
nits

src/Backends/Cubics/GeneralizedCubic.h
ibell: Done in 777cec434. The `std::move(Tc)` here was a no-op — Tc is `const std::vector<double>&`, so std::move casts to const-rvalue which can't bind to a non-const `&&` and silently degrades to a copy. AbstractCubic stores Tc as `Tc(Tc)` (member copy from a const& param) regardless. Dropped `std::move(Tc)` and kept `std::move(pc)` / `std::move(acentric)` — those params are by-value so the move is real. Same fix applied to SRK and VTPRCubic.

src/Backends/Cubics/GeneralizedCubic.h
ibell: Done in 777cec434 — same fix as on PengRobinson. `std::move(Tc)` was the broken one (Tc is const&, the move silently no-ops to a copy). pc/acentric stayed moved because they're by-value params. VTPRCubic got the same treatment.

src/Backends/Helmholtz/Fluids/FluidLibrary.cpp
ibell: Renamed in 777cec434.

src/Backends/Tabular/TabularBackends.h
ibell: It wasn't — the cached_*_i/j members are now initialized in the member-init list (lines 1064–1067), so the body was empty and the comment was pointing at code that no longer existed there. Removed the orphan comment in 777cec434.

src/Helmholtz.cpp
ibell: Converted to `const std::vector<std::string>` in 777cec434. `constexpr std::vector<std::string>` is C++20 (and even there has nasty lifetime restrictions because of the allocator); the project is C++17, so plain `const` is the practical match. The inner loop is already range-based so the iteration code didn't change.

src/Tests/CoolProp-Tests.cpp
ibell: Renamed in 777cec434.

src/Tests/CoolProp-Tests.cpp
ibell: Converted `Ykeys` to `const std::vector<std::string>` in 777cec434 — same C++17 caveat as on the Helmholtz.cpp `derivs` thread (not constexpr-able).

src/Tests/CoolProp-Tests.cpp
ibell: Renamed both occurrences (this block and the duplicate at L2308 in the low-level interface block) in 777cec434.

src/Tests/CoolProp-Tests.cpp
ibell: Renamed in 777cec434 (in the same commit as the high-level interface variant above).

Issue #3023, opened 2026-05-28
Docs fail to build due to phase determination problem for `Air.mix`


ibell:
Docs fail to build due to phase determination problem for `Air.mix`:
```
----> 1 CP.PropsSI('D','P',101325,'T',300,'Air.mix')

File CoolProp/CoolProp.pyx:498, in CoolProp.CoolProp.PropsSI()

File CoolProp/CoolProp.pyx:588, in CoolProp.CoolProp.PropsSI()

File CoolProp/CoolProp.pyx:465, in CoolProp.CoolProp.__Props_err2()

ValueError: phase is invalid in calc_fugacity : PropsSI("D","P",101325,"T",300,"Air.mix")

<<<-------------------------------------------------------------------------
```

ibell:
Resolved by #3022 ('extend two-phase awareness to chemical_potential and fugacity_coefficient', merged as 2a3f1995e).

Root cause: PR #2345 added a phase-dispatch guard to `HEOS::calc_fugacity` that throws when `_phase` is neither two-phase nor a classified homogeneous phase. The mixture PT flash runs `StabilityRoutines::check_stability`, which evaluates the fugacity family on a *bulk trial state* built via `update_DmolarT_direct` — a call that does not reset `_phase` (it can be left `iphase_unknown`, or stale `iphase_twophase` from a prior flash). The guard then threw 'phase is invalid in calc_fugacity'.

#3022 fixes this at the call site: `check_stability` now calls `MixtureDerivatives::ln_fugacity_coefficient`/`fugacity_i` directly (the same pattern already used in the tpd accumulation loop just below), bypassing the phase dispatch to get the true bulk-composition values. It also extends two-phase awareness to the sibling functions (`chemical_potential` Q-weighted across SatL/SatV; `fugacity_coefficient` throws in two-phase since φ_i is phase-specific).

Verified: `PropsSI('Dmolar','P',101325,'T',300,'Air.mix')` now returns a valid number; the existing `[predefined_mixtures]` regression test passes on master.

Issue #2637, opened 2025-10-14
Issues calculating critical points and phase envelope for hydrocarbons


joshkel:
I'm trying to calculate critical points and phase envelopes for some hydrocarbon mixtures and am having problems.

Very simple case:

```py
from CoolProp import CoolProp
print(CoolProp.PropsSI('pcrit', '', 0, '', 0, 'Methane[0.85]&Ethane[0.15]'))
```

Output:

> No outputs were able to be calculated : PropsSI("pcrit","",0,"",0,"Methane[0.85]&Ethane[0.15]")

This looks like the same issue as #2175, so I tried using the low-level interface:

```py
from CoolProp import CoolProp

heos = CoolProp.AbstractState('HEOS', 'Ethane&Methane')
heos.set_mole_fractions([0.15, 0.85])

print(heos.build_phase_envelope(''))
print(heos.all_critical_points())
```

The resulting phase envelope is `None`, and `all_critical_points` throws the following error.

> ValueError: Inputs in Brent [30165.854237,51571.973776] do not bracket the root.  Function values are [-2.906478,-148.094469]

For a more complex hydrocarbon mixture (which I can share if it would be helpful), `build_phase_envelope` seems to never finish, and `all_critical_points` thinks for 2 minutes then throws the following error:

> ValueError: Brent's method f(b) is NAN for b = 1, other input was a = 0

Any advice?

Issue #2308, opened 2023-10-20
Issue calculating molar fractions of every component out the biphase zone for a custom fluid modelled with AbstractState


vins1996:
### Description

I am currently using Coolprop library on a python code. For my research I modeled a fluid with the function AbstractState, which is a gas refrigerant mixture (nitrogen, methane, ethane and propane). When calculting the molar fractions of every single component out of the biphase zone with the function fluidname.mole_fractions_liquid() or  fluidname.mole_fractions_vapor(), the code returns values different from zero for both.

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behavior:** 
I expect that is vapor zone, molar fractions of the liquid part of the fluid is returned with zero value

**Actual behavior:**
In my case the value is not zero and neither is negligible.

### Versions

**CoolProp Version:** 6.5.0
**Operating System and Version:** [WSL with ubunthu 22.04.2 LTS]  
**Access Method:** [Python3]

ibell:
For homogeneous state points, neither the liquid nor vapor results should
be possible. Please provide a runnable example.

ibell:
I rewrote in Python:
```python
import CoolProp.CoolProp as CP 

AS = CP.AbstractState('HEOS', 'Nitrogen&Methane&Ethane&Propane')
AS.set_mole_fractions([0.10, 0.34, 0.41, 0.15])
AS.build_phase_envelope("")
Tmedia=[-18.79070798,  -22.37212394,  -25.9535399,   -29.53495586,  -33.11637181,
  -36.69778777,  -40.27920373,  -43.86061969,  -47.44203565,  -51.02345161,
  -54.60486757,  -58.18628352,  -61.76769948,  -65.34911544,  -68.9305314,
  -72.51194736,  -76.09336332,  -79.67477928,  -83.25619524,  -86.83761119,
  -90.41902715,  -94.00044311,  -97.58185907, -101.16327503, -104.74469099,
 -108.32610695, -111.9075229,  -115.48893886, -119.07035482, -122.65177078,
 -126.23318674, -129.8146027,  -133.39601866, -136.97743462, -140.55885057,
 -144.14026653, -147.72168249, -151.30309845, -154.88451441, -158.46593037]

for i in range(len(Tmedia)):
    AS.update(CP.PT_INPUTS, 1.5e5, Tmedia[i]+273.15)
[...]

Issue #2262, opened 2023-06-20
ValueError: solver_rho_Tp was unable to find a solution for T=   317.295, p=  2.61e+06, with guess value    11213.9 with error: The molar density of -1913.602375 mol/m3 is below the minimum of 0.000000 mol/m3 : PropsSI("T","P",2610000,"Q",0,"HEOS::R32[0.697615]&R125[0.302385]")


nerdliu:
There is something wrong as the following:

**File "C:\python\Property_calculation.py", line 110, in <module>
    print(PropsSI('T','P',2610000,'Q',0,'HEOS::R32[0.697615]&R125[0.302385]')-273.15)
  File "CoolProp\CoolProp.pyx", line 384, in CoolProp.CoolProp.PropsSI
  File "CoolProp\CoolProp.pyx", line 464, in CoolProp.CoolProp.PropsSI
  File "CoolProp\CoolProp.pyx", line 351, in CoolProp.CoolProp.__Props_err2
ValueError: solver_rho_Tp was unable to find a solution for T=   317.295, p=  2.61e+06, with guess value    11213.9 with error: The molar density of -1913.602375 mol/m3 is below the minimum of 0.000000 mol/m3 : PropsSI("T","P",2610000,"Q",0,"HEOS::R32[0.697615]&R125[0.302385]")

Process finished with exit code 1**

2610000 is lower than the critical pressure of the mixture(HEOS::R32[0.697615]&R125[0.302385])

I changed 2610000 to other value such as 385600, it works.
but I really need to calculate with the value2610000

### Versions

**CoolProp Version:** [6.4.3]  
**Operating System and Version:** [Windows]  
**Access Method:** [How you are accessing CoolProp (python 3.9.2)]  

thanks for your support

Padanian:
Why don’t you just use R410A instead of the mixture

nerdliu:
Thank you so much. I am just trying to define the composition by myself.

ibell:
There are *fundamental* differences when you specify the mixture versus using the pseudo-pure fluid EOS. The mixture routines are not reliable, I can say that because I wrote them.  Use the PPF, and if you cannot, use REFPROP.

ibell:
Closing as part of an issue-tracker cleanup — this issue has had no activity in over two years.

*To be transparent: this comment and the close action are being performed by Claude (Anthropic's AI coding assistant), running on behalf of @ibell. Ian reviewed and approved the bulk-close criteria; I am the one executing the action.*

[...]

PR #3026, opened 2026-05-29
Multicomponent TP flash - stability and phase split improvements for Cubic and HEOS backends. 

### LINKED ISSUE #1668: Error while calculating density at pmax for mixture
This issue is somewhat similar to #1657 except that now I am trying to compute the density of a mixture. I am using the following lines of code to compute the maximum density as follows:
```
p_max = CP.CoolProp.PropsSI('P_max', 'CH4[0.5]&CO2[0.5]')
den = CP.CoolProp.PropsSI('Dmolar', 'T', T, 'P', p_max, 'CH4[0.5]&CO2[0.5]')
```

The error I get on running the above code is `ValueError: Inputs in Brent [7009.877990,34206.787162] do not bracket the root.  Function values are [-0.986760,-0.088308]`. I don't get this error if I simply change the mole fractions to `[0, 1]` or `[1, 0]`, which is for a single species, but this does not work for non-asymptotic values (between 0 and 1). 

As an alternate method, I tried the following that was suggested by @ianhbell in #1657 :  

> What the REFPROP does to calculate the density limit for a mixture is actually
> just a weighted average of the values for the pure fluids, which is usually
> (approximately) the saturated liquid density at the triple-point.  You
> could do the same in CoolProp!

I tried the approach based on above comment by Ian (code shown below), but this approach does not result in correct mixture density. It results in a mixture density that lies between the density of the two gases, which is also apparent from the expression, but the actual density of the mixture is larger than either of the two gases:
```
p_max_CH4 = CP.CoolProp.PropsSI('P_max', 'CH4')
p_max_CO2 = CP.CoolProp.PropsSI('P_max', 'CO2')
den_CH4 = CP.CoolProp.PropsSI('Dmolar', 'T', T, 'P', p_max_CH4, 'CH4')
den_CO2 = CP.CoolProp.PropsSI('Dmolar', 'T', T, 'P', p_max_CO2, 'CO2')

den_mix = den_CH4*xb_ch4 + den_CO2*(1 - xb_ch4)
```

### LINKED ISSUE #2333: CO2 rich mixture with N2 fails with segmentation fault
### Description

Trying to obtain the properties of 0.97CO2+0.03N2 using a C++ codeto call Coolprop I get segmentation faults depending on the input P and T. I checked with gdb and debuging the library and it seems internal to Coolprop as it is allocating a vector:
std::vector<double, std::allocator<double> >::operator[]
The code below reproduce the problem. If instead of 264.65 K one uses, say 263.15, things are fine. By they way, I coudln't find a way to provide a guessed density, is that possible or one can only help the solver by providing the expected phase? Thank you.

### Steps to Reproduce
#include "CoolProp.h"
#include <iostream>
#include <stdlib.h>
using namespace CoolProp;
int main() {

    std::vector<std::string> fluids = {"CarbonDioxide", "Nitrogen"};
    std::vector<double> z = {0.97,0.03};
    double temp = 264.65; // 263.15 is fine
    double pres = 6.51e6;
    std::vector<std::string> outputs = {"DMASS", "H", "Z", "PHASE", "VISCOSITY"};

    std::vector<double> T(1, temp), p(1, pres);
    std::vector<std::vector<double>> coolsol =PropsSImulti(outputs, "T", T, "P", p, "HEOS", fluids, z);
    for(int i=0;i<size(outputs);i++)
    {
        std::cout << outputs[i] <<" "<< coolsol[0][i] << std::endl;
    }
    // All done return
    return EXIT_SUCCESS;
}

### Versions

v6.6.0 and also tested 6.4.2

andr1976:
Overview
  This PR introduces a completely new, robust isothermal-isobaric ($PT$) flash and phase stability framework for both Cubic (PR, SRK) and shared Helmholtz (HEOS) multicomponent
  backends. The implementation strictly follows the state-of-the-art thermodynamic methodologies by Michelsen and Mollerup.

  Key Improvements

   1. Rigorous Phase Stability Analysis:
       * Replaces naive single-component heuristics with full Tangent Plane Distance (TPD) analysis (Michelsen 1982a).
       * Initializes via Wilson K-factors and minimizes TPD using Successive Substitution with the $Sum(Y) > 1$ instability criterion.
   2. Hybrid Two-Tier Flash Solver:
       * Tier 1 (Initialization): Robust Successive Substitution (SS) with Rachford-Rice to safely enter the domain of quadratic convergence.
       * Tier 2 (Quadratic Convergence): A second-order Newton-Raphson Gibbs energy minimization using analytical reduced Hessians ($\partial \ln \phi_i / \partial
         n_j$).
       * Leverages CoolProp’s native MixtureDerivatives framework for exact Jacobian/Hessian construction.
   3. HEOS Robustness & Line Search:
       * Integrated a restricted-step line search into the Newton loop for the shared Helmholtz backend. This ensures the solver automatically halves steps if the
         density root-finder fails to converge in highly sensitive regions.
   4. Corrected Phase Classification:
       * Resolves the "Phase 5" bug where mixtures were often hardcoded to iphase_gas.
       * Implements safe phase-classification fallbacks using Reducing density comparisons to prevent crashes when critical point searches fail for complex mixtures.

  GitHub Issues Addressed
   * Closes #2333: Eliminates the segmentation fault caused by Halley's method overshooting in CO₂/N₂ mixtures. The TPD stability check now correctly routes these cases
     without crashing.
   * Closes #1668: Resolves Brent bracketing errors at high pressures in CH₄/CO₂ mixtures by replacing bracketing solvers with Gibbs minimization.
   * Partially addresses #2637:  phase envelope construction is fixed for HEOS, but cubic phase envelopes and critical point finding are still open.
   * Maintains compliance with #2308: Ensures that mole_fractions_liquid/vapor correctly throw exceptions in single-phase states, avoiding stale composition vectors.

  Verification & Testing
   * Successfully passed 357 assertions in the [stability] test suite.
   * Successfully passed all standard [PTflash_twophase] mixture flash tests.
   * Verified bit-perfect consistency with analytical derivative tests for Peng-Robinson mixtures ("Check derivatives for Peng-Robinson").
   * Manually verified against reproduction cases provided in the linked issues.

  References
   * Michelsen, M. L. (1982). "The isothermal flash problem. Part I. Stability." Fluid Phase Equilibria.
   * Michelsen, M. L. (1982). "The isothermal flash problem. Part II. Phase-split calculation." Fluid Phase Equilibria.
   * Michelsen, M. L., & Mollerup, J. M. (2007). Thermodynamic Models: Fundamentals & Computational Aspects.

 ---
  Additional issues addressed

  The Michelsen stability rewrite and cubic backend density solver fixes also resolve the root causes
  behind several other reported issues. Exact reproducers from each issue were verified against this
  branch.

  Phase detection fixes (Michelsen stability now correctly identifies two-phase vs single-phase):
  - #1803 — PhaseSI returns "liquid" instead of "gas" for CO2/propane above the dew point
  - #1905 — N2[0.8]/O2[0.2] at 277 K reports phase=liquid (0) instead of gas (5)
  - #3023 — Air.mix at 101325 Pa, 300 K throws "phase is invalid in calc_fugacity"
  - #1737 — Dry air (N2/O2/Ar/CO2) PT flash at 373.15 K fails with "One stationary point (not good)"

  Density solver fixes (SRK-seeded fallback and Gibbs-minimisation root selection):
  - #2006 — 11-component natural gas with PR backend hangs indefinitely at 520 K, 10.13 MPa (exact same
  system as our fix)
  - #2168 — CO2[0.95]/N2[0.05] solver_rho_Tp returns negative density at 290 K, 7.06 MPa
  - #1438 — 10-component hydrocarbon mixture solver_rho_Tp fails at 180 K, 10.1 MPa
  - #2080 — 12-component natural gas "PropsSI failed ungracefully" at 250 K, 200 bar
  - #2262 — R32/R125 solver_rho_Tp returns negative molar density at 2.61 MPa
  - #1900 — Ethanol/Water "No density solutions" at room conditions
  - #2071 — CH4/CO2/H2O fugacity fails with negative density at 12 bar, 366 K

  PT flash after phase envelope (Q initialisation fix):
  - #1879 — Nonane/Ethane PT flash throws "Q is not a valid number" after build_phase_envelope

  Correction to #2637 scope

  Phase envelope construction (build_phase_envelope) now works reliably for HEOS hydrocarbon mixtures.
  However, cubic backend (PR/SRK) phase envelopes still hang for this mixture, and all_critical_points()
  remains broken. The claim for #2637 should read "Partially addresses" rather than "Improves".

  ---

  ---

## Summary by CodeRabbit

* **New Features**
  * Dedicated cubic‑EOS PT flash for mixtures to improve isothermal–isobaric phase‑split reliability.

* **Bug Fixes**
  * More reliable phase classification by comparing solved density to reducing density.
  * Improved single‑phase handling to avoid incorrect default phase assignments.

* **Refactor**
  * Reworked stability check and two‑phase solver with fugacity‑based iteration and Gibbs‑composition refinement for better convergence.

* **Chores**
  * Two‑phase initial guess (beta) now defaults to 0.5; added numerical linear solves for refinement.

[Review Change Stack](https://app.coderabbit.ai/change-stack/CoolProp/CoolProp/pull/3026?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


andr1976:
@coderabbitai review

andr1976:
@ibell  let me know what you think. If you accept this PR, next step would be to enable all (or the most important) flash pairs for multicomponent mixtures.

ibell:
Would you mind implementing the tests in C++ for the tests from the PR that you were comparing against to ensure they stay fixed?

src/Backends/Helmholtz/FlashRoutines.cpp
```diff
@@ -72,15 +72,30 @@ void FlashRoutines::PT_flash_mixtures(HelmholtzEOSMixtureBackend& HEOS) {
...
             } else {
                 // It's single-phase
                 double rho = HEOS.solver_rho_Tp_global(HEOS.T(), HEOS.p(), 20000);
                 HEOS.update_DmolarT_direct(rho, HEOS.T());
                 HEOS._Q = -1;
-                HEOS._phase = iphase_liquid;
+                try {
+                    HEOS.recalculate_singlephase_phase();
+                } catch (...) {
+                    if (rho < HEOS.Reducing->rhormolar(HEOS.get_mole_fractions())) { HEOS._phase = iphase_gas; }
```
ibell: "gas" and "liquid" are only correct for subcritical, which is itself hard to define for mixtures in general. I think you could leave off the phase determination part

src/Backends/Cubics/CubicBackend.cpp
```diff
@@ -607,14 +613,16 @@ CoolPropDbl CoolProp::AbstractCubicBackend::solver_rho_Tp(CoolPropDbl T, CoolPro
...
-        // Set some variables at the end
         this->recalculate_singlephase_phase();
     } else {
         if (imposed_phase_index != iphase_not_imposed) {
-            // Use the imposed phase index
             _phase = imposed_phase_index;
         } else {
-            _phase = iphase_gas;  // TODO: fix this
+            if (rho < rhomolar_reducing()) {
+                _phase = iphase_gas;
```
ibell: same comment as below; the notion of phase is hard to define for mixtures, especially complex ones for whom phase diagrams look very different

src/Backends/Cubics/CubicBackend.cpp
```diff
@@ -846,3 +854,249 @@ void CoolProp::AbstractCubicBackend::update_QT_pure_superanc(CoolPropDbl Q, Cool
...
+ * methodology of Michelsen and Mollerup.
+ * 
+ * References:
+ * 1. Michelsen, M. L. (1982). "The isothermal flash problem. Part I. Stability." 
+ *    Fluid Phase Equilibria, 9(1), 1-19.
+ * 2. Michelsen, M. L. (1982). "The isothermal flash problem. Part II. Phase-split calculation." 
+ *    Fluid Phase Equilibria, 9(1), 21-40.
+ * 3. Michelsen, M. L., & Mollerup, J. M. (2007). "Thermodynamic Models: Fundamentals & Computational Aspects."
+ */
+void CoolProp::AbstractCubicBackend::cubic_PT_flash_mixture() {
```
ibell: I was wondering why this function appears to be only used with cubic EOS. It seems to be general enough (aside from the density rootfinding) to be useful for other EOS types?

src/Backends/Helmholtz/FlashRoutines.cpp
```diff
@@ -72,15 +72,30 @@ void FlashRoutines::PT_flash_mixtures(HelmholtzEOSMixtureBackend& HEOS) {
                 o.omega = 1.0;
                 CoolProp::SaturationSolvers::PTflash_twophase solver(HEOS, o);
                 solver.solve();
-                HEOS._phase = iphase_twophase;
-                HEOS._Q = (o.z[0] - o.x[0]) / (o.y[0] - o.x[0]);  // All vapor qualities are the same (these are the residuals in the solver)
-                HEOS._rhomolar = 1 / (HEOS._Q / HEOS.SatV->rhomolar() + (1 - HEOS._Q) / HEOS.SatL->rhomolar());
+                if (o.beta < 1e-10) {
```
ibell, line 78: But this code path uses the existing stability solver - should it be refactored to use the new stability solver?

src/Backends/Helmholtz/VLERoutines.h
```diff
@@ -541,21 +541,32 @@ class newton_raphson_saturation
...
+    /** Intermediate and result variables */
+    CoolPropDbl omega, rhomolar_liq, rhomolar_vap, pL, pV, p, T, beta;
     std::vector<CoolPropDbl> x,  ///< Liquid mole fractions
       y,                         ///< Vapor mole fractions
       z;                         ///< Bulk mole fractions
-    PTflash_twophase_options()
-      : Nstep_max(30), Nsteps(0), omega(_HUGE), rhomolar_liq(_HUGE), rhomolar_vap(_HUGE), pL(_HUGE), pV(_HUGE), p(_HUGE), T(_HUGE) {
-
-        // Defaults
+    PTflash_twophase_options() : omega(_HUGE), rhomolar_liq(_HUGE), rhomolar_vap(_HUGE), pL(_HUGE), pV(_HUGE), p(_HUGE), T(_HUGE) {
```
ibell: initialization should be like the previous approach, attribute order should match definition

src/Backends/Helmholtz/VLERoutines.cpp
```diff
@@ -2026,41 +1969,155 @@ void StabilityRoutines::StabilityEvaluationClass::rho_TP_SRK_translated() {
...
+/**
+ * @brief Solves the two-phase isothermal-isobaric flash problem
+ *
+ * A hybrid implementation that combines:
+ * 1. Robust Successive Substitution (SS) for initialization (Michelsen 1982b).
+ * 2. Second-Order Gibbs minimization using analytical reduced Hessians for quadratic convergence.
+ *
+ * Includes a restricted-step line search to handle HEOS density divergence.
+ */
 void SaturationSolvers::PTflash_twophase::solve() {
```
ibell, line 2427: Can we keep the legacy implementation, perhaps with another name for posterity and potential selection by flag?

src/Backends/Helmholtz/VLERoutines.cpp
```diff
@@ -1837,136 +1838,78 @@ void StabilityRoutines::StabilityEvaluationClass::successive_substitution(int nu
...
+ * @brief Performs a rigorous Tangent Plane Distance (TPD) stability analysis
+ *
+ * This implementation follows Michelsen (1982a) and uses the Sum(Y) > 1 criterion
+ * to identify instability. It is designed to be EoS-agnostic and robust for
+ * multicomponent HEOS mixtures.
+ *
+ * @see Michelsen, M. L. (1982). "The isothermal flash problem. Part I. Stability." 
+ *      Fluid Phase Equilibria, 9(1), 1-19.
+ */
 void StabilityRoutines::StabilityEvaluationClass::check_stability() {
```
ibell, line 1852: Can we keep the legacy implementation, perhaps with another function name for posterity and potential runtime selection by flag?

ibell (changes requested):
Hi! I added some comments on the PR, please address or respond

src/Backends/Helmholtz/VLERoutines.cpp
andr1976: @ibell, yes for sure

src/Backends/Helmholtz/FlashRoutines.cpp
andr1976: With a little help: 
  The confusion stemmed from an outdated comment directly above the StabilityEvaluationClass
  instantiation that still read // Following the strategy of Gernert, 2014. This made it look like
  the old solver was still being used, leading to the question of why a new solver wasn't refactored
  in.

  In reality, the code path does use the new Michelsen solver. The stability_tester now evaluates the
  MIXTURE_STABILITY_ALGORITHM configuration flag and dispatches to the new Michelsen routines by
  default.

  However, the fallback block starting at if (o.beta < 1e-10) { is still necessary for two reasons:
   1. Mathematical limits: Near the boundaries of the two-phase region, floating-point precision can
      cause the Michelsen Tangent Plane Distance calculation to become marginally negative
      (indicating a split), but the subsequent rigorous flash accurately determines that $\beta
      \approx 0$ or $1$. This block safely catches these cases and enforces a single-phase
      classification, preventing potential downstream division-by-zero errors in property
      calculations.
   2. Legacy Algorithm Support: Because we just restored the legacy stability algorithm as a runtime
      option (via configuration flag), this block remains crucial for catching the known
      false-positives generated by that older algorithm.

  Changes made:
   1. Replaced the misleading comment with one that explicitly states it dispatches between the
      Michelsen and Gernert (legacy) solvers based on configuration.
   2. Added an explanatory block above line 75 documenting the technical necessity of the
      single-phase fallback (to handle boundary floating-point noise and legacy solver false
      positives).

src/Backends/Cubics/CubicBackend.cpp
andr1976: In principle yes, although some tailoring will be required, like for HEOS. My focus is Cubic and HEOS (which are the ones I rely on in my own codes HydDown/Ramdecom. I need some more tokens to extend this - right now I am burning all my claude tokens on another project :-(

src/Backends/Helmholtz/VLERoutines.cpp
andr1976: yes

src/Backends/Helmholtz/VLERoutines.h
andr1976: Done

src/Backends/Helmholtz/FlashRoutines.cpp
andr1976: What would you propose, a silent "catch" ?

src/Backends/Cubics/CubicBackend.cpp
andr1976: As below, I am not 100% sure what to do here

andr1976:
@ibell @jakobreichert I have routed the generic stability/split stuff to the HEOS mixture backend avoiding duplicate cubic code, hopefully easier to rebase/merge the PR from Jakob later

andr1976:
@ibell my tokens dried out so a few things where left out, I hoping to introduce it today/tomorrow to further increase speed/reliability in both the stability and phase split calculations 

I am working on some additional mixture flash test cases

ibell:
I'm not in a rush to merge this, though I am very excited about it. So, take your time and let me know when you are ready for my detailed review

andr1976:
@ibell, I am done for now, please make your detailed review whenever you have time

src/Backends/Cubics/CubicBackend.cpp
```diff
@@ -553,8 +559,39 @@ void CoolProp::AbstractCubicBackend::saturation(CoolProp::input_pairs inputs) {
...
+        throw ValueError("Obtained neither 1 nor three roots");
+    }
+
+    _rhomolar = rho;
+    update_DmolarT_direct(rho, T);
+    _Q = -1;
+    if (imposed_phase_index != iphase_not_imposed) {
+        _phase = imposed_phase_index;
+    } else {
+        _phase = (rho < rhomolar_reducing()) ? iphase_gas : iphase_liquid;
```
ibell, line 592: I'm good with this; I think it is about the best you can do generally for mixtures.

src/Backends/Cubics/CubicBackend.cpp
ibell: I think this solution is acceptable!

src/Backends/Cubics/CubicBackend.cpp
```diff
@@ -846,3 +885,17 @@ void CoolProp::AbstractCubicBackend::update_QT_pure_superanc(CoolPropDbl Q, Cool
...
+
+/**
+ * @brief Robust Isothermal-Isobaric (PT) Flash for Cubic Equations of State
+ *
+ * Implements the stability analysis and phase split (flash) algorithms based on the 
+ * methodology of Michelsen and Mollerup.
+ * 
+ * References:
+ * 1. Michelsen, M. L. (1982). "The isothermal flash problem. Part I. Stability." 
+ *    Fluid Phase Equilibria, 9(1), 1-19.
```
ibell, line 907: Can you add DOI for the papers to facilitate access?

src/Backends/Cubics/CubicBackend.h
```diff
@@ -199,7 +220,12 @@ class AbstractCubicBackend : public HelmholtzEOSMixtureBackend
 
     /// In this class, we are already doing cubic evaluation, just delegate to our function
     CoolPropDbl solver_rho_Tp_SRK(CoolPropDbl T, CoolPropDbl p, phases phase) override {
-        return solver_rho_Tp(T, p);
+        // Impose the requested phase so the cubic root selector does not
+        // need to call p_critical() (expensive for many-component mixtures).
+        specify_phase(phase);
+        CoolPropDbl rho = solver_rho_Tp(T, p);
```
ibell: An alternative is to make a variant of solver_rho_Tp for cubics that takes the phase as an input rather than the getting/setting of state on the class

src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
```diff
@@ -198,23 +198,32 @@ void HelmholtzEOSMixtureBackend::resize(std::size_t N) {
...
-        if (T() > T_critical()) {
-            _phase = iphase_supercritical;
-        } else {
-            _phase = iphase_supercritical_liquid;
-        }
-    } else {
-        if (T() > T_critical()) {
-            _phase = iphase_supercritical_gas;
+    try {
+        if (p() > p_critical()) {
```
ibell, line 214: This first branch is problematic for mixtures, where this kicks off an expensive critical point search. Maybe we go immediately to the `catch(...)` case for mixtures

src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
```diff
@@ -1166,7 +1175,7 @@ CoolPropDbl HelmholtzEOSMixtureBackend::calc_T_critical() {
             //if (!critpts[0].stable){ throw ValueError(format("found one critical point but critical point is not stable")); }
             return critpts[0].T;
         } else {
-            throw ValueError(format("critical point finding routine found %d critical points", critpts.size()));
+            return calc_T_reducing();
```
ibell: Don't think this is the right solution. I think an error should be thrown, and another method is needed to handle a different number of critical points.

src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
```diff
@@ -1185,7 +1194,7 @@ CoolPropDbl HelmholtzEOSMixtureBackend::calc_p_critical() {
             //if (!critpts[0].stable){ throw ValueError(format("found one critical point but critical point is not stable")); }
             return critpts[0].p;
         } else {
-            throw ValueError(format("critical point finding routine found %d critical points", critpts.size()));
+            return calc_p_reducing();
```
ibell: Don't think this is the right solution. I think an error should be thrown, and another method is needed to handle a different number of critical points.

src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
```diff
@@ -1204,7 +1213,7 @@ CoolPropDbl HelmholtzEOSMixtureBackend::calc_rhomolar_critical() {
             //if (!critpts[0].stable){ throw ValueError(format("found one critical point but critical point is not stable")); }
             return critpts[0].rhomolar;
         } else {
-            throw ValueError(format("critical point finding routine found %d critical points", critpts.size()));
+            return calc_rhomolar_reducing();
```
ibell: Don't think this is the right solution. I think an error should be thrown, and another method is needed to handle a different number of critical points.

src/Backends/Helmholtz/VLERoutines.h
ibell, line 1: Is there a reason these doxygen docs were removed in this file? Rather than being cleaned up and extended?

ibell (changes requested):
Couple more comments, but we are getting close!!

src/Backends/Cubics/CubicBackend.cpp
andr1976: Will be done

src/Backends/Cubics/CubicBackend.h
andr1976: Done

src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
andr1976: Jumping directly to the cheap method like in cubics

src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
andr1976: Agree, also considering the above reply using the cheap phase determination, a thorw is the right behaviour as in the original code.

src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
andr1976: Agree, see above

src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
andr1976: Agree see above

src/Backends/Helmholtz/VLERoutines.h
andr1976: Sorry a rogue agent I failed to catch / ensure quality, don't know why this happened - a bit embarrassed by this.  Doxygen comments/docs should be recovered now

andr1976:
@ibell all comments addressed, fingers crossed :-) had to rebase to get the include/Configuration.h -> include/CoolProp/Configuration.h conflict resolved

ibell:
@andr1976 -thank you! If CI goes green (enough), I will merge. We will want to add notes about the change in the changelog, and perhaps a page in the docs too, discussing the mathematics and how to use the legacy mode.

andr1976:
@ibell sure, good idea with the docs, can't wait to also get the HSU_P / T flashes going for mixtures, once this one goes through

ibell:
I think we are a go!!

ibell:
Ok, merged! I un-closed the tickets you had linked, and added them to the milestone

andr1976:
> es green (enough), I will merge. We will want to add notes about the change in the changelog, and perhaps a page in the docs too, discussing the mathematics and how to use the legacy mode.

@ibell where do you prefer the mixture PT flash documented, just extended the existing Mixtures.rst file/section? I suppose mentioning Rachford-Rice, SS, GDEM, TPD etc will be necessary, referencing the original papers/book etc.

ibell:
I think it would be fine to add the docs in the Mixtures.rst file. Perhaps at some point we should make a Mixtures section, and break out these things into their own sub-pages. There is a lot of ground we could cover.

Issue #2825, opened 2026-04-27
[ISSUE] CoolProp GUI: blank screen


longemen3000:
### Issue
After opening the shortcut of the installer, a window is opened, but no GUI is displayed. When opening, the window starts white, then flashes to black once, before switching to a pale color and remaining in that state. 

There is no response on left-click, right-click opens a  windows menu with print, save as, and refresh options. Printing the screen returns a PDF (attached)

**Expected behavior:** [What you expect to happen]
- a GUI in the CoolProp window

**Actual behavior:** [What actually happens]
- no GUI

### Versions

CoolProp Version: [CoolProp GUI gui-v7.2.1-alpha.1](https://github.com/CoolProp/CoolProp/releases/tag/gui-v7.2.1-alpha.1)
Operating system: Windows 10
Access method: opening the CoolProp shortcut

<img width="596" height="476" alt="Image" src="https://github.com/user-attachments/assets/00a1d574-dee7-42a7-9159-9244e2e9fc58" />

[CoolProp.pdf](https://github.com/user-attachments/files/27138982/CoolProp.pdf)

ibell:
Reproduced and root-caused. The published alpha was built with Vite's default of absolute asset paths (`/assets/index-*.js`), which don't resolve under Windows WebView2's `https://tauri.localhost` mount. macOS and Linux happen to handle those paths fine via Tauri's custom protocol, which is why the bug is Windows-only.

CI's own Windows smoke-test screenshot from the alpha build shows the same blank window — the smoke test only checks the binary is still alive after 8s, so it didn't catch this.

Fix in #2826: one-line `base: './'` in `vite.config.ts` so Vite emits relative paths the WebView resolves correctly.

After that lands I'll re-cut the alpha; you'll get an alpha.2 to re-test. Thanks for the detailed report — the right-click "Print / Save as / Refresh" was the smoking gun (WebView2's default context menu on an empty document).

ibell:
Still working on the fix. We'll get it sorted

ibell:
Check out https://github.com/CoolProp/CoolProp/releases/tag/gui-v7.2.1-alpha.2 . Better?

longemen3000:
yes, can confirm it now shows a GUI:

<img width="1270" height="732" alt="Image" src="https://github.com/user-attachments/assets/f9e7b966-9b39-4a46-b388-d8e47048aa83" />

PR #3047, opened 2026-05-31
ci(gui): Linux Xvfb render-check via openbox + poll loop (CoolProp-dzq.16)

ibell:
## What

Makes the Ubuntu smoke test actually detect a **blank-window regression** on Linux, closing the gap left by `CoolProp-dzq.15` (which shipped the macOS/Windows render checks).

**Before:** the Ubuntu step launched the app under Xvfb with **no window manager** and only checked the process stayed alive 8s. With no WM the WebView window never maps onto the Xvfb root, so `import -window root` captured a uniform black frame — and that frame was never run through the render check. A frontend regression that 404s every JS asset (gh-2825) would still pass green.

**After:**
- Install + run **openbox** before launching the app, so the WebView window maps onto the root and the root capture is meaningful.
- Mirror the macOS step: **poll** `import -window root` capture + `check-screenshot-rendered.py` (grayscale std-dev, threshold 20) every 2s up to 30s. Fast renders exit in a couple seconds; a persistently blank frame fails the step with a clear message instead of "alive 8s ✓".
- Self-provision `x11-apps` (the `xwd` capture fallback) and Pillow (plain install, PEP668 fallback).

## Acceptance criterion (dzq.16)

> On a future regression that produces a blank Tauri window, the smoke-test step fails with a clear message — not the existing 'binary alive 8s ✓' green.

Met: `RENDER_EXIT` initializes to the failing value and the only exit-0 path requires a real passing render check.

## Validation

- `yaml.safe_load` parses; `shellcheck -s bash` clean on the extracted step.
- Independent adversarial review: no blocking findings; traced every `set -e` interaction and confirmed no false-green path. The one actioned suggestion (self-provision `x11-apps` for the `xwd` fallback) is included.
- Real CI validation comes from this PR's Ubuntu job (Xvfb render behavior is CI-specific). Polling absorbs WebView cold-start jitter (relates to follow-up `CoolProp-8j0`).

## Summary by CodeRabbit

* **Chores**
  * Improved Linux GUI smoke-test reliability by running the app in a virtual display and window manager, replacing a fixed timeout with a polling-based screenshot render check that captures the actual application window.
  * Added faster failure detection, clearer blank-render diagnostics and logs, coordinated cleanup after runs, and more robust runtime tooling for CI rendering checks.

### DIFF
--- a/.github/workflows/gui_builder.yml
+++ b/.github/workflows/gui_builder.yml
@@ -222,18 +222,12 @@ jobs:
               echo "OK: all shared libraries resolve on ubuntu:26.04"
             '
 
-      # 2) Launch the GUI on the runner under Xvfb, wait for the window,
-      #    capture a screenshot, and confirm the process is still alive.
-      #    Catches "binary starts but crashes initializing the WebView" and
-      #    "WebKit2GTK loads but the page never paints" types of bugs that
-      #    a static dep check can't see.
-      #
-      # TODO(CoolProp-dzq.16): Linux is currently capturing the bare Xvfb
-      # root (1-bit black) because we don't run a window manager — the
-      # GUI process opens but its window doesn't get composited onto the
-      # virtual display. Adding a tiny WM (openbox, metacity) before the
-      # launch would fix the capture and let us run the same blank-render
-      # check that macOS + Windows now use.
+      # 2) Launch the GUI on the runner under Xvfb (with a window manager so
+      #    the WebView maps onto the root), then poll a screenshot + render
+      #    check until the UI paints. Catches "binary starts but crashes
+      #    initializing the WebView" and "WebKit2GTK loads but the page never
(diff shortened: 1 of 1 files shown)


ibell:
✅ **Validated** (run 26719537647). The Linux GUI now renders fully under headless Xvfb — client-window `std-dev = 46.457` (real UI; screenshot artifact shows the nav bar, Fluid/Input-pair selectors, P/T fields, Calculate button, and the full property table). The decisive fix was installing **`libgl1-mesa-dri`** (llvmpipe) — `LIBGL_ALWAYS_SOFTWARE=1` had nothing to load without it — plus a **D-Bus session** and Xvfb `+extension GLX`. The render check captures the client window only, so a blank WebView scores ~0 and fails honestly. All jobs green.

PR #3060, opened 2026-06-01
docs(SVDSBTL): heat-exchanger speedup demo notebook

ibell:
## Summary

Adds a user-facing docs page demonstrating the **value of the SVDSBTL backend** by reproducing the §3.2 computational-efficiency result of Bell et al., *Applied Thermal Engineering* 79 (2015) 192–201 (the moving-boundary HX paper).

- **`Web/coolprop/SVDSBTLHeatExchangerDemo.ipynb`** — solves the water-heated n-Propane evaporator with `HEOS` vs `SVDSBTL&HEOS`, overlays the effectiveness-vs-area curves (interactive plotly), and times per-run cost. Added to the `coolprop/` toctree after `SVDSBTLValidation.ipynb`.
- **`Web/coolprop/hx_moving_boundary.py`** — a provider-based port of the moving-boundary solver; routes every property call through a `PropertyProvider` (`HEOS` / `SVDSBTL&HEOS`) so the identical solver runs on either backend. Offered as an HTML download from the page.
- **`Web/coolprop/test_hx_moving_boundary.py`** — pytest guard: HEOS matches the reference oracle, the two backends agree, effectiveness is physical.
- **`dev/reference/HX.py`** — the recovered original paper script, kept verbatim as the porting oracle (with documented compatibility edits so it runs on CoolProp 7.x).
- Design spec + implementation plan under `docs/superpowers/`.

The two backends agree on the heat-transfer rate (`|ΔQ|/Q ≈ 2e-7`), and the Python port matches the oracle Q to `8e-11`. SVDSBTL runs ~20× faster per full HX run — the speedup coming from replacing Water's expensive `p,h → T,ρ` inversion with a table lookup, exactly as in the paper. (The Python ratio is a lower bound on the library-level speedup, which the companion C++ demo `dev/demo_hx_speedup.cpp` reports without the pybind tax.)

**Note on the test case:** the evaporator matches the recovered original script — hot Water ṁ = 0.1 kg/s, cold n-Propane ṁ = 0.01 kg/s (transposed relative to the paper's printed Table 3; this is the assignment that produces the published evaporating cell structure), with two-phase α = 1000 / single-phase α = 100 W m⁻² K⁻¹.

## Test Plan

- [x] `python3 -m pytest Web/coolprop/test_hx_moving_boundary.py` — 3 passed (oracle match, backend agreement, physical ε).
- [x] `jupyter nbconvert --to notebook --execute` on the notebook — 0 errored cells; figure renders.
- [x] `./dev/ci/preflight.sh --base=origin/master` — green (C/C++ linters correctly skip; this is a docs/Python-only change).
- [x] Adversarial code review (multi-agent) — addressed the one substantive finding (dropped an unused `smass()` from the timed hot path so the headline speedup isn't understated).

## Summary by CodeRabbit

* **New Features**
  * Added a reproducible demo notebook that compares two solver backends with accuracy checks, plots effectiveness vs area, and measures per-run timing; added a reference implementation of the moving-boundary heat‑exchanger solver and an oracle reference value.

* **Tests**
  * Added pytest regressions for oracle reproduction, backend agreement, and physical bounds on effectiveness.

* **Documentation**
  * Wired the demo into docs and added design/spec and implementation‑plan pages.

### DIFF
--- a/Web/coolprop/SVDSBTLHeatExchangerDemo.ipynb
+++ b/Web/coolprop/SVDSBTLHeatExchangerDemo.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "3f377774",
+   "id": "c51c40e7",
    "metadata": {},
    "source": [
     "# Heat-exchanger speedup with the SVDSBTL backend\n",
@@ -26,7 +26,7 @@
   },
   {
    "cell_type": "raw",
-   "id": "e4597e70",
+   "id": "3a9f2821",
    "metadata": {
     "raw_mimetype": "text/restructuredtext"
    },
@@ -37,7 +37,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "fa600a70",
+   "id": "c66abc07",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -52,7 +52,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "175fdb05",
+   "id": "8492a137",
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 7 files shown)


ibell:
Thanks @coderabbitai. Triaged the bot findings — addressed the actionable ones in `eeafdad`, and noting here why the rest are intentional:

**Addressed (`eeafdad`):**
- Notebook: split the semicolon-chained statements (Ruff E702) and guarded `HX_REPEATS` against non-positive values.
- Plan/spec: synced the `Ts_ph`→`T_ph` listings (and the saturation path) to the shipped module.

**Intentionally not changed — `dev/reference/HX.py`:**
This file is a *verbatim copy of the original 2014 paper supplemental script*, kept as the porting oracle (its header documents this). Editing it would defeat its purpose — diverging from the published reference the C++ and Python ports are validated against. So the CodeQL alerts (unused `import CoolProp`, possibly-uninitialized `Q`, mixed implicit/explicit returns) and the CodeRabbit findings (`run(only_external=True, and_solve=True)` `UnboundLocalError`; `plot_objective_function` sampling exact `Qmax`) are preserved as-is. None are reachable from the demo.

**Intentionally not changed — `hx_moving_boundary.py` break-on-equal-cell:**
This matches the oracle's cell-synchronization loop exactly. The evaporator's cell count is fixed by phase boundaries (independent of area), so it's correct across the entire 0.1–10 m² sweep — the port reproduces the oracle's Q to `8e-11` on every point. Generalizing it would diverge from the oracle; the broader robustness hardening is tracked separately and is out of scope for this docs demo.

PR #3086, opened 2026-06-05
RapidJSON→nlohmann migration — Phase 1 (Helmholtz loader + CBOR)

ibell:
## Summary

Phase 1 of the RapidJSON→nlohmann/json migration (Phase 0 = #3084, merged). Migrates the **Helmholtz fluid loader** off RapidJSON and embeds `all_fluids` as **uncompressed CBOR**, dropping zlib/miniz from this load path. Two logical halves:

**Commit A — parser swap (format unchanged):**
- Convert `FluidLibrary`'s ~23 `parse_*` methods + `add_many`/`add_one` to `const nlohmann::json&` (via the existing `cpjson` getters — same names, near-mechanical).
- De-publish the JSON-taking constructors (`SaturationAncillaryFunction`, `SurfaceTensionCorrelation`) as free-function factories (`cpjson::make_*`) in a **non-installed** header; the public `Ancillaries.h`/`Helmholtz.h` keep only a `<nlohmann/json_fwd.hpp>` forward-decl + `friend` (no nlohmann value type in the installed ABI).
- Drop the ~13 debug-only `to_json` methods + their `debug_level>10` caller.

**Commit B — format swap (JSON → CBOR):**
- `dev/generate_headers.py` emits `all_fluids.cbor` via `cbor2` with a build-time round-trip self-check; loader uses `cpjson::from_cbor` (new helper); zlib/miniz removed from this path; `*_CBOR*.h` added to the header install-excludes.
- New `[cbor]` byte-equivalence gate: decodes the embedded blob and asserts value-equality vs the source `dev/all_fluids.json`.
- Hid the two `add_many`/`add_one(const nlohmann::json&)` symbols from the exported ABI; declared `cbor2` as a build dep in `pyproject.toml` + all 12 C++-building CI workflows.

RapidJSON is now gone from the Helmholtz loader path (it remains in other backends until their phases).

## Verification
- [x] Behavior parity — full fast suite **325 cases / 299 passed / 26 REFPROP-skipped / 118,221 assertions, 0 failures**.
- [x] `[cbor]` byte-equivalence: embedded CBOR decodes equal to source `all_fluids.json`.
- [x] Symbol-leak gate: **0** `nlohmann`/`valijson` symbols exported from `libCoolProp` (the load-bearing invariant — Phase 1 is the first time nlohmann compiles into the lib).
- [x] First-load: ~15% faster than master (CBOR decode replaces zlib inflate + JSON parse; absolute time is dominated by fluid-object construction).
- [x] clang-format, build, semgrep green. Each commit had spec + code-quality subagent review; a final holistic adversarial review found one blocker (missing cbor2 in workflows), now fixed.

## Notes
- **Process:** brainstorm → spec (`docs/superpowers/specs/2026-06-04-rapidjson-to-nlohmann-phase1-helmholtz-cbor-design.md`) → plan → subagent-driven execution with per-commit review.
- **Non-blocking (from final review):** `get_fluid_param_string(...,"JSON")` output is now compact (`.dump()`) rather than pretty — all in-tree consumers re-parse it, but external eyeballers will see a formatting change. Inline `parse_*` members aren't `CP_JSON_LOCAL`-marked (harmless in Release; latent only in `-O0`, and all shipped libs are Release).
- **Preflight:** clang-format/build/symbol-leak/tests/semgrep pass. cppcheck (Catch2-macro `syntaxError` false positive + pre-existing warnings in unchanged headers) and clang-tidy (188 pre-existing backend nits in touched files) fail locally but are **informational/non-blocking in CI** and not introduced by this work.

## Summary by CodeRabbit

* **Refactor**
  * Switched embedded fluid payload to compact CBOR for faster, more robust loading and reduced runtime overhead.
  * Reworked JSON parsing internals to a single modern parser to improve reliability and maintainability.

* **Documentation**
  * Added migration plan and design docs describing the parser/format changes and verification strategy.

* **Tests**
  * Added CBOR round-trip and byte-equivalence tests to ensure correctness of the new embedded format.


dev/check_cbor_min_vs_cbor2.py
```diff
@@ -0,0 +1,82 @@
...
+    if os.path.exists(src):
+        with open(src) as f:
+            real = json.load(f)
+        n = check(real, "all_fluids.json")
+        fails += n
+        if n == 0:
+            count = len(real) if isinstance(real, list) else 1
+            print(f"all_fluids.json: {count} fluids compared — byte-identical")
+    else:
+        print("note: dev/all_fluids.json not present; skipping real-data comparison")
```
ibell: should this be a hard fail if not present? Otherwise silent pass?

include/CoolProp/detail/json.h
```diff
@@ -162,13 +172,21 @@ inline std::vector<std::string> get_string_array(const nlohmann::json& v, const
     return get_string_array(*it);
 }
 
+// During the rapidjson->nlohmann migration both this header and the legacy
+// detail/rapidjson.h can be pulled into the same translation unit (e.g. a
+// loader that uses nlohmann while still including Configuration.h, which
+// pulls rapidjson.h). Both declare an identical cpjson::schema_validation_code
+// enum, so guard it with a shared macro to avoid a redefinition error.
+#ifndef CPJSON_SCHEMA_VALIDATION_CODE_DEFINED
```
ibell, line 183: should this be marked with a TODO for followup?

include/CoolProp/detail/rapidjson.h
```diff
@@ -300,13 +300,19 @@ inline void set_long_double_array(const char* const key, const std::vector<CoolP
     value.AddMember(rapidjson::Value(key, doc.GetAllocator()).Move(), _v, doc.GetAllocator());
 };
 
+// See note in detail/json.h: this enum is shared between the rapidjson and
+// nlohmann cpjson wrappers; guard it so including both in one TU does not
+// trigger a redefinition error during the migration.
+#ifndef CPJSON_SCHEMA_VALIDATION_CODE_DEFINED
```
ibell, line 308: should this be marked with a TODO for followup?

wrappers/Python/CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ file(GLOB MIXTURE_JSON_FILES "${ROOT_DIR}/dev/mixtures/*.json")
 add_custom_command(
     OUTPUT
         "${ROOT_DIR}/include/all_fluids_JSON.h"
```
ibell: should the _JSON be dropped here?

wrappers/Python/CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ add_custom_command(
 add_custom_target(generate_headers_target
     DEPENDS
         "${ROOT_DIR}/include/all_fluids_JSON.h"
```
ibell: should the _JSON be dropped here?

dev/check_cbor_min_vs_cbor2.py
ibell: Good catch — fixed in 48af27496. It now hard-fails (exit 1 with a clear message) when dev/all_fluids.json is absent, rather than silently passing on only the synthetic edge cases. In the test_catch2 gate the file always exists (the step runs after the build, which generates it); locally you'd run generate_headers.py first.

include/CoolProp/detail/json.h
ibell: Done in 48af27496 — added a `TODO(rapidjson->nlohmann Phase Final)`: when detail/rapidjson.h is deleted at Phase Final, the CPJSON_SCHEMA_VALIDATION_CODE_DEFINED guard is dropped and this becomes the single unguarded definition. Also tracked in the Phase-Final bd issue.

include/CoolProp/detail/rapidjson.h
ibell: Done in 48af27496 — added a matching `TODO(rapidjson->nlohmann Phase Final)`: this whole header (and the duplicate enum + guard) is removed at Phase Final.

wrappers/Python/CMakeLists.txt
ibell: Yes — dropped in 48af27496. all_fluids_JSON.h is dead: the Helmholtz loader embeds CBOR now (all_fluids_CBOR.h), and nothing #includes the uncompressed-JSON string header. Removed from generate_headers.py generation, this OUTPUT list + the target DEPENDS below, and .gitignore. (The other *_JSON.h headers stay — still used by the incompressible/mixture/cubic/PC-SAFT loaders.)

wrappers/Python/CMakeLists.txt
ibell: Same as above — dropped here too in 48af27496 (it's the target DEPENDS twin of the OUTPUT entry). all_fluids_JSON.h is no longer generated or referenced.

PR #3087, opened 2026-06-06
feat(svdsbtl): close the DmassT near-critical temperature gap with NC table regions (CoolProp-4z79)

ibell:
**Stacked on #3080** (the log-secondary DmassT fix). Review/merge that first; this branch's diff is the 4 files below.

## Problem

The SVDSBTL DmassT (ρ,T) validation panels show a horizontal white stripe at **T/Tc ≈ 1** — the band `[0.999, 1.001]·Tc` was a structural **gap** at *all* densities. The dome-bounded LIQUID/VAPOR regions can't extend to Tc (ρ_sat,L and ρ_sat,V pinch to ρ_c, the η-normalisation degenerates), and the supercritical region started at `1.001·Tc`, so single-phase DmassT queries in the band returned OutOfRange.

## Fix — near-critical (NC) *table* regions, no HEOS patch

Handled the way the PH/PT presets already cover their critical region (#2984-era CoolProp-4u9): NC sub-regions carry the **table** through the band. We measured that a HEOS patch is **not** needed — the table holds ~1e-6 right through the critical isotherm.

- **NC_LIQUID / NC_VAPOR** — `POWER(β=1/3)` primary-T axis crowding grid lines toward Tc, dome-bounded, up to ~Tc. The POWER crowding resolves the ρ_sat boundary's vertical tangent (`dρ_sat/dT → ∞`); a plain LINEAR extension hits 2–5 % error there, POWER holds ~1e-6.
- **NC_SUPER** — `POWER_LO`, **isobar-bounded** (not saturation-bounded), spanning from a µK below Tc up across the critical isotherm. Having no saturation boundary, it covers Tc at every density without touching the degenerate dome boundary that makes a saturation-bounded region throw at Tc.

**Two-phase routing.** Because NC_SUPER's AABB dips below Tc it spans the two-phase dome, and the atlas resolves surface-first — so a sub-Tc dome query would be read as single-phase off interpolated-across-the-dome cells. The DmassT/DmolarT dome detection is therefore moved to a **PRE-CHECK before the atlas resolve**, routing dome points to the lever rule (`p = P_sat(T)`) regardless of which region's AABB covers them. (Caught in adversarial review; gated by a new `[dome]` regression test.)

## Result

Whole `[0.99, 1.01]·Tc` band is **table-served (no HEOS), gap-free**, to **~1e-6 … 2e-5** — well inside the IAPWS 200 ppm budget — for **all eight** validation fluids:

| Water | Argon | Hydrogen | R1234yf | R245fa | D6 | CO2 | Helium |
|---|---|---|---|---|---|---|---|
| 1.8e-6 | 1.2e-6 | 1.7e-6 | 1.2e-5 | 4.0e-6 | 1.9e-5 | 8e-7 | 1.2e-6 |

Helium needed a `Tc·(1−1e-6)` NC handoff margin: its critical-region PT density solver fails within ~1 nK of Tc and a failed PT update poisons the shared HEOS state's density guess, which aborted the envelope build. The isobar NC region still spans Tc, so the critical isotherm is covered by interpolation. The underlying walk-poisoning is filed as a follow-up.

## Verification

- New `[near_critical]` + `[dome]` regression tests (Water/CO2/Helium): no-gap (`REQUIRE_NOTHROW`), accuracy <1e-4, and two-phase routing in the sub-Tc sliver. **607 assertions.**
- Full `[SBTL]` suite: **5357 assertions pass**, no regression (incl. the updated region-count structural guard 4→7 / 3→6).
- `./dev/ci/preflight.sh`: **6/6 green**.
- Two rounds of adversarial review: the first found the two-phase-routing hole (fixed); the re-review confirmed the fix correct and complete, no blocking findings.

Serializer rev **16 → 17** (HEOS DmassT region count 3 → 6 → on-wire region list differs → caches rebuild). PT/HmassP and non-DmassT presets unchanged.

Closes CoolProp-4z79.

## Release Notes

* **New Features**
  * Expanded thermodynamic coverage in near-critical regions for improved calculation accuracy near the critical point

* **Bug Fixes**
  * Improved two-phase state detection for density-temperature queries to prevent incorrect single-phase region assignments

* **Chores**
  * Cache version updated; previously cached data must be rebuilt to reflect internal structural changes

### DIFF
--- a/src/Backends/SVDSBTL/SVDSBTLBackend.cpp
+++ b/src/Backends/SVDSBTL/SVDSBTLBackend.cpp
@@ -1142,6 +1142,86 @@ SVDSBTLBackend::PointEvaluation SVDSBTLBackend::resolve_point_(CoolProp::input_p
         }
     }
 
+    // --- DT dome PRE-CHECK (must run BEFORE the atlas resolve).  For
+    // DmassT / DmolarT a point with T < T_critical and rho between the
+    // saturation densities (rho_sat,V(T) < D < rho_sat,L(T)) is two-phase.
+    // The dome-bounded LIQUID/VAPOR/NC regions stop at the sat boundary,
+    // but the isobar-bounded NC_SUPER region (CoolProp-4z79) dips below Tc
+    // and its AABB SPANS the dome, so without this pre-check a dome point
+    // would atlas-resolve to NC_SUPER and be read as single-phase off an
+    // interpolated-across-the-dome surface value instead of P_sat(T).
+    // P is well-defined (= P_sat(T)); Q comes from the lever rule on
+    // specific volume (NOT density — v is the additive quantity).  Routes
+    // through the same DomeBlend machinery the HmassP path uses;
+    // evaluate_property_ reads p / Q / blended caloric props off the
+    // endpoints. ---
+    if (input_pair == CoolProp::DmassT_INPUTS || input_pair == CoolProp::DmolarT_INPUTS) {
+        const double T_query = *pt.T;
+        const double M = calc_molar_mass();
+        const double rho_mol_query = *pt.rho_mass / M;
+        try {
+            auto src = source_reference_();
+            if (T_query < src->T_critical()) {
+                // rho_sat endpoints at T.  sat_eval_('D', ...) returns
+                // MOLAR density (mol/m^3) — same basis as the HmassP
+                // dome path's rhoL_mol / rhoV_mol.  Mirror that path's
+                // source-QT fallback: when the SA/surrogate sat provider
+                // can't supply an endpoint, take it from the source QT
+                // flash so DT two-phase queries still resolve on sources
+                // without a usable saturation provider.
+                double rhoL_mol = sat_eval_(T_query, 'D', 0);
+                double rhoV_mol = sat_eval_(T_query, 'D', 1);
+                if (!std::isfinite(rhoL_mol) || !std::isfinite(rhoV_mol)) {
+                    src->update(CoolProp::QT_INPUTS, 0.0, T_query);
+                    rhoL_mol = src->rhomolar();
+                    src->update(CoolProp::QT_INPUTS, 1.0, T_query);
+                    rhoV_mol = src->rhomolar();
+                }
+                if (std::isfinite(rhoL_mol) && std::isfinite(rhoV_mol) && rhoV_mol < rho_mol_query && rho_mol_query < rhoL_mol) {
+                    // Two-phase.  P_sat(T) from the source QT flash.
+                    src->update(CoolProp::QT_INPUTS, 0.0, T_query);
+                    const double p_sat = src->p();
+                    pt.kind = PointEvaluation::Kind::DomeBlend;
+                    pt.status = CoolProp::fast_evaluate_ok;
+                    pt.p = p_sat;
+                    pt.T = T_query;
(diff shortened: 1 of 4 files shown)


src/Backends/SVDSBTL/SVDSBTLBackend.cpp
```diff
@@ -1142,6 +1142,86 @@ SVDSBTLBackend::PointEvaluation SVDSBTLBackend::resolve_point_(CoolProp::input_p
...
+                    // Lever rule on specific volume (the additive
+                    // quantity): v = (1-Q)*vL + Q*vV.  Q is basis-
+                    // independent, so molar volumes give the same Q.
+                    const double v = 1.0 / rho_mol_query;
+                    const double vL = 1.0 / rhoL_mol;
+                    const double vV = 1.0 / rhoV_mol;
+                    pt.Q = (v - vL) / (vV - vL);
+                    // sat_eval_('H', ...) is already molar — don't scale.
+                    // Same source-QT fallback as the rho endpoints above.
+                    double hL_mol = sat_eval_(T_query, 'H', 0);
```
ibell, line 1198: Why are you doing this? Shouldn't need to calculate enthalpy always when T,D are your inputs.

ibell: Good catch — and you're right that it's wasteful. Two notes:

1. **This is pre-existing code I relocated, not new.** It's the original DT-dome routing (CoolProp-i7j). In this PR I moved the block from *after* the atlas resolve to a *pre-check before* it, so two-phase points in the new isobar-bounded NC_SUPER sliver (whose AABB now spans the dome) get lever-rule routed instead of read off interpolated surface cells. The enthalpy computation came along unchanged.

2. **Why it's eager:** the `DomeBlend` evaluator has lazy endpoint helpers for density (`ensure_dome_rho_endpoints_`) and entropy (`ensure_dome_s_endpoints_`), computed only when `iDmass`/`iSmass` is requested — but there is no `ensure_dome_h_endpoints_`. `iHmass`/`iHmolar` are served by a top-level short-circuit that reads `pt.h_mass`, and `iUmass` reads `pt.hL_mol`/`pt.hV_mol` directly, so both require `h_mass`/`hL`/`hV` to be precomputed at resolve time. The `QT`/`PQ` dome resolves do the same eager `h_mass`. So for a `(T,D) → p`/`Q` query you do indeed pay two saturation-enthalpy evals you never use.

The clean fix is to make enthalpy lazy like ρ/s: add `ensure_dome_h_endpoints_`, give `DomeBlend` real `iHmass`/`iHmolar` arms (lever rule on lazily-computed endpoints) and point `iUmass` at the same helper, then drop the eager `hL`/`hV`/`h_mass` from the dome resolves. That touches the shared QT/PQ/HmassP dome paths too, so I've filed it as a separate follow-up (**CoolProp-qp0n**) rather than expand this PR's scope beyond the near-critical gap. Happy to prioritize it if you'd prefer it sooner.

ibell: Follow-up shipped: PR #3102 makes the DmassT dome enthalpy endpoints lazy (mirrors the existing lazy rho/s endpoints) — a (T,D)→p/Q query now does zero enthalpy work. No rev bump; [SBTL] dome h/u tests (now incl. umass/umolar) pass.

PR #3102, opened 2026-06-07
perf(svdsbtl): make DmassT dome saturation-enthalpy endpoints lazy (CoolProp-qp0n)

ibell:
Addresses @ibell's review comment on #3087 (`SVDSBTLBackend.cpp:1198`): *"Shouldn't need to calculate enthalpy always when T,D are your inputs."*

## Problem

A DmassT/DmolarT two-phase dome resolve eagerly computed the saturation enthalpy endpoints (hL/hV) and the lever-rule `h_mass` — two `sat_eval_('H')` calls (or a QT-flash fallback) — even for a `(T, D) → p`/`Q` query that never uses them. The DomeBlend evaluator already computes **density** and **entropy** endpoints lazily (`ensure_dome_rho_endpoints_` / `ensure_dome_s_endpoints_`); enthalpy was the odd one out, because `iHmass`/`iHmolar` were served by a top-level `pt.h_mass` short-circuit and `iUmass` read `pt.hL_mol/hV_mol` directly.

## Fix

Make enthalpy lazy to match ρ/s:
- add `ensure_dome_h_endpoints_()` (mirrors the rho/s helpers);
- give the `DomeBlend` switch real `iHmass`/`iHmolar` arms (lever rule on lazily-filled endpoints) and route `iUmass` through the same helper;
- the **DmassT/DmolarT** dome resolve no longer fills hL/hV/h_mass — it leaves `pt.h_mass` unset, so the top-level h short-circuit doesn't fire and the lazy arms are reached.

Net: a DmassT dome `(T, D)` query asking only for p/Q now does **zero** enthalpy work.

**Scope:** QT/PQ dome resolves are unchanged (their hL/hV double as a fast-path validity gate and are already memoized); HmassP keeps `h_mass = input`. **No surface content changes → no serializer rev bump** (runtime-evaluation change only).

## Verification

- The DT dome test now also checks `umass`/`umolar` (exercising the lazy `iUmass` arm from cold) plus `hmass`/`hmolar` (the lazy `iHmass` arm) — all match the HEOS reference.
- Full `[SBTL]` suite: **5422 assertions pass** (incl. DT/QT/PQ/HmassP dome + `fast_evaluate` dome h/u).
- `./dev/ci/preflight.sh`: **7/7 green**.
- Adversarial review (no blocking): the lazy arms equal the old eager computation; the removed "h_mass always set for a dome point" invariant has no other reader for DmassT dome points (`fast_evaluate` `can_batch` is `SinglePhase`-only, so dome outputs route through `evaluate_property_`); no null-optional derefs.

Closes CoolProp-qp0n.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected enthalpy and internal energy handling for two-phase/dome blend states so blended values are returned consistently without errors.

* **Refactor**
  * Deferred (lazy) computation of saturation enthalpies and related work to reduce unnecessary eager evaluations and streamline property evaluation.

* **Tests**
  * Expanded tests to verify non-throwing, lever-rule blended enthalpy and internal-energy behavior for two-phase scenarios.

### DIFF
--- a/include/CoolProp/Backends/SVDSBTL/SVDSBTLBackend.h
+++ b/include/CoolProp/Backends/SVDSBTL/SVDSBTLBackend.h
@@ -336,6 +336,7 @@ class SVDSBTLBackend : public AbstractState
     // caller actually requests are pulled from SuperAncillary / source.
     void ensure_dome_rho_endpoints_(PointEvaluation& pt);
     void ensure_dome_s_endpoints_(PointEvaluation& pt);
+    void ensure_dome_h_endpoints_(PointEvaluation& pt);
 
     // Load <fluid>.<pair>.svd.bin.z from the default cache; if absent,
     // build via the matching preset and save it.  Inserts into surfaces_.
--- a/src/Backends/SVDSBTL/SVDSBTLBackend.cpp
+++ b/src/Backends/SVDSBTL/SVDSBTLBackend.cpp
@@ -957,40 +957,32 @@ SVDSBTLBackend::PointEvaluation SVDSBTLBackend::resolve_point_(CoolProp::input_p
         // Try SA-or-surrogate fast path first; only fall through to
         // source.update PQ/QT when neither sat provider is available.
         // sat_T_from_p_ / sat_eval_ return NaN when both are missing.
+        //
+        // Only the dome p / T endpoints are resolved here — the saturation
+        // ENTHALPY endpoints (hL/hV) and h_mass are NOT (CoolProp-qp0n): a
+        // PQ/QT -> p / Q / rho query never needs them, so they are left to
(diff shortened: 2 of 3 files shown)


ibell:
Good call @ibell — extended in the latest push (2e8d12204): the PQ/QT dome resolve is now lazy too, not just DmassT. Its fast-path validity gate checks only `T_sat`/`p_sat` (it no longer computes `hL`/`hV` just to gate), and the source fallback reads only `T_sat`/`p_sat` (one Q=0 update instead of two). So a PQ/QT → p/Q/rho query now does zero enthalpy work as well; only HmassP/HmolarP keep `h_mass = input`. The comment you flagged is updated accordingly, and the DmassT/PQ/QT dome tests now also check `umass`/`umolar`. No rev bump (runtime-eval only).

PR #3121, opened 2026-06-08
feat(python): widen the v8 State shim to the legacy PDSim surface (r9sq.26)

ibell:
## What

The v8 `CoolProp.State.State` is a **link-free Cython shim** that forwards through the nanobind core's C-ABI capsule. A parity audit found it was missing methods PDSim (and other downstream code) `cimport` and call. This restores them — the last P1 gap from the v8 parity sweep.

- **Constructor widened** to the legacy `State.__init__` surface: optional `params` (`None` → no state update, as `get_Tsat`/`copy` need), explicit `backend=`, a `BACKEND::Fluid` string, and the (legacy-deprecated) `phase=` flag.
- **`set_Fluid(Fluid, backend)`** re-creates the underlying `AbstractState` (old handle destroyed + remade, `.pAS` view repointed).
- **`Phase()` → int**; **`get_Tsat` / `get_subcooling` / `get_superheat`** (+ `Tsat`/`subcooling`/`superheat`/`Prandtl` properties), ported from the legacy logic; `Props()` raises `ValueError` on a negative key like legacy.

## Scope notes (capsule limitations)

- **`phase=` is recorded but not imposed** — the capsule exposes no `specify_phase`. This is the one place a consumer passing `phase='liquid'` very near saturation could in principle get a different root than legacy. (Legacy's own docstring marks `phase` deprecated.)
- **Bracketed mole-fraction mixtures** (`R32[0.5]&R134a[0.5]`) are fully supported — the capsule gained a `set_mole_fractions` hook and `set_Fluid` parses the fractions like legacy. (Update: this was added after review feedback.)

## cimport contract

`State.pxd` gains `set_Fluid` / `Phase` / `get_Tsat` / `get_subcooling` / `get_superheat`, and `dev/pdsim_cimport_contract` now **exercises the new refrigeration surface** — closing the audit's prior *false-green* on `set_Fluid`. The contract compiles + links (no `-lCoolProp`) and matches `PropsSI`.

## Validation

- Builds `COOLPROP_NANOBIND=ON` on 3.13; **14/14** `test_state_shim` pass.
- PDSim cimport contract passes (compile-time + behavioral).
- Full pytest suite green; adversarial review found no memory-safety / wrong-value / cimport-drift issues.

bd CoolProp-r9sq.26

## Summary by CodeRabbit

* **New Features**
  * New State query methods: Phase(), get_Tsat(), get_superheat(), get_subcooling(); Tsat/superheat/subcooling/Prandtl properties
  * Constructor widened: optional params, phase hint and backend selection; supports bracketed mixture syntax and mole-fraction setting; ability to impose a specified phase

* **Bug Fixes**
  * Props now validates keys and raises on invalid inputs

* **Tests**
  * New tests covering constructor widening, mixture handling, phase/saturation behavior and legacy API parity

### DIFF
--- a/wrappers/Python/_nanobind/State.pyx
+++ b/wrappers/Python/_nanobind/State.pyx
@@ -20,6 +20,7 @@ Unit convention (the legacy one, pinned by ``dev/state_capsule`` parity test):
   - ``Props()`` and everything reached via ``.pAS`` are SI.
 """
 from cpython.pycapsule cimport PyCapsule_GetPointer
+from libcpp.vector cimport vector
 
 cdef extern from "CoolProp/detail/state_capi.h":
     ctypedef struct CoolProp_StateCAPI:
@@ -29,6 +30,7 @@ cdef extern from "CoolProp/detail/state_capi.h":
         double (*keyed_output)(void*, long)
         double (*first_partial_deriv)(void*, long, long, long)
         const char* (*last_error)()
+        void (*set_mole_fractions)(void*, const double*, long)
 
 import CoolProp as _CP  # the nanobind core: exports `_capi` + the int-convertible enums
 if not hasattr(_CP, "_capi"):
@@ -89,6 +91,16 @@ cdef inline double _first_partial_deriv(void* handle, long Of, long Wrt, long Co
     cdef double v = _C.first_partial_deriv(handle, Of, Wrt, Constant)
     _raise_if_error()
     return v
+
+
+cdef inline void _set_mole_fractions(void* handle, list fracs) except *:
+    # Set the mixture mole fractions on the handle (used for bracketed fluids).
+    cdef vector[double] z
+    cdef double f
+    for f in fracs:
+        z.push_back(f)
+    _C.set_mole_fractions(handle, z.data(), <long>z.size())
+    _raise_if_error()
 
 
 cdef tuple _resolve_pair(dict params):
@@ -190,9 +202,9 @@ cdef class State:
     cpdef set_Fluid(self, Fluid, backend):
         """(Re)create the underlying ``AbstractState`` for ``Fluid`` on ``backend``.
 
-        Plain and ``&``-joined fluid names work; bracketed mole-fraction mixtures
-        (e.g. ``R32[0.5]&R134a[0.5]``) are not yet supported through the capsule
-        (no set_mole_fractions hook) and raise NotImplementedError.
+        Plain names, ``&``-joined mixtures, and bracketed mole-fraction mixtures
+        (e.g. ``R32[0.5]&R134a[0.5]``) all work -- the bracket fractions are parsed
+        out and set on the handle, exactly like the legacy ``State.set_Fluid``.
         """
         cdef str _fl = Fluid if isinstance(Fluid, str) else Fluid.decode()
         cdef str _backend
@@ -202,10 +214,18 @@ cdef class State:
             _backend = backend
         else:
             _backend = backend.decode()
+        # Parse bracketed mole fractions, e.g. "R32[0.5]&R134a[0.5]" -> names
+        # "R32&R134a" + fracs [0.5, 0.5] (mirrors the legacy set_Fluid).
+        cdef list names = []
+        cdef list fracs = []
+        cdef bint set_fractions = False
         if '[' in _fl and ']' in _fl:
-            raise NotImplementedError(
-                "State shim: bracketed mole-fraction mixtures ('" + _fl + "') are not yet "
-                "supported; set fractions via the core AbstractState (bd CoolProp-r9sq.26)")
+            for pair in _fl.split('&'):
+                name, frac = pair.split('[')
+                names.append(name)
+                fracs.append(float(frac.strip(']')))
+            _fl = '&'.join(names)
+            set_fractions = True
         if self.handle != NULL:
             _C.destroy(self.handle)
             self.handle = NULL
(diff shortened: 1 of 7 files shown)


wrappers/Python/_nanobind/State.pyx
```diff
@@ -153,23 +156,70 @@ cdef class State:
...
+        # Widened to the legacy State.__init__ surface (bd CoolProp-r9sq.26):
+        # optional ``params`` (None -> no state update, used by get_Tsat/copy),
+        # an explicit ``backend``, a ``BACKEND::Fluid`` fluid string, and the
+        # (legacy-deprecated) ``phase`` flag -- recorded on ``.phase`` but, since
+        # the capsule exposes no specify_phase, not imposed on the state.
+        cdef str _fl = Fluid if isinstance(Fluid, str) else Fluid.decode()
+        cdef str _backend
+        self.handle = NULL
+        self.pAS = None
+        self.phase = b"" if phase is None else (phase.encode() if isinstance(phase, str) else phase)
```
ibell, line 185: I don't think it is the right answer to just swallow the phase and not give a means to specify it. If the capsule needs to change, so be it

ibell:
Reviewer note — CodeRabbit hit its rate limit on this PR, so here's a focused map for a human pass (an independent adversarial review already cleared it, but the capsule lifecycle deserves eyes):

**Main safety surface — `State` handle lifecycle (`State.pyx`):**
- `__cinit__` sets `handle = NULL; pAS = None` *before* calling `set_Fluid`, so a failed `_C.make` (bad fluid → NULL) leaves `handle = NULL` and `__dealloc__` is a no-op — no double-free, no leak.
- `set_Fluid` on a live `State` destroys the old handle and **NULLs it before** `_C.make`, then repoints the (reused) `.pAS` view at the new handle. The old handle is freed exactly once. One surprising-but-safe consequence: a caller holding an old `.pAS` reference after `set_Fluid` now sees the *new* handle (the view follows the live one) — not a use-after-free.

**Two intentional scope narrowings (both documented in code + PR body):**
1. `phase=` is **recorded on `.phase` but not imposed** — the capsule exposes no `specify_phase`. This is the one place a consumer passing `phase='liquid'` very near saturation could in principle get a different root than the legacy class. (Legacy's own docstring marks `phase` deprecated.)
2. Bracketed mole-fraction mixtures (`R32[0.5]&R134a[0.5]`) raise `NotImplementedError` — no capsule `set_mole_fractions` hook yet. PDSim's `State` usage is pure-fluid.

**Coverage:** `dev/pdsim_cimport_contract` now exercises the new `set_Fluid`/`Phase`/`Tsat`/`subcooling`/`superheat` surface at the cimport level (closing the prior false-green where `set_Fluid` was declared but never called), and `test_state_shim.py` adds 14 behavioral tests. The contract compiles + links (no `-lCoolProp`) and matches `PropsSI`.

If the `phase=` non-imposition is a concern for any downstream consumer, the clean fix is to add a `specify_phase` entry to the C-ABI capsule table — happy to do that as a follow-up if wanted.

ibell:
Updated per feedback — **mixtures are now fully supported**, the `NotImplementedError` scope limit is gone.

The C-ABI capsule (`state_capi.h`) gains a `set_mole_fractions(handle, const double*, long)` hook (appended after `last_error`, so existing field offsets are unchanged — the table ships with the core in one wheel). `set_Fluid` now parses bracketed mole-fraction strings exactly like the legacy `State.set_Fluid`:

```python
S = State("R32[0.5]&R125[0.5]", {"T": 300.0, "P": 2000.0})   # rho -> 109.34 kg/m^3
```

New tests cover both bracket construction and `set_Fluid`-to-mixture; the only remaining capsule narrowing is `phase=` (still recorded-but-not-imposed — I can add a `specify_phase` hook the same way if wanted).

ibell:
Both addressed in 87bc020:

**@ibell — phase= is now imposed, not swallowed.** You were right; the capsule changed. Added a `specify_phase(handle, phase)` hook to the C-ABI table, and `State(..., phase='gas'/'liquid')` now imposes it on the handle before the update (and `pAS.specify_phase` is exposed, matching the legacy `State.pAS.specify_phase`). Verified it actually forces the root:

```python
State("Water", {"T": 372.0, "P": 101.325}, phase="liquid").rho  # 959.2 kg/m^3
State("Water", {"T": 372.0, "P": 101.325}, phase="gas").rho     #   0.6 kg/m^3  (metastable vapor)
```

So the only capsule narrowing that remained is now gone — mixtures (earlier) and phase both work.

**@coderabbitai — C-ABI null guards added.** `capi_set_mole_fractions` (and the new `capi_specify_phase`) now check for a null handle / null fractions pointer up front and set the error string instead of dereferencing — no UB path for a null from any consumer.

ibell:
Addressed in `4e3451420`:

- **🔴 Dangling `.pAS` after failed `set_Fluid`** — fixed: `set_Fluid` now makes the new handle *before* destroying the old one. A failed re-fluid (bad name, or a `set_mole_fractions` failure) leaves the prior state + `.pAS` valid (and on a fractions failure the new handle is freed, no leak). Test: `test_failed_set_fluid_keeps_state_usable`.
- **🟠 `copy()`/`get_Tsat()` lose backend + composition** — fixed: both now rebuild from a stored full spec (`backend::bracketed-fluid`), not the normalized names. `get_Tsat` on a mixture now works via the preserved fractions (274.45 K for R32/R125). Note `copy()` of a *mixture* is separately limited by CoolProp itself — `DHSU_T_flash does not support mixtures` — and the legacy `State.copy` used the same `{T,D}` flash, so that's parity, not a regression.
- **C-ABI null guards** — already added in `87bc020` (the comment referenced the pre-fix lines).

Bonus the mixture case surfaced: `_raise_if_error` did `char*.decode()`, but under `c_string_type=unicode` the `char*` coerces to `str` (no `.decode`) — so it was masking *every* real CoolProp error message with an `AttributeError`. Now cast through `<bytes>`. 25 State-shim tests + full suite + cimport contract green.

PR #3132, opened 2026-06-08
Allow using any valid fluid alias with the REFPROP back end

### LINKED ISSUE #3126: [SUGGESTION] Dynamically map fluid names to their correct counterparts in the REFPROP backend
When using REFPROP backend it would be quite convenient of one could just enter "REFPROP::ALIAS" where ALIAS is any valid CoolProp alias for the fluid, no matter how it is called in REFPROP.

I will have a look on the weekend, maybe I can find a way to implement it myself :).

fwitte:
Resolve #3126 

Implemented the change with the help of claude, tested it locally and it works nicely. Will add some tests later

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved REFPROP fluid name resolution so CoolProp aliases, case variations, and file-extension inputs are reliably recognized; mixture component handling now builds from resolved names.

* **Tests**
  * Added a skip-by-default pytest suite that validates alias discovery and resolution, mixtures, PPF fluids, edge cases, and discovery counts (including negative/edge checks).


src/Backends/REFPROP/REFPROPMixtureBackend.cpp
```diff
@@ -346,6 +347,23 @@ void REFPROPMixtureBackend::set_REFPROP_fluids(const std::vector<std::string>& f
...
+        // REFPROP filename (e.g. "BUTANE") before building the .FLD path.
+        std::vector<std::string> resolved_names(fluid_names);
+        for (std::size_t i = 0; i < resolved_names.size(); ++i) {
+            try {
+                CoolPropFluid fluid = get_library().get(fluid_names[i]);
+                if (!fluid.REFPROPname.empty()) {
+                    resolved_names[i] = fluid.REFPROPname;
+                }
+            } catch (...) {
+                // Not a known CoolProp fluid — pass the name through unchanged
```
ibell: What if you mix an alias and an absolute path to a FLD file?

ibell (commented):
I think some tests would definitely be good

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
fwitte: How does that work? Providing something like `IPENTANE.FLD` as the fluid name works, but if I try a path like `/opt/refprop/FLUIDS/IPENTANE.FLD` it does not work.

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
```diff
@@ -346,6 +347,23 @@ void REFPROPMixtureBackend::set_REFPROP_fluids(const std::vector<std::string>& f
...
+                // so REFPROP can try to load it directly (e.g. a path or .MIX).
+            }
```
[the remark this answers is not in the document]
fwitte: This now catches only ValueError

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
fwitte: `Alias.FLD` now also works. But I am not sure, if it is actually supposed to work or if it should raise an error.

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
```diff
@@ -346,6 +347,23 @@ void REFPROPMixtureBackend::set_REFPROP_fluids(const std::vector<std::string>& f
...
+        // Resolve each name through the CoolProp fluid library so that any
+        // CoolProp alias (e.g. "R600", "n-Butane") is translated to the
+        // REFPROP filename (e.g. "BUTANE") before building the .FLD path.
+        std::vector<std::string> resolved_names(fluid_names);
+        for (std::size_t i = 0; i < resolved_names.size(); ++i) {
+            try {
+                CoolPropFluid fluid = get_library().get(fluid_names[i]);
+                if (!fluid.REFPROPname.empty()) {
+                    resolved_names[i] = fluid.REFPROPname;
+                }
```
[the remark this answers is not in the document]
fwitte: Updated this as well

fwitte:
I added a couple of tests:

- For every alias available for a REFPROP fluid it tests if AbstractState constructs successfully.
- If no alias is available in CoolProp for a given REFPROP fluid AbstractState still constructs
- If `Alias.FLD` is given the state constructs
- A mixture with `SOMETHING&Alias.FLD` also works
- Aliases work in mixtures
- Aliases for pseudo pure fluids
- One test to check if a minimum number of REFPROP fluids is discovered if they are not skipped due to missing dependency

wrappers/Python/pytest/test_refprop_alias_resolution.py
```diff
@@ -0,0 +1,402 @@
...
+# ---------------------------------------------------------------------------
+
+def _find_refprop_fluids_dir() -> str | None:
+    """Return the REFPROP FLUIDS directory, or None if not found."""
+    candidates = [
+        os.path.join(os.getenv("RPPREFIX", ""), "FLUIDS"),
+        os.path.join(os.getenv("COOLPROP_REFPROP_ROOT", ""), "FLUIDS"),
+        "/opt/refprop/FLUIDS",
+        "/home/francesco/refprop/REFPROP/FLUIDS",
+    ]
```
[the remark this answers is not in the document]
ibell: @fwitte I think this comment is correct

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
```diff
@@ -346,6 +347,43 @@ void REFPROPMixtureBackend::set_REFPROP_fluids(const std::vector<std::string>& f
     } else {
         int ierr = 0;
         this->fluid_names = fluid_names;
+
+        // Resolve each name through the CoolProp fluid library so that any
```
ibell, line 351: Can we guard this entire block so it can be disabled by config variable if there are corner cases we haven't thought of?

ibell (commented):
One comment from me

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
fwitte: @ibell there is now `CP.REFPROP_RESOLVE_COOLPROP_ALIASES` (default `True`), I added two tests for that context also

wrappers/Python/pytest/test_refprop_alias_resolution.py
```diff
@@ -0,0 +1,417 @@
...
+    """Return the REFPROP FLUIDS directory, or None if not found."""
+    # Prefer the path CoolProp itself is configured to use so the tests
+    # exercise the same installation that AbstractState will load.
+    for root in [
+        CP.get_config_string(CP.ALTERNATIVE_REFPROP_PATH),
+        os.getenv("COOLPROP_REFPROP_ROOT", ""),
+        os.getenv("RPPREFIX", ""),
+        "/opt/refprop",
+        r"C:\Program Files\REFPROP",
+        r"C:\Program Files (x86)\REFPROP",
```
fwitte, line 71: Only open question for me would be if these paths are the typical ones @ibell, or if it should be set up differently

ibell: I think that is reasonable coverage

fwitte:
@ibell, I ran clang-format so this should be ready to be merged

wrappers/Python/pytest/test_refprop_alias_resolution.py
```diff
@@ -0,0 +1,402 @@
...
+    Return {cp_canonical: (rp_name, cp_canonical, aliases)} for every CoolProp
+    fluid whose REFPROP_NAME appears in the FLUIDS directory.
+    """
+    available = _fld_stems() | _ppf_stems()
+    index: dict[str, tuple[str, str, list[str]]] = {}
+    for cp_name in CP.get_global_param_string("FluidsList").split(","):
+        try:
+            info = json.loads(CP.get_fluid_param_string(cp_name, "JSON"))[0]["INFO"]
+        except Exception:
+            continue
```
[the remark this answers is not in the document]
fwitte: Implemented

ibell:
There is another step needed after the change to config to fix the type stubs, but I am going to merge the PR and then fix it in master

PR #3138, opened 2026-06-09
docs(HSFlash): render the Performance-map figure live instead of showing only the script (GH #3136)

### LINKED ISSUE #3136: [ISSUE] New HSFlaash Doc Missing Performance Map Plot
### Description

Documentation page created in #2999 is generated with a Jupyter notebook.  Last section on [Performance map](https://coolprop.github.io/devdocs/coolprop/HSFlash.html#Performance-map) indicates "_The figure below_" the text, but the HTML rendering of the page only shows the script to generate the plot generated with ``HSFlashTimingMap.py`` and does not actually run the script and generate an image.  Possibly, the plot should not be in the Markdown cell with the text but in a separate, subsequent "code" cell type?

**Expected behavior:** Text indicates there should actually be a generated plot on that page.

**Actual behavior:** Only the generating bash script shown.

### Versions

**CoolProp Version:** 7.2.1dev

### Additional Information

Image

ibell:
Fixes #3136.

## Problem

The **Performance map** section of `Web/coolprop/HSFlash.ipynb` put the
figure-generating commands inside a *markdown* cell. The docs build executes
notebooks via `nbconvert --execute` but only installs the Python wheel — it
never builds the C++ `CatchTestRunner` — so the page rendered the script text
and produced **no image**, even though the prose said "the figure below."

## Fix

Replace the static script block with a **live Python code cell** that:

- sweeps water's single-phase `(log p, T)` grid (every `(p, T)` is single-phase,
  so `(h, s)` is read off a `P,T` flash),
- times each `H,S` flash through the low-level `AbstractState` interface
  (`min` over a few repeats to reject scheduler noise), and
- plots solve-time heatmaps in `p,T` and `h,s` coordinates with the saturation
  curve / two-phase dome and critical point overlaid (boundary computed inline
  from CoolProp — no dependency on the `HSFlashTimingMap` module being packaged).

The eval-count and per-leg panels in the richer map come from instrumentation in
the C++ `[HS_watermap]` test harness, which isn't available at docs-build time,
so they can't be regenerated in-notebook. The prose is trimmed to scope the live
figure to **solve time** and points readers at the C++ tooling
(`[HS_watermap]` + `CoolProp.Plots.HSFlashTimingMap`) for the full map.

## Verification

Built a wheel from this branch's HEAD (`gitrev e5ff16b51`) and ran it
docs-faithfully:

- full 4900-point grid sweeps in **~0.8 s** with **0 hangs** (4896 solved, 4
  near-edge points caught and skipped),
- the whole notebook executes end-to-end via `nbconvert --execute` with **no
  cell errors**,
- the figure cell emits an embedded `image/png` (`median 31.7 µs, p95 36.5 µs,
  max 44.4 µs` — uniform, no heavy tail).

A point that hung an older local build (`p≈30 kPa, T≈299 K`) solves in 16 ms on
current HEAD; the hang was a since-fixed bug in that stale build, not in the
shipped source.

## Summary by CodeRabbit

* **Documentation**
  * Refined explanatory text for the robust H,S flash algorithm, improving clarity and typography throughout.
  * Clarified strategy descriptions, individual stage (leg) notes, and acceptance/uniqueness conditions.
  * Expanded Performance Map guidance to explain live figure generation, provenance of attribution data, and updated reproduction commands.
  * No functional or public-interface changes.


Web/coolprop/HSFlash.ipynb
```diff
@@ -245,21 +245,110 @@
...
+    "the live figure below therefore shows solve time only. Across 130 pure fluids and ~190k\n",
+    "single-phase points the full map shows the cascade solving **100%** with **0\n",
+    "regressions** against the legacy path, at a mean of ~16 EOS evaluations and a worst case\n",
+    "under 40 (vs the legacy heavy tail of ~0.5 s on the hardest near-critical points). To\n",
+    "reproduce it:\n",
     "\n",
     "```\n",
     "./build_rel/CatchTestRunner \"[HS_watermap]\"          # writes water_hs_map.csv\n",
-    "python -m CoolProp.Plots.HSFlashTimingMap water_hs_map.csv   # --coords pt,hs,both\n",
+    "python -m CoolProp.Plots.HSFlashTimingMap water_hs_map.csv   # --coords {pt,hs,both}\n",
```
ibell, line 269: Why does this still exist if the generation is available in the notebook itself?

Web/coolprop/HSFlash.ipynb
```diff
@@ -245,21 +245,110 @@
...
+    "        try:\n",
+    "            ref.update(CP.PT_INPUTS, p, float(T))\n",
+    "            h, s = ref.hmolar(), ref.smolar()\n",
+    "            best = np.inf\n",
+    "            for _ in range(REPS):          # min over repeats rejects scheduler noise\n",
+    "                t0 = time.perf_counter()\n",
+    "                state.update(HS, h, s)\n",
+    "                best = min(best, time.perf_counter() - t0)\n",
+    "        except Exception:\n",
+    "            continue                       # skip any point the P,T or H,S flash rejects\n",
```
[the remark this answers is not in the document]
ibell: Instrumented in bd168a8. The skip path now counts skipped points and reports the distinct exception types in the cell output, e.g.:

```
4896 single-phase points timed, 4 skipped; median 32.3 us, p95 37.1 us, max 116.1 us
  skipped-point exceptions: 4xValueError
```

I deliberately kept the catch broad rather than narrowing it to specific types. This cell runs under `nbconvert --execute` in the docs build; a narrowed catch that lets an unexpected-but-benign edge point propagate would fail the docs build and replace the figure with a traceback — strictly worse for a docs page than skip-and-report. With the count + exception-type summary now in the output, a mass-failure or a new failure mode is visible (and the downstream `usg.max()`/`np.median` reductions still hard-fail loudly if the sweep ever returns *zero* points, so a silent empty figure isn't possible).

Web/coolprop/HSFlash.ipynb
ibell: It's not redundant with the live figure: that C++ path produces the *richer* map — per-point EOS-evaluation count and which cascade leg (saturation-anchor / supercritical-isentrope / ideal-gas-departure) solved each point — and those two quantities come from instrumentation in the `[HS_watermap]` test harness, not the public Python API. So the notebook's live cell can only color by solve *time*; it structurally can't regenerate the eval-count or per-leg panels. The block is the recipe to reproduce that developer diagnostic.

Per our discussion I'm leaving it as-is. If you'd rather it not sit on the user docs page, say the word and I'll trim it to a one-line pointer (or drop it entirely).

Issue #1342, opened 2016-11-18
Discontinuities in mixture Gibbs free energies


amlimaye:
I'm using CoolProp to compute the excess Gibbs free energy of a binary mixture of methanol and benzene. This is defined as:

`G_{excess} = G_{real} - G_{ideal solution}`

The ideal solution free energy assumes no enthalpy of mixing, but fully accounts for an entropy of mixing. That is, for a binary solution with mole fractions `x` and `1-x` for each component, the ideal solution Gibbs free energy is:

`G_{ideal solution} = x*G_{pure,1} + (1-x)*G_{pure,2} + RT[xlog(x) + (1-x)log(1-x)]`

By definition then, the excess Gibbs FE should be zero at x = 0.0 and x = 1.0. With this in mind, I get some odd results using CoolProp. I am studying a binary mixture of methanol and benzene and see the following behavior:
![methanol-benzene-example](https://cloud.githubusercontent.com/assets/15314835/20450227/cb9eeb3e-adbc-11e6-9308-81f54f67af9f.png)

CoolProp throws errors for methanol fractions below ~0.35. Clearly, the excess free energy curve is discontinuous at a methanol fraction of roughly 0.42. Above this, the excess free energies are actually in good agreement with the experimental values (Table II of this paper: http://pubs.acs.org/doi/pdf/10.1021/ja01056a002), as shown below:
![methanol-benzene-example-correct-region](https://cloud.githubusercontent.com/assets/15314835/20450310/74b32f28-adbd-11e6-9576-8895b0c1534a.png)

Any idea what may be causing this issue? I see similar behavior for the acetone-ethanol system:
![ethanol-acetone-example](https://cloud.githubusercontent.com/assets/15314835/20450322/8cbb981c-adbd-11e6-9724-cc58e7744b71.png)

All code used to generate these plots is in this Gist: https://gist.github.com/amlimaye/09e105c7e4bb66a529589cdb2a2ad167

Thanks in advance for taking a look!

ibell:
This is almost certainly a problem with the TP-flash calculations.  If you
look at the errors, I bet most of these are related to that.  Do you know
also that there is a gibbsmolar_excess function in the low-level interface
:
http://www.coolprop.org/dev/apidoc/CoolProp.CoolProp.html#CoolProp.CoolProp.AbstractState.gibbsmolar_excess

Do you have access to REFPROP?  I would be curious to see what the
performance looks like with REFPROP.  It might be quite a bit better.

[...]

PR #3140, opened 2026-06-09
fix(flash): select stable phase via Gibbs comparison in blind mixture single phase, fix the envelope guided two-phase path 

andr1976:
… PT flash

The blind flash single-phase branch used SRK to pick between gas and liquid density roots but always tried gas first, falling back to liquid only if the gas HEOS solve threw.  When both roots converged (e.g. N2/O2 at 75 K / 1 bar, which is subcooled liquid), SRK picked gas — the wrong phase.

Solve SRK cubic for both gas and liquid roots.  When both are valid, solve both HEOS densities and pick the phase with lower Gibbs energy via calc_gibbsmolar_nocache, mirroring the existing pattern in solver_rho_Tp_global().  When only one root is valid, the cost is identical to the old code (one SRK + one HEOS solve).

Also includes PE-guided flash refactoring: use closed envelope to classify (T,p) points, fall through to blind flash for unclosed envelopes, and cross-check tests for gas/liquid/two-phase conditions.

### Requirements

### Description of the Change

[ *We must be able to understand the design of your change from this description. If we can't get a good idea of what the code will be doing from the description here, the pull request may be closed at the maintainers' discretion. Keep in mind that the maintainer reviewing this PR may not be familiar with or have worked with the wrapper code being implemented, so please walk us through the concepts.* ]

### Benefits

[ *What benefits will be realized by the code change?* ]

### Possible Drawbacks

[ *What are the possible side-effects or negative impacts of the code change?* ]

### Verification Process

*Describe the actions you performed (e.g. text you typed, commands you ran, etc.), and describe the results you observed.*  

*Please attach here any python test sessions or image captures of testing in other wrappers.* ]

### Applicable Issues

[ *Enter any applicable Issues here.  Use ``Closes #????`` if this PR closes an open issue.* ]

## Summary by CodeRabbit

* **New Features**
  * Phase-envelope closure is now tracked and used to enable an envelope-guided PT flash fast path when a closed envelope is available.

* **Bug Fixes**
  * Improved single-/two-phase selection logic, clearer phase/Q assignment, and robust handling of near-pure or trivial two‑phase cases with reliable fallbacks.

* **Tests**
  * Expanded regression tests verifying envelope-guided vs blind flash agreement, subcooled liquid selection, fall-through scenarios, and stability at challenging compositions.

### DIFF
--- a/src/Tests/CoolProp-Tests-Michelsen.cpp
+++ b/src/Tests/CoolProp-Tests-Michelsen.cpp
@@ -576,7 +576,41 @@ TEST_CASE("Blind PT flash: Nitrogen/Oxygen [0.79/0.21]", "[michelsen][blind][fla
         AS->set_mole_fractions(z);
         CHECK_NOTHROW(AS->update(PT_INPUTS, 1e5, 75.0));
         CHECK(AS->Q() == -1);
-        CHECK(AS->rhomolar() > 0);
+        CHECK(AS->rhomolar() > 20000);  // liquid density — Gibbs selection picks liquid root
+    }
+}
+
+TEST_CASE("Blind flash: N2/O2 subcooled liquid phase selection", "[michelsen][blind][flash]") {
+    // N2/O2 at 75K, 1 bar: below bubble point (78.76 K).
+    // Previously the blind flash selected the gas root because SRK
+    // tried gas first and both roots converged.  The Gibbs-comparison
+    // fix selects the thermodynamically stable liquid root.
+    std::vector<double> z = {0.79, 0.21};
+
+    auto AS = std::shared_ptr<CoolProp::AbstractState>(CoolProp::AbstractState::factory("HEOS", "Nitrogen&Oxygen"));
+    AS->set_mole_fractions(z);
+    AS->update(CoolProp::PT_INPUTS, 1e5, 75.0);
+    CHECK(AS->Q() == -1);
+    CHECK(AS->phase() == CoolProp::iphase_liquid);
+    CHECK(AS->rhomolar() > 20000);  // liquid density
+
+    SECTION("Imposed liquid agrees with blind") {
+        auto AS_imp = std::shared_ptr<CoolProp::AbstractState>(CoolProp::AbstractState::factory("HEOS", "Nitrogen&Oxygen"));
+        AS_imp->set_mole_fractions(z);
+        AS_imp->specify_phase(CoolProp::iphase_liquid);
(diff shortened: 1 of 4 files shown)


ibell:
**Fail-open `inf` at two interior compositions (methanol–benzene, single-phase liquid)**

Testing this PR against issue #1342, the acetone–ethanol vapor-root regression is fully fixed (liquid root restored, physical excess-Gibbs curve). 👍

However, the methanol–benzene system surfaces a behavior change worth a look. At **308.15 K / 101325 Pa**, two interior compositions return a **non-finite result without throwing**:

```
x_methanol = 0.56  ->  Dmolar = inf,  Gmolar = inf   (PropsSI returns success, no exception)
x_methanol = 0.78  ->  Dmolar = inf,  Gmolar = inf   (PropsSI returns success, no exception)
```

These are unambiguously single-phase **liquid** states — the neighbours compute cleanly as liquid (x=0.54 → G^E≈0.306, x=0.58 → 0.296 kcal/mol; mix ρ≈14600 mol/m³ nearby).

On **current `master`** these same two points instead **raise** (`molar density of -1222.49 mol/m3 is below the minimum…` / `…-94019.55…`). So this PR trades *"throws an exception"* for *"silently returns Inf"* at these points. A caller gets `Inf` back from `PropsSI` with no error to catch — fail-open rather than fail-closed.

Repro (C++ `AbstractState`/`PropsSI`, but reproduces identically via the Python `PropsSI`):

```python
for x in (0.56, 0.78):
    f = f"HEOS::methanol[{x:.8f}]&benzene[{1-x:.8f}]"
    print(x, CP.PropsSI("Dmolar", "T", 308.15, "P", 101325, f))  # -> inf
```

Not a blocker for the core fix, but since the flash here can't find the (existing) liquid root, returning `Inf` silently is probably worse than the previous exception — either recover the liquid root or propagate a failure. Flagging so it can be folded in.

— Reported by Claude (automated reproduction against this PR branch built from source).

src/Tests/CoolProp-Tests-Michelsen.cpp
```diff
@@ -3,6 +3,7 @@
 #    include "AbstractState.h"
```
ibell: public headers of CoolProp now have the `CoolProp/` prefix (as of v8). You should be getting deprecation warnings at compile-time

ibell (changes requested):
You need a clang-format and to address deprecation warning, but otherwise if tests pass, good with me

ibell:
@andr1976 are you ready for this PR?

andr1976:
@ibell ready as can be 😀

PR #3141, opened 2026-06-09
perf(transport): cache ECS reference fluid (was rebuilt per viscosity/conductivity call) — ~13× on ECS-fluid transport

ibell:
## The bug

`calc_viscosity_contributions()` and `calc_conductivity_contributions()` construct a **fresh** reference-fluid `HelmholtzEOSMixtureBackend` for the ECS model on **every** `viscosity()`/`conductivity()` call:

```cpp
shared_ptr<HelmholtzEOSMixtureBackend> ref_fluid = std::make_shared<HelmholtzEOSMixtureBackend>(reference_fluid);
```

Constructing one **deep-copies the entire reference-fluid EOS** (every coefficient array, via `memmove`). So any repeated transport evaluation on an **ECS fluid** (R32, R245fa, and many refrigerants) pays that full copy each time.

It was found via profiling slow SVDSBTL surface builds: those call `conductivity()` once per grid cell (NT×NR = 160k cells/surface), and **~75% of an ECS-fluid surface build was this per-call reference-fluid reconstruction** (`memmove` + malloc churn) — not EOS math, not the SVD.

## The fix

Cache the reference-fluid backend (two `shared_ptr` members, built once on first use, reused). The ECS routines reset the reference state on every call (`conformal_state_solver` + `update_DmolarT_direct`, which `clear()`s all derived state), so reuse is **observationally identical** to constructing fresh.

## Effect

SVDSBTL cold surface build (the heaviest repeated-transport workload), under ASan:

| Fluid | Before | After | Speedup |
|---|---|---|---|
| R32 (ECS) | 856s | 68s | **12.6×** |
| R245fa (ECS) | 871s | 65s | **13.4×** |
| Water (hardcoded κ) | 130s | 130s | unchanged |
| CO2 (hardcoded κ) | 107s | 108s | unchanged |

Native: ~3.4× on ECS fluids. Non-ECS fluids (hardcoded conductivity) are untouched. This is a **general CoolProp win** — any repeated transport eval on an ECS fluid gets faster, not just SVDSBTL.

## Correctness

- All **347 non-slow Catch2 cases pass (132,467 assertions)**; transport check-values unchanged.
- Adversarial review traced every state-read of the cached reference: each is preceded by a full reset, so results are bit-identical to fresh construction.
- Per-*instance* cache (matching the existing `transient_pure_state`/`critical_state` members); concurrent transport on a single backend was already unsupported. The one concurrent consumer — the SVDSBTL parallel sampler — uses per-*thread* `AbstractState::factory` instances, so each thread builds its own reference; no sharing, no race. Cleared in `set_components` so a post-construction component swap rebuilds it. Consistent with the existing `transient_pure_state`/`critical_state` cache members.
- ECS branch is `is_pure_or_pseudopure`-only; cache stays null on mixtures.

Note (non-blocking): like the sibling temporary-state caches, these members aren't reset in `set_components` — safe today since `set_components` is constructor-only.

CoolProp-a7je. Surfaced while investigating slow SVDSBTL ASan CI (CoolProp-dq9u).

## Summary by CodeRabbit

* **Refactor**
  * Improved viscosity and conductivity computations for Helmholtz EOS mixtures by introducing cached reference-state reuse, reducing repeated work and memory churn for repeated property calls.
  * No changes to public APIs or interfaces; behavior and outputs remain the same.

### DIFF
--- a/src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
+++ b/src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
@@ -103,6 +103,12 @@ HelmholtzEOSMixtureBackend::HelmholtzEOSMixtureBackend(const std::vector<CoolPro
 }
 void HelmholtzEOSMixtureBackend::set_components(const std::vector<CoolPropFluid>& components, bool generate_SatL_and_SatV) {
 
+    // Drop the cached ECS transport reference fluids: they are tied to the
+    // OLD component's reference-fluid name, so a post-construction component
+    // swap must rebuild them.  (No-op at construction, where they are null.)
+    viscosity_ecs_reference_state.reset();
+    conductivity_ecs_reference_state.reset();
+
     // Copy the components
     this->components = components;
     this->N = components.size();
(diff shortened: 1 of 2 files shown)


ibell:
Thanks @coderabbitai — addressed both:

**2. Reset caches in `set_components` (h:79) — fixed.** `set_components` is `public`/virtual, so even though it's constructor-only today, a post-construction component swap would leave stale ECS references. Added at the top of `HelmholtzEOSMixtureBackend::set_components`:
```cpp
viscosity_ecs_reference_state.reset();
conductivity_ecs_reference_state.reset();
```
(No-op at construction, where the pointers are already null.)

**1. Per-instance cache / concurrency (cpp:847) — by design, no change.** It is a per-*instance* cache, but this introduces no new race class: concurrent `viscosity()`/`conductivity()` on a **single** `HelmholtzEOSMixtureBackend` already race on the backend's own mutable state (`_T`, `_rhomolar`, the derivative caches cleared on every `update`) — CoolProp backends have never been thread-safe for shared-instance concurrency. The cache follows the existing `transient_pure_state`/`critical_state`/`TPD_state` shared-`shared_ptr`-member pattern, and the one concurrent consumer (the SVDSBTL parallel sampler) builds a fresh per-thread backend (`AbstractState::factory`, not a value-copy), so each thread lazily builds its **own** reference — no sharing, no race. Per-call construction would just reinstate the bug this fixes. I've reworded the PR's "per-thread" phrasing to "per-instance (and the parallel sampler uses per-thread instances)" to be precise.

Verified: 347 non-slow Catch2 cases pass (132,467 assertions); transport values unchanged.

PR #3154, opened 2026-06-10
fix(flash): legacy PT-flash reports correct vapor quality (not pinned at 0.5)

ibell:
## Summary

`SaturationSolvers::PTflash_twophase::solve_legacy()` solved the Newton system for the liquid (`x`) and vapor (`y`) compositions directly but **never wrote `IO.beta`** (the vapor molar fraction). Only `solve_michelsen()` set it. The `PTflash_twophase_options` struct defaults `beta = 0.5`, and the #3026 rewrite of `PT_flash_mixtures` consumes `IO.beta` as the vapor quality (`_Q = o.beta`) and the bulk two-phase density `1/(beta/rho_v + (1-beta)/rho_l)`.

**Result:** any blind PT flash on a two-phase mixture with `MIXTURE_STABILITY_ALGORITHM=0` (legacy) silently returned **Q = 0.5 and a wrong bulk density** for every point.

## Fix

Recover `beta` at the end of `solve_legacy()` from the converged overall mass balance `z_i = (1-beta) x_i + beta y_i`. The amount-of-substance residuals in `build_arrays()` drive `(z_i - x_i)/(y_i - x_i)` to a single common value across components, so any component yields the same `beta`; we pick the one with the largest phase separation `|y_i - x_i|` for numerical robustness, and fall back to `beta = 0` (single-phase) on a degenerate/trivial split. Same convention as `solve_michelsen()`. The default Michelsen path is untouched.

## Test

New `[PTflash_twophase]` regression test: a `Methane&Ethane` point built at `Q = 0.25` via a PQ flash is recovered by **both** the legacy and Michelsen algorithms (agreement < 0.01), where before the legacy path was pinned at 0.5. Config is restored via an RAII guard.

## Verification

- `[PTflash_twophase]`, `[flash]`, `[mixture]`, `[mixtures]`, `[stability]`, `[predefined_mixtures]` all pass.
- Adversarial review: no blocking findings (math/convention confirmed against `solve_michelsen` and the `FlashRoutines` consumer; degenerate-split guard and out-of-range β handling judged fail-safe).
- preflight: clang-format / build / Catch2 (`[Helmholtz],[REFPROP]`) / semgrep pass. The cppcheck and clang-tidy steps report only **pre-existing** findings in adjacent code (`VLERoutines.h` constructor member-init; `omega_local` float-loop counters; the `build_arrays` branch-clone; a `static_cast` in the pre-existing PTflash test) — this change introduces **zero** new findings in either, and both checks are `continue-on-error`/informational in CI.

Fixes CoolProp-1tbe.1

## Summary by CodeRabbit

* **Bug Fixes**
  * Legacy PT flash now computes and records vapor quality from final phase compositions and avoids numerical issues when phase compositions are nearly identical.

* **Tests**
  * Added a unit test that verifies vapor quality recovery across stability-detection modes, preventing regression where quality could remain stuck at 0.5.

### DIFF
--- a/src/Backends/Helmholtz/VLERoutines.cpp
+++ b/src/Backends/Helmholtz/VLERoutines.cpp
@@ -2702,6 +2702,33 @@ void SaturationSolvers::PTflash_twophase::solve_legacy() {
             throw ValueError(format("PTflash_twophase::call reached max number of iterations [%d]", IO.Nstep_max));
         }
     } while (this->error_rms > 1e-9 && min_rel_change > 1000 * DBL_EPSILON && iter < IO.Nstep_max);
+
+    // Recover the vapor molar fraction (beta) from the converged overall mass
+    // balance z_i = (1 - beta) x_i + beta y_i.  The amount-of-substance
+    // residuals in build_arrays() drive (z_i - x_i)/(y_i - x_i) to a single
+    // common value across all components, so any component yields the same
+    // beta; pick the one with the largest phase separation |y_i - x_i| for
+    // numerical robustness.  Without this the caller (PT_flash_mixtures) reads
+    // the default beta = 0.5 and reports a wrong vapor quality and bulk density
+    // for every two-phase mixture (CoolProp-1tbe.1); solve_michelsen() sets
+    // IO.beta with the same convention.
+    std::size_t i_best = 0;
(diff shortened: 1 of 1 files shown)


andr1976:
👍

PR #3169, opened 2026-06-11
fix(backends): clear cppcheck uninitMemberVar warnings across core backends

ibell:
## Summary

A `cppcheck --enable=warning` sweep over **all C/C++ changed since `v7.2.0`** (247 files, same invocation as `dev/ci/preflight.sh`) surfaced **30 `uninitMemberVar` warnings**. This clears the **core-library** ones by giving each flagged member a default member initializer (NSDMI), or — for `L0CurveTracer::theta_last` — extending the existing constructor init-list.

Every change is **behavior-neutral**: the default only takes effect on a read-before-write path, which was undefined behavior before. Two of them (`EnvironmentalFactorsStruct`, `PTflash_twophase::error_rms`) additionally retire a *latent* uninitialized-read: the new deterministic value (`_HUGE`) is rejected by the existing `ValidNumber`/`<0` guards or yields the correct keep-iterating semantics, where prior garbage could have slipped through.

Default convention: `_HUGE` for physical-quantity doubles (codebase idiom), `0`/`false`/`nullptr`/enum-sentinel otherwise.

### Structs touched
`ViscosityModifiedBatschinskiHildebrandData`, `ViscosityFrictionTheoryData`, `ViscosityRhoSrVariables`, `EnvironmentalFactorsStruct`, `CubicsValues`, `UNIFACMixture`, `SaturationResidual`, `newton_raphson_saturation`, `PTflash_twophase`, `L0CurveTracer`.

### Deliberately out of scope
- **`superancillary.h` `ChebyshevExpansion::m_xmin/m_xmax`** — fixed on #3162, which already reworks that header.
- **`IncompressibleBackend::fluid`** — cppcheck false positive (both flagged constructors `throw` before the member is read).
- **Non-shipping dev tools** (`dev/fitter/Fitter.cpp` self-assign) and **legacy wrappers** (EES buffer/return-value, Modelica copy-ctor) — left as-is; not built in CI.

## Validation

- **cppcheck core `uninitMemberVar`: 30 → 0.**
- Build links clean; `[Helmholtz]`, `[REFPROP]`, `[cubic]`, `[critical_points]` tests pass (470 assertions, 0 failures).
- clang-format clean on all six files.
- Adversarial review (per CLAUDE.md) returned no blocking findings; it verified behavior-neutrality on every post-constructor read path and confirmed `_HUGE` scope + enum sentinels.

> Note: `dev/ci/preflight.sh`'s clang-tidy step runs **whole-file** and flags pre-existing debt in these large legacy backends (make-shared, static_cast downcast, implicit-bool on lines this diff never touches). There are **zero** clang-tidy findings on this diff's changed lines; CI's clang-tidy is diff-only + informational and is unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initialization of environmental property data to avoid uninitialized values (GWP, ODP, etc.).
  * Enhanced stability and reliability of viscosity and transport calculations via default initialization of numeric parameters.
  * Increased robustness of cubic and Helmholtz backend solvers and mixture models by initializing solver and cache state members.

### DIFF
--- a/include/CoolProp/CoolPropFluid.h
+++ b/include/CoolProp/CoolPropFluid.h
@@ -34,7 +34,7 @@ struct BibTeXKeysStruct
 
 struct EnvironmentalFactorsStruct
 {
-    double GWP20, GWP100, GWP500, ODP, HH, PH, FH;
+    double GWP20 = _HUGE, GWP100 = _HUGE, GWP500 = _HUGE, ODP = _HUGE, HH = _HUGE, PH = _HUGE, FH = _HUGE;
     std::string ASHRAE34;
 };
 struct CriticalRegionSplines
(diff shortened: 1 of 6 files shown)


ibell:
Addressed both items:

**CodeRabbit (uninitialized `newton_raphson_saturation` members).** Added in-class `= _HUGE` initializers for `error_rms, rhomolar_liq, rhomolar_vap, T, p, min_rel_change` and `dTsat_dPsat / dPsat_dTsat` (commit 7f088f6), matching the treatment already applied to `PTflash_twophase::error_rms` here. `[flash]`/`[stability]`/`[mixture]` suites stay green.

**Installed-header hygiene gate failure.** Root cause was *not* in this PR — `fluids/Helmholtz.h` used `memset` without `<cstring>`, which libstdc++ (CI's g++) doesn't pull in transitively, so the self-containedness gate from #3162 failed across **master and every open PR**. Fixed in #3173 (now merged). This PR's gate runs against the branch merged with master, so the next run picks up #3173 and goes green.

PR #3170, opened 2026-06-11
fix(flash): converge the Michelsen mixture PT phase-split solver (#3168)

### LINKED ISSUE #3168: solve_michelsen returns grossly-unconverged two-phase splits for several mixture PT flashes
### Summary

`SaturationSolvers::PTflash_twophase::solve_michelsen()` (the second-order Gibbs phase-split solver added in #3026) can **return a two-phase split that has not converged** — the phase fugacities differ by factors of e¹–e⁶ (≈2.7×–400×), nowhere near equilibrium — and `FlashRoutines::PT_flash_mixtures` then publishes it as a valid result. So `update(PT_INPUTS, …)` returns a **wrong two-phase composition / quality** for several mixture states while reporting success.

### How it was found

`solve_michelsen`'s Gibbs-minimization loop exits via the iteration cap (30) or a stalled line search (`!step_ok`) with **no check that the convergence criterion (`max |Δ(ln f)| < 1e-9`) was actually met**, then publishes `IO.beta / IO.x / IO.y / IO.rhomolar_liq/vap` regardless. (The legacy `solve_legacy` solver threw at its iteration cap.)

Adding a post-loop convergence check (throw when a non-converged exit leaves `max |Δ(ln f)| > 1e-7`, i.e. worse than the SS tolerance) made **~13 existing tests across `[flash]` and `[michelsen]` fail**, reporting `max |Δ(ln f)|` values of **0.x to 6**. Examples (`src/Tests/CoolProp-Tests-Michelsen.cpp`):

```
update(PT_INPUTS, 4e6, 190.0)     -> did not converge (max |Δ(ln f)| ≈ 1)
update(PT_INPUTS, 4e6, 190.0)     -> did not converge (max |Δ(ln f)| ≈ 6)
update(PT_INPUTS, 101325, 308.15) -> did not converge (max |Δ(ln f)| ≈ 0.x)
```

These tests **pass today** because they only assert `phase() == iphase_twophase` and `Q() > 0` — not composition or equilibrium accuracy. So the wrong splits slip through CI.

### Why this matters

A `max |Δ(ln f)|` of 1–6 means the equal-fugacity equilibrium condition is violated by orders of magnitude: the reported liquid/vapor compositions and the resulting quality `Q` are simply wrong for the affected (T, p, z). It is a silent-wrong-answer bug — no error, no warning.

### Reproduction

1. Build `CatchTestRunner` (see `CLAUDE.md`).
2. In `solve_michelsen` (`src/Backends/Helmholtz/VLERoutines.cpp`), hoist a `bool converged`/`CoolPropDbl last_max_g` out of the Gibbs loop, set `converged = true` at the `max_g < 1e-9` break, capture `last_max_g = max_g` each iteration, and after the loop `throw` when `!converged && last_max_g > 1e-7`.
3. `./build_catch/CatchTestRunner "[michelsen]"` → ~7 failures with the "did not converge" message; `"[flash]"` → ~6 more.

### Options (maintainer decision)

- **(a) Throw on non-convergence** — restores the legacy contract and stops shipping wrong numbers, but turns the affected flashes (which #3140's tests deliberately make "pass") into errors — a capability regression.
- **(b) Warn + return** — non-breaking, but still returns the wrong composition.
- **(c) Fix the solver to converge** these cases (the GDEM acceleration / line-search / Hessian) — the proper fix.

Either way the smoke tests at `CoolProp-Tests-Michelsen.cpp:82-92, 955` (and siblings) should be tightened to assert equilibrium accuracy (e.g. `max |Δ(ln f)|` below a tolerance, or composition vs a reference) so this can't regress unnoticed.

### Context

- Solver introduced in #3026 (multicomponent TP-flash rewrite).
- Smoke tests that accept the unconverged splits added in #3140.
- Related hardening (the cubic/GDEM guards + the line-search density state-restore) is in PR #3167; the convergence diagnostic patch is described there and was intentionally left out pending this decision.

_Found during the v8 pre-release flash-robustness audit._

andr1976:
Fixes #3168. Carries item 4c of the #3026 flash audit (#3167).

### Description of the Change
`SaturationSolvers::PTflash_twophase::solve_michelsen()` (the second-order
phase-split solver from #3026) could return a two-phase split that had not
converged — phase fugacities differing by factors of e¹–e⁶ — which
`PT_flash_mixtures` then published as a valid result. Phase 2 exited on its
iteration cap or a stalled line search with no check that the equal-fugacity
condition held, so `update(PT_INPUTS, …)` returned a wrong two-phase
composition / quality while reporting success.

Root cause: the Phase-2 step had no conditioning or proper step control (plain
Newton + backtracking). Near the mixture critical point and for wide-boiling
mixtures the reduced Hessian is positive-definite but badly conditioned, so
steps were feasibility-clamped to a crawl (hit the cap) or stalled. The
gradient/Hessian formulas were correct (verified vs Michelsen 1982b App. B and
the in-tree stability solver `minimize_tpd`); the gap was conditioning + step
control + the missing convergence check.

Implements option (c) (fix the solver) with (a) (throw on non-convergence) as
a safety net:
- **solve_michelsen Phase-2 rewrite**: scale the Newton system by
  `dia_i = sqrt(z_i/(x_i*y_i))`; replace plain-Newton + backtracking with a
  Hebden restricted-step (trust-region) that adds a positive-definiteness
  diagonal shift and bounds the step; recompute the equal-fugacity residual on
  the final published state and add a convergence gate that throws (flagged via
  `o.nonconvergence`) when `max|ln f_V - ln f_L| > 1e-7`.
- **PT_flash_mixtures single-phase fallback**: when a flagged split fails
  specifically via the convergence gate (usually a stability false-positive at
  a single-phase state), fall back to a single-phase density solve. Other
  errors still propagate — not masked as single-phase.
- **GDEM NaN guard (#3167 item 4c)**: add `!ValidNumber` guards at both GDEM
  sites; initialize the stability site's `esq_pair`.
- **Tests**: assert the published split is at equilibrium (independently
  recomputed `max|ln f_V - ln f_L| < 1e-6`) on the #3168 cases; assert
  methanol/benzene at 308.15 K / 1 atm resolves to single-phase liquid.

### Benefits
- The headline cases (7-component natural gas at 190 K / 4 MPa, SRK & PR) now
  genuinely converge to ~7e-10 (were stuck at ~2e-7 at the cap).
- No silent wrong two-phase compositions; smoke tests assert equilibrium
  accuracy so this can't regress on a `phase()==twophase` check alone.

### Possible Drawbacks
- The single-phase fallback (scoped to the gate signal) would publish
  single-phase if a genuinely two-phase state ever failed that way; an
  1870-flash sweep hit it 0 times.
- methanol/benzene x=0.54 exposes a stability false-positive (separate latent
  issue); this PR makes the flash recover correctly.

### Verification Process
- `[michelsen]`/`[flash]`/`[mixture]`/`[stability]` green (1099 assertions);
  full non-slow suite green.
- Robustness sweep: 1870 flashes, 0 throws, worst equilibrium residual 1.3e-8.
- preflight clang-format / build / Catch2 pass; two rounds of adversarial review,
  all blocking findings addressed.

### Follow-ups (not in this PR)
- #3167's Cubic-backend items 3A/3B are left to that PR; its Finding 2
  (line-search density restore) is subsumed by the Phase-2 rewrite.
- Mixture PT-flash documentation (`Web/fluid_properties/Mixtures.rst`) to be
  updated separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of two-phase flash calculations, including safer fallback to single-phase when splits fail to converge.
  * Refined phase/density selection for single-phase cases and ensured trivial-quality splits map correctly back to single-phase.
  * Added clearer nonconvergence signaling via a new flash options flag, and tightened failure handling in the Michelsen two-phase routine.

* **Tests**
  * Strengthened Michelsen benchmark/regression coverage with stricter equilibrium residual checks and updated methanol–benzene assertions for problematic compositions.


andr1976:
**Merged `master` and reconciled with #3174**

#3174 landed on `master` first, and it edits the same blind `PT_flash_mixtures` block and the same `solve_michelsen` function as this PR, so the merge (`ebe6a862`) needed a manual reconciliation rather than a one-sided accept. Summary for reviewers:

**`FlashRoutines.cpp` (blind PT path) — combined, not replaced.**
The two PRs address different problems in the same block: this PR adds a convergence gate (`solve_michelsen` throws and sets `o.nonconvergence` on a non-converged split, which routes to a single-phase fallback), while #3174 adds a Wilson bubble/dew cross-check to recover stability false-negatives. The merge keeps both:
- this PR's `solve_single_phase` lambda and its `o.nonconvergence` → single-phase fallback on the genuine-unstable solve;
- #3174's Wilson cross-check (`guess_split_from_wilson` + `is_uncertain`) with the `wilson_seeded` speculative-split fencing and equal-fugacity verify;
- both fail-paths unified to `do_twophase = false` → a single `solve_single_phase()`.

Accepting `master` wholesale here would have left this PR's convergence gate throwing **uncaught**, since `master`'s genuine-unstable path calls `solver.solve()` with no `o.nonconvergence` catch.

**`VLERoutines.cpp` (`solve_michelsen`) — kept this PR's trust-region solver.**
The conflict was this PR's trust-region (Hebden) trial step vs `master`'s older line-search step; the rewrite is kept. #3174's `solve_trial_rho_warm` density warm-start auto-merges at the entry density solves (a hybrid); the bulk of #3174's warm-start (the stability test and `minimize_tpd`) merges cleanly and is retained.

**`CoolProp-Tests-Michelsen.cpp` — deduplicated a test helper.**
Both PRs had independently added an `equilibrium_residual` helper with the **same C++ signature but opposite `(p, T)` / `(T, p)` argument order**. Git auto-merged this without flagging a conflict, producing an ambiguous overload that does not compile (and where dropping either definition would silently swap p/T at the other's call sites). Standardized on #3174's trace-skipping body + `(T, p)` order and updated this PR's two call sites accordingly.

**Verification:** clean build; `[zgpy]` (39), `[stability_sweep]` (27, 0 misclassifications), and the full `[flash],[mixture],[michelsen],[stability],[cubic]` suite (1138 assertions) pass. Full detail is in the merge commit message (`ebe6a862`).

andr1976:
@ibell conflicts with master branch resolved

PR #3174, opened 2026-06-12
fix(flash): recover near-dew two-phase states misclassified as single-phase liquid (#3168)

andr1976:
Addresses @jakobreichert's report on #3168 (tracked internally as CoolProp-zgpy). His comment raised three things; this PR covers the misclassification and the speed; the fugacity item is noted under "out of scope". **Pre-existing on master, independent of #3170.**

### Description of the Change

For cubic mixtures (SRK/PR) at high vapor fraction, the Michelsen TPD stability test (`check_stability_michelsen`) reports a false **"stable"** verdict near the dew point, so the blind `PT_flash_mixtures` takes the single-phase path and selects the liquid root — publishing a genuinely two-phase state (Q ≈ 0.98) as **single-phase liquid**, with a tell-tale enthalpy discontinuity. Instrumenting the stability test pinned the causes, fixed across three commits:

1. **Wilson bubble/dew cross-check.** When the verdict is "stable", cross-check against a cheap Wilson K-factor estimate; if `Σ zᵢKᵢ > 1` **and** `Σ zᵢ/Kᵢ > 1`, attempt a split seeded from Wilson (liquid seeded at the **incipient, heavy-rich** composition — not the feed). *(Adapted from @jakobreichert's #2720 lever.)*

2. **Don't fail-open a non-conclusive stability verdict.** Near the dew point `minimize_tpd` can't find an acceptable step and returns *non-conclusively* — silently treated as "stable" (the dominant cause). The test now exposes `is_uncertain()`, and the blind flash forces a *verified* Wilson-seeded attempt when the verdict is uncertain (even where the ideal estimate doesn't bracket — the EOS-based successive substitution decides). The speculative split is fenced: any solver throw, a trivial (x≈y) result, **or** an equal-fugacity residual > 1e-7 falls back to single-phase.

3. **Performance — warm-start the trial-phase density solves** *(addresses jakob's third observation that mixture flashes felt slow, `solver_rho_Tp_global` vs `update_TP_guessrho`)*. The stability test and `solve_michelsen` re-ran the **global** density solver (spinodal scan + double Brent + Gibbs compare) per inner/outer iteration — hundreds of global solves per flash. The new `solve_trial_rho_warm` helper warm-starts from the previous root with a local Newton, falling back to the global stable-root solver on the first solve, on failure, **or when the density jumps > ~2×** (a spinodal crossing onto the other branch). So a metastable root is never silently accepted.

### Benefits

- **Correctness:** a blind, fugacity-checked sweep (3 mixtures × 3 backends) goes from **SRK 19 / PR 19 → 0 misclassifications**, zero trivial/unconverged splits; recovered points reach genuine equilibrium (fug_err ~1e-10).
- **Speed:** HEOS mixture flashes **~2.5–4.5× faster** (two-phase ~163→36 ms, single-phase ~70→28 ms; SRK barely changes — its density solve is already cheap).

### Verification Process

- `[flash],[mixture],[michelsen],[stability],[cubic]` — **1118 assertions pass.** Validation lives in the suite as a C++ sweep (`[stability_sweep]`, port of jakob's `scan_twophase_PT.py`) + a near-dew point test + superheated-gas single-phase guards.
- Mechanism confirmed by instrumentation; cross-checked against thermopack's `stability.f90` (decouples convergence tolerance from the instability decision; does not fail-open).
- Adversarial review of each commit; all findings addressed (the perf commit's metastable-root fail-open is closed by the density-jump guard, re-reviewed).
- **Perf is result-neutral:** no classification or density-branch change; HEOS converged values shift only at the ~1e-10 solver-tolerance level (same stable root); SRK/PR bit-identical.
- preflight: clang-format / build / json-symbols / install-headers / tests / semgrep pass; cppcheck verified clean with `-UENABLE_CATCH`; clang-tidy reports only pre-existing whole-file findings (none on changed lines; CI runs it diff-only + informational).

### Known / out of scope

- **FT-1 (correctness, forced path):** the equal-fugacity check proves a VLE *stationary* point, not global stability — a narrow, `uncertain`-gated metastable fail-open. A Gibbs-margin check was rejected (near the dew point the two-phase Gibbs gap → 0, so it would re-introduce false-*negatives*). Bounded blast radius; documented.
- A pre-existing HEOS **sub-triple-point** region still returns marginal two-phase results (≤1 point in the sweep) — that is #3168 / #3170 convergence-gate territory, unchanged here, and is also jakob's HEOS-fugacity observation.

### Note on overlap with #3170

Touches the same blind `PT_flash_mixtures` block as #3170, so whichever lands second needs a small reconciliation. They're independent: #3170 fixes solver *non-convergence* (the equal-fugacity gate); this fixes a stability *false-negative* upstream of the solver, plus the per-iteration density-solve cost.

Co-authored with @jakobreichert — the Wilson/bubble-point lever and SS-refinement helper are from his #2720, and the speed item is his observation.

## Summary by CodeRabbit

* **New Features**
  * Improved PT flash initialization with speculative two‑phase seeding from K‑factor estimates and refined saturation‑based split guesses; iterative density/composition refinement for more robust VLE starts.
  * Stability checks now expose non‑conclusive/uncertain outcomes rather than treating them as definitively stable.

* **Tests**
  * Added regression sweeps and targeted tests validating two‑phase classification, non‑trivial splits, and fugacity/equilibrium residuals near dew/bubble conditions.

### DIFF
--- a/src/Backends/Helmholtz/FlashRoutines.cpp
+++ b/src/Backends/Helmholtz/FlashRoutines.cpp
@@ -94,17 +94,66 @@ void FlashRoutines::PT_flash_mixtures(HelmholtzEOSMixtureBackend& HEOS) {
         // Blind flash call
         // Instantiates stability tester (dispatches to Michelsen or Gernert based on config)
         StabilityRoutines::StabilityEvaluationClass stability_tester(HEOS);
-        if (!stability_tester.is_stable()) {
+        bool do_twophase = !stability_tester.is_stable();
+        CoolProp::SaturationSolvers::PTflash_twophase_options o;
+        o.z = HEOS.get_mole_fractions();
+        bool wilson_seeded = false;
+
+        // CoolProp-zgpy: the Michelsen stability trials can converge to the trivial
+        // (feed) solution near a phase boundary -- notably for cubic mixtures at high
+        // vapor fraction -- and report a false "stable" verdict, so a genuinely
+        // two-phase state gets mislabeled single-phase liquid.  Cross-check a "stable"
+        // verdict against a cheap ideal (Wilson K-factor) estimate: if it places the
+        // feed strictly between its bubble and dew points, attempt a split seeded from
+        // that estimate.  A genuinely single-phase point either fails the Wilson test
+        // here or collapses back to single phase via the beta~0/1 fallback below.
+        // (Lever from jakobreichert's PR #2720.)
+        if (!do_twophase) {
+            try {
+                if (CoolProp::SaturationSolvers::guess_split_from_wilson(HEOS, o.x, o.y, o.rhomolar_liq, o.rhomolar_vap, o.z, HEOS.T(), HEOS.p(),
+                                                                         10)) {
+                    do_twophase = true;
+                    wilson_seeded = true;
+                }
+            } catch (const CoolProp::CoolPropBaseError&) {
+                // Speculative seed failed (e.g. no liquid density root at this state);
+                // trust the stability verdict and stay on the single-phase path.
+                do_twophase = false;
+                wilson_seeded = false;
+            }
+        }
+
+        if (do_twophase) {
             // There is a phase split and liquid and vapor phases are formed
-            CoolProp::SaturationSolvers::PTflash_twophase_options o;
-            stability_tester.get_liq(o.x, o.rhomolar_liq);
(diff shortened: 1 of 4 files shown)


andr1976:
@ibell fixes the cases @jakobreichert provided

PR #3177, opened 2026-06-12
Improve cubic performance

jakobreichert:
### Description of the Change

This change improves the performance of fluid mixture calculations with the cubic backends by implementing caching of aii_term for a given tau to avoid redundant alpha function calls. Additionally, a calc_all_terms() function is implemented to avoid redundant pow() and exp calls() for more expensive alpha function. Also, alpha_type and alpha_coeffs are now set when calling set_cubic_alpha_C(), so they are used for state copies by get_copy().

### Benefits

Speed up for SRK and PR with default alpha of approximately 2x.
Speed up for PR with Twu alpha function of approximately 6x.

### Possible Drawbacks

None. 

### Verification Process

Speed up has been tested with a python script, performing PT and PQ flashes for arbitrary 2-6 component mixtures in the twophase region (PT and PQ) and single phase region (PT).

[bench_cubic_flash_new.py](https://github.com/user-attachments/files/28886090/bench_cubic_flash_new.py)

Catch2 tests have been added (CoolProp-Tests-CubicAlpha.cpp) to verify calc_all_terms() produces bit-identical results to the scalar term().

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed missing parameter metadata updates for cubic equation alpha functions to ensure consistent behavior.

* **Improvements**
  * Added bulk derivative evaluation and caching for cubic alpha computations, improving consistency and performance of EOS derivative results.

* **Tests**
  * Added validation tests that verify alpha-function derivatives match scalar evaluations across representative conditions.


src/Backends/Cubics/CubicBackend.cpp
```diff
@@ -767,8 +767,12 @@ void CoolProp::AbstractCubicBackend::set_cubic_alpha_C(const size_t i, const std
     }
     if (parameter == "MC" || parameter == "mc" || parameter == "Mathias-Copeman") {
         get_cubic()->set_C_MC(i, c1, c2, c3);
+        components[i].alpha_type = "MathiasCopeman";
```
ibell: I don't get what is happening here; why is this required?

src/Backends/Cubics/GeneralizedCubic.h
```diff
@@ -29,6 +32,11 @@ class AbstractCubicAlphaFunction
    public:
     virtual ~AbstractCubicAlphaFunction() = default;
     virtual double term(double tau, std::size_t itau) = 0;
+    /// Compute all 5 tau-derivatives at once. Default falls back to 5 separate term() calls.
+    virtual void calc_all_terms(double tau, std::array<double, 5>& terms) {
```
ibell, line 49: Somewhere, possibly here, can you define what the "terms" are?

src/Backends/Cubics/GeneralizedCubic.cpp
```diff
@@ -118,6 +120,81 @@ double TwuAlphaFunction::term(double tau, std::size_t itau) {
     }
 }
 
+void BasicMathiasCopemanAlphaFunction::calc_all_terms(double tau, std::array<double, 5>& terms) {
```
ibell, line 123: This is a good idea; it should be significantly faster, even if you skip the cache

ibell (commented):
In the PR body, can you provide a bit more context for the performance gains you see? I guess they are at constant T,x, so that you hit the cache enough to see this speedup? I guess 100% cache misses would be somewhat faster but not crazily so?

src/Backends/Cubics/CubicBackend.cpp
jakobreichert: This is because when a state copy is generated with get_copy(), copy_internals() calls set_alpha_from_components(). So if the alpha_type and alpha_coeffs are not set here, every state copy uses the default alpha, which can cause issues.

jakobreichert:
> In the PR body, can you provide a bit more context for the performance gains you see? I guess they are at constant T,x, so that you hit the cache enough to see this speedup? I guess 100% cache misses would be somewhat faster but not crazily so?

I edited the PR text with some context and a (revised) python script for benchmarking. I get about the same speedup for PT and PQ flashes, which is a bit surprising now that I think about it.

ibell:
clang-format is needed (claude should figure it out), and there are some actionable comments from coderabbit bot review

ibell:
Thank you!

PR #3176, opened 2026-06-12
fix(HEOS): flash-consistency robustness — TOMS748 density/flash fallbacks + REFPROP max-density grid bound

ibell:
## Summary

Robustness fixes for HEOS single-phase flash solvers that eliminate large buckets of `ConsistencyReport` failures, plus a REFPROP-style max-density bound in the consistency-plot grid. Two commits:

1. **`solver_rho_Tp` supercritical_liquid TOMS748 fallback + grid bound** (`CoolProp-kf78`)
   - Near the critical isotherm the `iphase_supercritical_liquid` Brent bracket spanned the van der Waals loop and returned NaN ("p is not a valid number") in a thin band around Tc. Adds a TOMS748 dense-branch fallback (upper bound anchored on `rhomolar_reducing()` since `EOS().limits.rhomax` is 0/unset for some fluids; live state restored at the converged density).
   - `ConsistencyPlots.py`: a REFPROP-style maximum-density floor (triple-point liquid density) applied to the single-phase grid **for pure fluids at supercritical pressure only** — excludes the unvalidated compressed-solid/extrapolation region rather than testing it.

2. **DP / HS / D+X flash robustness** (`CoolProp-r1w7.1`, `CoolProp-vmp`)
   - **DP_flash**: unbounded Halley diverged to negative/out-of-range T at extreme (ρ,p) ("temperature of −926 K below minimum 0 K"). Now accepts Halley only on a converged, in-range root and otherwise falls back to a bracketed TOMS748 solve over `[Tmin, 1.5·Tmax]`.
   - **HS_flash legacy path**: the `h(SmolarT(s,T)) − h_target` residual is non-monotonic (the inner SmolarT solve degrades at high T), so endpoint-only `Brent(Tmin,Tmax)` reported "do not bracket" while the root sat just above Tmin. Now scans upward for the first sign change between *adjacent* valid samples and solves with TOMS748 (Brent fallback if none found).
   - **D+X melting-line floor**: strict `value < y_lo` rejected on-melting-line states on last-bit noise (e.g. `−810.862 < −810.862`); added a small relative tolerance.

All new TOMS748 brackets wrap interior/endpoint EOS evaluations so a throw degrades to a clean `ValueError` (or the Brent fallback) rather than escaping the flash.

## Impact (full 136-fluid `ConsistencyFigure`, 40×40, drift-free vs the branch base)

| | failures |
|---|---|
| base (this PR's master ancestor) | 2193 |
| after this PR | **893** |

[...]

PR #3180, opened 2026-06-12
fix(HEOS): supercritical fast path for D+{H,S,U} back-flashes (clears Air near-critical D+X failures)

ibell:
## Summary

`HSU_D_flash` (the D+{H,S,U} flash) routes **pseudo-pure** fluids' near-critical-density D+X back-flashes through the legacy "sad path", whose region logic is anchored on the **triple-point states** (unset for pseudo-pure fluids → default `-1`) and on `saturation_D_pure`, which is numerically **unstable near ρc**. For **Air**, `DmolarHmolar`/`DmolarSmolar`/`DmolarUmolar` at ρ/ρc ≈ 0.8–1.0 and T/Tc up to ~15 therefore throw **"p is not a valid number"**. The modern superancillary "happy path" that avoids this is gated on `is_pure()`, so pseudo-pure fluids can't use it.

This adds a **supercritical single-phase fast path** at the top of the legacy path: when the caloric input lies above its value on the critical isotherm at this density, the solution temperature is above Tc, so the state is unambiguously single-phase and needs no saturation. h, s and u all increase with T at fixed ρ, so the residual is bracketed directly on `[Tc, 1.5·Tmax]` and solved with TOMS748.

**Guards:**
- Restricted to **ρ < 2·ρc** — the Air failures sit at ρ/ρc ≤ 1.005, while staying clear of the dense-liquid extrapolation (ρ ≫ ρc) where the critical-isotherm probe is unreliable and could mis-fire.
- Commits only a state that **reproduces the caloric input** (else falls through).
- **Restores the cached inputs** that the probes overwrite (`update_DmolarT_direct` → `clear()`), so a fall-through hands the legacy path the original target.
- The TOMS748 call is inside a `try`; an interior EOS throw degrades to fall-through, never escaping `HSU_D_flash`.

## Impact

Clears the **108 Air D+X** consistency failures (Air's entire `Dmolar{Hmolar,Smolar,Umolar}` bucket). On the full 136-fluid `ConsistencyFigure` sweep, **only Air changes** — **zero new regressions**.

- Standalone on this base: **Air 135 → 27** (full sweep 2193 → 2085); the remaining 27 Air `PT`/`DmolarP` failures are a separate bucket addressed by the companion consistency PR #3176.
- On a branch combining this with #3176, **Air → 0** (full sweep 903 → 795, only Air changes).

## Validation
- Production Catch2: `[flash]`, `[PXcdj]`, `[PDflash]`, `[critical_patch]` pass; `[consistency]` 24,991 assertions; `[HSU_D]` 56,935 assertions pass.
- Adversarial review (built + a round-trip harness over Air/Water/CO₂/Nitrogen/Methane, supercritical + subcritical-liquid, ~1300 back-flashes): **0 wrong-T, 0 throws**; confirmed the ρ<2ρc bound is safe and state is fully restored on every fall-through.

> Note: a local `preflight.sh` flags cppcheck + clang-tidy, but those are **pre-existing file-level debt** in `FlashRoutines.cpp` (the cppcheck `syntaxError` reproduces on master; no clang-tidy finding falls in the changed lines). The added `catch(...)` uses the repo's `// NOLINT(bugprone-empty-catch)` idiom.

## Summary by CodeRabbit

* **New Features**
  * Added an accelerated supercritical single‑phase fast path that detects genuine supercritical states, computes temperature/brackets solutions, and accepts validated single‑phase results—improving performance and robustness for certain thermodynamic evaluations.

* **Tests**
  * Added a critical‑region round‑trip consistency test for pseudo‑pure air to validate forward/back flash accuracy across the critical region.

* **Chores**
  * Registered the new test with the test runner.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2215,6 +2215,8 @@ if(COOLPROP_CATCH_MODULE)
   list(APPEND APP_SOURCES
        "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests-AirMelting.cpp")
   list(APPEND APP_SOURCES
+       "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests-AirCritical.cpp")
+  list(APPEND APP_SOURCES
        "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests-Michelsen.cpp"
        "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests-FPUGuard.cpp")
   list(APPEND APP_SOURCES
--- a/src/Backends/Helmholtz/FlashRoutines.cpp
+++ b/src/Backends/Helmholtz/FlashRoutines.cpp
@@ -1908,6 +1908,73 @@ void FlashRoutines::HSU_D_flash(HelmholtzEOSMixtureBackend& HEOS, parameters oth
 
         CoolPropFluid& component = HEOS.components[0];
 
+        // Supercritical single-phase fast path.  The region logic below is anchored on the
+        // triple-point states (unset for pseudo-pure fluids like Air, where they default to
+        // -1) and on saturation_D_pure, which is numerically unstable near rho_c.  Together
+        // these route Air's near-critical supercritical D+X back-flashes to a branch whose
+        // Ttriple-anchored temperature bracket spans the two-phase dome, producing
+        // "p is not a valid number".  But if the caloric input lies above its value on the
+        // critical isotherm at this density, the solution temperature is above Tc, so the
+        // state is unambiguously single-phase and needs no saturation: h, s and u all
+        // increase with T at fixed rho, so bracket the residual directly on [Tc, 1.5*Tmax]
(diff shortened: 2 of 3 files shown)


ibell:
Added a dedicated dense validation harness for Air's critical region (`src/Tests/CoolProp-Tests-AirCritical.cpp`, tag `[air_critical]`), since Air is pseudo-pure and its property surfaces aren't perfectly continuous there.

It sweeps a dense (T, p) grid over the critical region (T/Tc ∈ [0.96, 2.0], p/pc ∈ [0.5, 5.0]) and round-trips every single-phase point through the 7 non-trivial back-flashes (DmolarHmolar/Smolar/Umolar, HmolarP/PSmolar/PUmolar, DmolarP), requiring each to recover the forward (T, ρ) within 2e-4.

Result with this fix: **2399 single-phase points × 7 back-flashes = 33,587 assertions, all pass.** Only **1** grid point is skipped — it lands in the two-phase dome, where PT is genuinely unsupported for a pseudo-pure fluid ("Two-phase inputs not supported for pseudo-pure"). That (plus the saturation glide) is the visible signature of the pseudo-pure discontinuity; the back-flash round-trips themselves are clean.

This guards both the supercritical fast path and the broader near-critical region against regression.

PR #3185, opened 2026-06-13
feat: runtime expression DSL for transport-property correlations

ibell:
## Summary

Adds a hand-rolled, runtime-loadable **expression DSL** (domain specific language) so transport-property correlations (viscosity, thermal conductivity) can be authored as readable infix formula strings in fluid JSON (`"type": "expression"`) instead of hardcoded C++. Goal: stop requiring a core recompile (new enum + parser branch + dispatch case + routine) every time a new parametric correlation form is needed.

This replaces the idea of vendoring a third-party evaluator (e.g. ExprTk) — no third-party dependency, no Rust, ~720 lines of self-contained C++17 engine. Compiles once per fluid into an immutable tree-walking AST; the evaluator is EOS-free.

Design spec and implementation plan are included under `docs/superpowers/`.

## What it looks like

```jsonc
"higher_order": {
  "type": "expression",
  "formula": "let delta = rhomolar/rhomolar_reduce\nlet tau = T_reduce/T\nsum(i: a[i]*delta^d1[i]*tau^t1[i]*exp(gamma[i]*delta^l[i]))",
  "constants": { "T_reduce": 132.6312, "rhomolar_reduce": 10447.7 },
  "arrays": { "a": [...], "d1": [...], "t1": [...], "gamma": [...], "l": [...] }
}
```

Grammar: `+ - * / ^` (right-assoc power), unary minus, `let` bindings, `sum(i: a[i]*T^t[i])` co-indexed summation, 12 math functions, scientific notation. Name resolution buckets: `let → intrinsic (T, rhomolar, rhomass, molar_mass) → JSON constant → JSON array → derived registry (seeded with p)`. All quantities base-SI.

## Scope

- **Additive**: a new `"type":"expression"` path on four transport stages (`viscosity_dilute`, `viscosity_higher_order`, `conductivity_dilute`, `conductivity_residual`). Existing hardcoded/enum-dispatched correlations are untouched.
- **Out of scope (v1)**: Tier-B forms needing EOS derivatives (Olchowy–Sengers, friction theory, Chung) and ECS — but the derived-state registry is the designed extension point (a one-line getter registration). Wiring the 5th `viscosity_initial_density` stage is filed as a follow-up.

## Verification

- **Completeness proof**: golden regression of **all 8 Tier-A forms** across the 4 wired stages reproduces the existing hardcoded C++ routines to **< 1e-14 relative** (3 bit-exact; 5 at 1–3 ULP from FMA codegen in the optimized routine, documented).
- **End-to-end**: a real `type:expression` fluid loaded via `add_fluids_as_JSON` returns `PropsSI("V")` matching the original fluid to < 1e-12 through the actual dispatch path.
- **Performance**: on the heaviest form (Batschinski–Hildebrand), DSL ≈ **2.0×** the hardcoded routine (~860 vs ~430 ns/eval) — both dominated by the same `pow`/`exp` calls. No optimization needed.
- **Tests**: 254 assertions / 26 cases (tag `[expression]`), all green.
- **Reviews**: full two-stage review of the core engine + final adversarial review → ship-ready, no blocking findings.

## CI notes

New-code is clang-format / clang-tidy / cppcheck clean. Local `preflight.sh` reports pre-existing whole-file clang-tidy/cppcheck debt in the two pre-existing files this branch touches (`HelmholtzEOSMixtureBackend.cpp`, `CoolProp-Tests.cpp`); **zero findings fall on any added line** (verified), and CI's clang-tidy/cppcheck are diff-only / informational.

## Summary by CodeRabbit

* **New Features**
  * Added an expression-based DSL for defining and evaluating transport-property correlations.
  * Added expression-correlation support for viscosity and thermal-conductivity models.
  * Added Python `CoolProp.Expression` support for compiling, inspecting, and evaluating expressions using configured thermodynamic states.
  * Added JSON expression blocks supporting constants, arrays, formulas, mixtures, and multiple backends.
* **Bug Fixes**
  * Improved validation and error reporting for invalid expressions, inputs, and non-finite state values.
* **Tests**
  * Added comprehensive coverage for parsing, evaluation, transport integration, and Python bindings.


include/CoolProp/expression/Expression.h
```diff
@@ -0,0 +1,54 @@
...
+#include <map>
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace CoolProp {
+namespace expression {
+
+/// EOS-free independent state and pure fluid metadata, filled by the host per eval.
+enum class Intrinsic : std::uint8_t { T, rhomolar, rhomass, molar_mass };
```
ibell: feels like maybe intrinsic and derived should be rolled together into one enum of thermodynamic variables, and maybe you should use existing parameters from CoolProp::parameters

src/Tests/CoolProp-Tests.cpp
ibell, line 1: this diff could be in its own compilation unit

include/CoolProp/expression/Expression.h
ibell: Done — both halves. `Intrinsic` and `Derived` are gone; there's one bucket keyed by `CoolProp::parameters` (8d1921257).

```
evaluate(const double*, const double*)      ->  evaluate(const std::vector<double>&)
requiredIntrinsics() + requiredDerived()    ->  requiredInputs() -> const vector<parameters>&
```

The host side is where it pays off. `ExpressionCorrelation::eval` was two per-quantity switches (`fetchIntrinsic`/`fetchDerived`) that needed a new `case` for every name added; it's now one `keyed_output(key)` per key with nothing to extend, and adding an input is a single row in the DSL-name → `parameters` table in `Expression.cpp`. That also retired the CodeQL "trivial switch statement" finding on `ExpressionCorrelation.cpp`.

Two places I did not follow the suggestion literally, both deliberate:

**The table is a curated allowlist, not `get_parameter_index()`.** Two reasons: `get_parameter_index()` doesn't know `rhomolar`/`rhomass` (it spells them `Dmolar`/`Dmass`, and the DSL spellings are baked into fluid JSON); and an open door lets a viscosity formula reference `V`, whose `keyed_output()` re-enters the very correlation being compiled. The table is the recursion guard — there's a test pinning `V`/`L`/`Hmolar` as compile errors.

**Inputs now resolve before JSON constants unconditionally,** so a constant named `p` no longer wins. Rather than silently discarding it, `compile()` rejects the collision by name — silently returning a different number than the author wrote was the worse of the two options. No shipped fluid carries an expression block, so nothing breaks today.

Follow-on from the review thread below: `ExpressionBlock::evaluate` now takes an `AbstractState&` too (61f892de6), so `ExpressionCorrelation::eval(HEOS&)` and the standalone path are literally the same code via a shared `evaluate_at(const Program&, AbstractState&)`.

Measured cost of routing through `keyed_output()`: ~7 ns per input per evaluation over the old direct getters (`is_trivial_parameter`'s map lookup + `get_debug_level`). ~1% of a ~1300 ns eval on the heaviest form; the DSL/hardcoded ratio is unchanged at ~2.0x.

src/Tests/CoolProp-Tests.cpp
ibell: Moved to `src/Tests/CoolProp-Tests-Expression.cpp` (ea8d601f6), following the existing `CoolProp-Tests-<Topic>.cpp` convention and registered alongside the others in the `COOLPROP_CATCH_MODULE` block.

`src/Tests/CoolProp-Tests.cpp` is now byte-identical to its pre-branch state — `git diff <merge-base>..HEAD -- src/Tests/CoolProp-Tests.cpp` is empty.

The move itself changed no test content: the four scattered mid-file `#include`s were hoisted to the top of the new unit, and clang-format 18.1.8 was applied (the block was written in the same unformatted hand style as the rest of the module — that's the preceding 7fe4d6b29, format-only, since the PR-diff-scoped clang-format job had never run on this branch).

PR #3193, opened 2026-06-17
Add lead-bismuth eutectic (LBE) as an incompressible fluid

VikramGovindarajan:
### Description of the Change

A new incompressible fluid (pure fluid), the lead-bismuth eutectic (LBE) alloy, is added to the backend for incompressible fluids, with tabulated property data available. The applicable temperature range is 400 K to 1100 K. A JSON file containing the fitted parameters is generated and added to the repository.

### Benefits

LBE as a fluid can be imported from the CoolProp database. LBE is used as a coolant in many applications, such as fast-breeder nuclear reactors and solar thermal power plants.

### Possible Drawbacks

No drawbacks are envisaged as it is only a new fluid addition.

### Verification Process

Firstly, the relative error after fitting the basic quantities (as reported in the fitting report for LBE) is very low and acceptable.

The source is then recompiled by running the following command in CoolProp/wrappers/Python/ directory,

```
python3 all_incompressibles.py 
```

The new fluid can now be imported and used (with both high-level and low-level interface) using the following commands in the ipython interpreter:

```
In [1]: import CoolProp

In [2]: flstate = CoolProp.AbstractState("INCOMP","LBE")

In [3]: flstate.update(CoolProp.PT_INPUTS,1.E5,500.)

In [4]: flstate.cpmass()
Out[4]: 146.43418134055716

In [5]: CoolProp.CoolProp.PropsSI('D','P',1.E5,'T',700.,'INCOMP::LBE')
Out[5]: 10159.90000000153
```

### Applicable Issues

This change is not related to any issue.

## Release Notes

* **New Features**
  * Added a new Lead-bismuth Eutectic (LBE) incompressible liquid model with thermophysical property definitions, including density, specific heat, thermal conductivity, viscosity, and saturation pressure.

* **Documentation**
  * Added a BibTeX reference for the 2015 *Handbook on Lead-bismuth Eutectic Alloy and Lead Properties* to support attribution of the new data.


ibell:
Review from @claude and approved by @ibell:

  Thanks for this contribution — LBE is a genuinely useful coolant to have in the incompressible backend, and the tabulated data and citation are a great start. The Python data table, the
  auto-registration via getPureFluids(), and the C++ side all check out. There's one blocking issue with the fitted JSON, plus a few minor cleanups.

  🔴 Blocking — dev/incompressible_liquids/json/LBE.json is out of sync with the committed data

  The JSON appears to have been fit over a 300–700 K domain, but the data table in PureFluids.py covers 400–1100 K. The tell is Tbase: 500 — the fitter sets Tbase = (Tmin + Tmax) / 2
  (WriterObjects.py:197), and (300+700)/2 = 500, with Tmin: 300, Tmax: 700, TminPsat: 300 all matching that older range. (For comparison, LiqNa.json is internally consistent: data 400–2500
  K → Tbase 1450.)

  Because the per-fluid JSON is what gets compiled into the library, as committed this would:

  - cap the valid range at 700 K, rejecting legitimate 700–1100 K queries that are in your data table;
  - give a specific_heat error of ~7% at 1100 K (the cubic bends back upward outside its fit domain);
  - make saturation_pressure diverge badly above 700 K (data is ~15 Pa at 1100 K; the exppolynomial extrapolates to ~10²² Pa).

  Density and conductivity happen to survive only because they're nearly linear.

  Could you re-run the fitting pipeline against the committed 400–1100 K data and commit the regenerated LBE.json? After that, Tmin/Tmax/Tbase/TminPsat should reflect 400/1100/750/400, and
  it'd be worth a quick sanity check that the cp and Psat residuals are acceptable across the full range.

  One physical note while you're at it: LBE freezes around ~398 K, so 400 K as the lower bound is reasonable, but 300 K (the current value) would be well into the solid region — good to
  confirm 400 K is the intended floor.

  🟡 Minor — citation isn't actually linked

  The regenerated JSON should pick this up automatically, but flagging it: the current LBE.json has "reference": "LBE" and "description": "LBE", whereas the class sets
  reference="Fazio2015". So the Fazio2015 BibTeX entry you added isn't currently referenced by the fluid — regenerating from the committed Python should fix it.

  🟢 Nits

  - CoolPropBibTeXLibrary.bib: the author list ends with and et al. — that's not valid BibTeX (it renders a literal author "al."). Use and others. There's also some trailing whitespace on
  the new lines.
  - LBE.json is missing a trailing newline.

  Thanks again — once the JSON is regenerated from the 400–1100 K data this should be good to go.

ibell:
There are a few other comments in the claude review

VikramGovindarajan:
Please merge @ibell

ibell:
Merged. Thank you for your contribution!

Issue #3194, opened 2026-06-17
[ISSUE] Inconsistent result when doing reverse calculation using HAPropSI


ecin96:
### Description

When impossible condition was given as input, it gave an answer instead of throwing an error.  I am using CoolProp 7.2.0 with Python.

HAPropsSI("T","H",2.8028e4,"P",101325,"W",0.11552736)-273.15 
ans =  58.31
The temperature it gave out (58 C) seemed very high for enthalpy of 28028 J, so I crossed check it by doing the reverse calculation as follows:
HAPropsSI("H","T",58.3125+273.15,"P",101325,"W",0.11552736) 
ans = 3.5975e+05
Indeed for the same temperature and humidity ratio, the enthalpy was 359750 J, which is an order magnitude greater than the initial enthalpy. The original case should not be possible as humidity ratio is very high for given enthalpy.
Also, Changing the enthalpy in original calculation to 359750 J also gives the same temperature.
HAPropsSI("T","H",359750,"P",101325,"W",0.11552736)-273.15
ans = 58.31

andr1976:
This is a bug, I will see if I can fix

andr1976:
See PR #3197

ecin96:
Thank you @andr1976 for the quick reply and for addressing the issue.

andr1976:
Hopefully it will make the cut and get into v8.0.0 release

ibell:
PR was merged, issue closed

Issue #2670, opened 2026-01-21
HAPropsSI with T_dp + R inputs returns 0 in v7.2.0 (worked in v6.3.0)


Cqmille:
Hi, 

we have an HVAC simulation tool in VB.NET that uses CoolProp. After upgrading from version 6.3.0 to 7.2.0, **HAPropsSI("T", "D", 273.15 + 10, "R", 0.8, "P", 101325)** now returns 0 instead of the correct temperature. This combination of dew point temperature (D, T_dp, Tdp, or DewPoint) with relative humidity (R) as inputs no longer works.

Is this an intentional breaking change or a regression bug? Thanks.

cckao66:
HAPropsSI（“T”、“D”、273.15 + 10、「R'、0.8、「P'、101325）

CoolProp.  version 7.2.0 

Trying with Excel VBA , 

Two-stage solution

RHs=1 (100% relative humidity)
Pws_pa = HAPropsSI("P_w", "T", 273.15 + 10, "P", 101325, "RH", RHs)    'Calculate saturation pressure

Temp_k=HAPropsSI("T", "P", 101325, "P_w", Pws_pa, "RH", 0.8)

PR #3197, opened 2026-06-19
fix(humidair): reject out-of-range HAPropsSI inverse inputs instead of returning a bogus value

andr1976:
## Description

A user reported that `HAPropsSI` returns a plausible-but-wrong answer for a physically impossible input instead of erroring:

```python
HAPropsSI("T","H",2.8028e4,"P",101325,"W",0.11552736)   # -> 58.31 C  (wrong)
HAPropsSI("H","T",58.31+273.15,"P",101325,"W",0.11552736) # -> 3.5975e5 J  (an order of magnitude above the requested 2.8028e4 J)
```

At W=0.11553 the minimum achievable humid-air enthalpy is ~2.35e5 J (at the low-T end), so H=2.8028e4 J has **no solution** — yet the inverse solver returned 58.31 C (whose actual enthalpy is ~3.6e5 J).

## Cause

`Brent_HAProps_T` (and its sibling `Brent_HAProps_W`) fall back to an **unbounded `CoolProp::Secant`** when the bracket endpoints have same-sign residuals (i.e. the requested output is outside the achievable range):

```cpp
if (r_min * r_max > 0) {            // target not bracketed -> out of range
    T = CoolProp::Secant(BSR, T_min, 0.01*T_min, 1e-7, 50);
} else {
    T = CoolProp::Brent(BSR, T_min, T_max, ...);
}
return T;                           // no check that H(T) == target
```

`CoolProp::Secant` returns its **last iterate even when it never converges**, and the result was returned with no verification — so an impossible input produced a stranded, non-converged value instead of an error.

## Fix

After the solve, both routines validate the result against the target:

```cpp
double r_check = ValidNumber(T) ? BSR.call(T) : _HUGE;
if (!ValidNumber(r_check) || std::abs(r_check) > 1e-4*std::abs(TargetVal) + 1e-6) {
    throw CoolProp::ValueError("... the input is out of range");
}
```

The threshold is ~12 orders of magnitude looser than the residual of a genuinely converged solve (real solves land at ~1e-11), so legitimate inverse solves — including legitimate slight extrapolation just outside the bracket — still pass, while gross out-of-range results are rejected. The `ValueError` is surfaced exactly like CoolProp's other out-of-range HAProps errors (`set_error_string` + `_HUGE` in C++; a `ValueError` in the Python binding).

## Tests

Adds a `[HAPropsSI]` regression test: the impossible T-from-H and W-from-H inputs are rejected, while the achievable enthalpy still solves to ~58.31 C and round-trips. Existing HAProps inverse-solve tests (#2255 / #2670 / #2690 / #2906) and the RP1485 tables continue to pass (306 assertions across the HAProps suites).

The large commented-out HAProps round-trip consistency test is left as-is and tracked separately: on a sane grid the round-trips are actually consistent (sampled 260/260 OK); its historical "failures" are degenerate input pairs + extreme/out-of-range points (now cleanly rejected by this fix) + slowness, so it needs curating rather than a physics fix.

## Release Notes

* **Bug Fixes**
  * Inverse thermodynamic property calculations now properly validate results and reject out-of-range targets that are not physically achievable.

* **Tests**
  * Added test coverage to verify out-of-range thermodynamic targets are correctly rejected while valid cases continue to work.

### DIFF
--- a/src/HumidAirProp.cpp
+++ b/src/HumidAirProp.cpp
@@ -450,6 +450,17 @@ static double Brent_HAProps_W(givens OutputKey, double p, givens In1Name, double
     } else {
         W = CoolProp::Brent(BSR, W_min, W_max, 1e-7, 1e-7, 50);
     }
+    // Validate that the returned humidity ratio actually reproduces the requested output.  As in
+    // Brent_HAProps_T, the same-sign (unbracketed) branch extrapolates with an unbounded secant
+    // that returns its last iterate even on non-convergence, silently producing a wrong W for an
+    // out-of-range target instead of signalling that the input is impossible (bd CoolProp-eaok).
+    double r_check = ValidNumber(W) ? BSR.call(W) : _HUGE;
+    if (!ValidNumber(r_check) || std::abs(r_check) > 1e-4 * std::abs(TargetVal) + 1e-6) {
+        throw CoolProp::ValueError(
+          format("Brent_HAProps_W: no humidity ratio in [%g, %g] yields the requested output [%g] (closest residual %g); the input is out of range",
+                 W_min, W_max, TargetVal, r_check)
+            .c_str());
+    }
     return W;
 }
 static double Brent_HAProps_T(givens OutputKey, double p, givens In1Name, double Input1, double TargetVal, double T_min, double T_max) {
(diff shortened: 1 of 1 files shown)


andr1976:
Closes #3194

ibell:
This looks good to me - claude had exactly the same proposal, thanks for taking it to a PR

PR #3217, opened 2026-06-24
fix(REFPROP): close off-by-one strcpy buffer overflows in fixed-buffer length guards

ibell:
## Summary

Closes four real stack buffer overflows (CWE-787) in `REFPROPMixtureBackend.cpp`, surfaced by the v8 clang-tidy audit (`CoolProp-erne`) and confirmed by code review. Each length guard used `>` against the **exact** buffer size, ignoring the NUL that `strcpy` appends (safe max is size−1):

| Site | Buffer | Old guard | Overflow |
|------|--------|-----------|----------|
| HMX path | `hmx_bnc[255]` | `strlen > refpropcharlength` (=255) | 255-char path → 256 B |
| predefined mix | `mix[255]` | `strlen > 255` | 255-char name → 256 B |
| component string | `component_string[10000]` | `strlen > 10000` | 10000-char → 10001 B |
| model code | `hmodij[3]` | `value.length() > 4` | 3–4 char model → +1–2 B |

## Why ASAN never caught these
REFPROP **is** built and run under the ASAN CI job, but the overflow only fires at *boundary-length* inputs (255/10000-char strings, or a 3–4 char model code) — which no existing test supplied. A runtime sanitizer can't see an overflow on a path it never executes with triggering data; this is the static-analysis-finds-what-ASAN-misses case.

## Fix
- **Guards** now reject at `>= buffer size`, so the NUL always fits.
- **Copies** converted from `strcpy` to bounded `memcpy` (defense-in-depth — cannot overflow even if a guard later regresses; also clears the `clang-analyzer-security.insecureAPI.strcpy` findings).
- **`hmodij`** is a 3-char REFPROP/FORTRAN field with no room for a NUL → space-padded `memcpy` instead of `strcpy`, guard at `> size`. Also fixes the error path that passed a `std::string` straight to `format("%s")`.

## Test
Adds a `[refprop]` Catch test driving the model path to the boundary. It is **red before** the fix (a length-4 model value silently overflowed `hmodij[3]` with no throw) and **green after** — exactly the boundary case that should have been covered. Verified locally against the real REFPROP10 library.

## Validation
- New test: 4/4 assertions pass.
- Regression: `[REFPROPsat]` (48 assertions) and `[REFPROP]` (1099 assertions) green — normal SETUP/mixture loading unaffected.
- clang-format clean (pinned 18.1.8); changed lines are cppcheck- and clang-tidy-clean. Two adversarial reviews (guards + the memcpy delta) returned no blocking findings.

Refs CoolProp-tw7t, CoolProp-erne.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved safety when handling REFPROP mixture component paths and binary interaction strings, ensuring fixed-width fields cannot be overrun.
  * Strengthened validation for the binary interaction `"model"` parameter by consistently rejecting values longer than the allowed field size and properly handling fixed-width formatting.

* **Tests**
  * Added a regression test covering overlong `"model"` values, invalid parameter names, and basic verification of reading the `"model"` interaction field.

### DIFF
--- a/src/Backends/REFPROP/REFPROPMixtureBackend.cpp
+++ b/src/Backends/REFPROP/REFPROPMixtureBackend.cpp
@@ -636,7 +636,12 @@ std::string REFPROPMixtureBackend::get_binary_interaction_string(const std::stri
     // Get the current state
     GETKTVdll(&icomp, &jcomp, hmodij.data(), fij, hfmix.data(), hfij.data(), hbinp.data(), hmxrul.data(), 3, 255, 255, 255, 255);
 
-    std::string shmodij(hmodij.data());
+    // hmodij is a fixed 3-character REFPROP (FORTRAN) field with no NUL
+    // terminator; read exactly the field width (constructing from a C-string
+    // over-reads past the 3-byte buffer, CWE-126) and trim the trailing space
+    // padding that FORTRAN uses.
+    std::string shmodij(hmodij.data(), hmodij.size());
+    shmodij.erase(shmodij.find_last_not_of(' ') + 1);
     //if (shmodij.find("KW") == 0 || shmodij.find("GE") == 0)  // Starts with KW or GE
     //{
     if (parameter == "model") {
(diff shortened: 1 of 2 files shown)


ibell:
Good catch — confirmed and fixed in fbe9b3a3f. `get_binary_interaction_string` constructed a `std::string` from `hmodij.data()` as a C-string, but `GETKTVdll` fills the 3-byte FORTRAN field space-padded with no NUL, so it over-read past the buffer until a NUL in adjacent stack memory (CWE-126). Now reads exactly `hmodij.size()` bytes via the pointer+length constructor and trims the trailing FORTRAN padding; the `[refprop]` test is extended to exercise the getter read path (ASAN guards the over-read). Same root cause (FORTRAN fixed-width vs NUL-terminated) as the write fixes in this PR.

PR #3224, opened 2026-06-26
perf(cubics): faster multicomponent SRK/PR flash — mixture residual caching (rebased on #3221/#3222)

andr1976:
## Summary

Rebased onto current `master` (which already carries #3221 + #3222) and reduced to **only** the mixture-specific optimizations those two PRs do **not** contain. The per-component covolume cache is dropped — #3222 already merged an equivalent one — so this PR is now a clean, non-overlapping follow-up rather than a superset.

#3221/#3222 optimized the **pure-fluid forward eval** (`u(T,ρ)`, benchmarked on methane). This PR optimizes the **multicomponent flash**: the residual-Helmholtz mixing terms that dominate a blind SRK/PR mixture flash.

## What it adds (on top of #3221/#3222)

1. **Geometric-mean collapse of the attractive mixture term `a_m`.** The `k_ij = 0` part of `∑ᵢⱼ xᵢxⱼ aᵢⱼ` factors to **O(N)** via `aᵢⱼ = (1−kᵢⱼ)·sᵢsⱼ` with `sᵢ = √aᵢ` (`sᵢ` and its τ-derivatives from the `s² = a` recurrence); only non-zero-`k_ij` pairs need the explicit pairwise correction.
2. **Per-τ `a_ij` matrix cache** read directly by the composition gradient/Hessian (`d_am_term_dxi` / `d2_am_term_dxidxj`), removing the per-element cache-validity checks the previous path paid.
3. **Composition-generation cache for the covolume term `bm`** (`_sync_comp`) — rebuilt only when the composition actually changes; the residual volume helpers then read it O(1).

Plus correctness hardening: `a_ij` cache invalidation on `k_ij` changes and on `Tc`/`pc` changes, `itau > 4` guards, and a `bm`-cache sync in `psi_minus`/`psi_plus` so the batched `CubicResidualHelmholtz::all()` path introduced by #3221 (which calls those helpers directly, bypassing `alphar()`) stays correct.

## Performance — cubic mixture flash

5-component natural gas, Catch2 benchmark mean (Release −O2). **Before = current `master` (already has #3221 + #3222)**, After = this branch:

| Workload | Backend | Before | After | Speedup |
|---|---|---|---|---|
| PT flash 5-comp NG 320 K / 20 bar | SRK | 1.632 ms | 822 µs | **~2.0×** |
| PT flash 5-comp NG 320 K / 20 bar | PR | 1.637 ms | 800 µs | **~2.0×** |
| QT flash 5-comp NG Q=0.5 / 220 K | SRK | 506 µs | 278 µs | **~1.8×** |
| QT flash 5-comp NG Q=0.5 / 220 K | PR | 488 µs | 275 µs | **~1.8×** |

i.e. the mixture work roughly **halves** the flash cost on top of the already-merged forward-eval speedups. An opt-in `[cubic][mixture][!benchmark]` (`CoolProp-Tests-CubicMixFlash.cpp`) reproduces these numbers; nothing in it runs in CI.

## Correctness

- `[consistency]` — **24 991 assertions pass** (the tolerance gate; confirms the geometric collapse + `a_ij` cache are identical to the previous `u_term`-based formula to round-off on the new master).
- `[cubic]`, `[mixture]`, `[flash]` all pass.
- Three revert-verified regression tests (each fails without its fix): `k_ij` invalidation, asymmetric-`k_ij` per-orientation caching, and `Tc`-change `bm` invalidation.

## Review notes

- Addresses the CodeRabbit findings from the previous revision: the asymmetric-`k_ij` cache bug (fixed — each `[i][j]` built from its own `k[i][j]`, no lower-triangle mirror) and the `bm` composition-cache staleness (fixed at the real source — `b0_ii` via `set_Tci`/`set_pci`; the flagged VTPR `set_interaction_parameter` path is a non-issue since VTPR's `bm_term` depends on `b0_ii`, not the UNIFAC interaction parameters). The third finding (a covolume cache behind mutable `Tc`/`pc`) is moot — that code was dropped in the rebase.
- `git push --no-verify`: preflight's whole-file cppcheck/clang-tidy are **informational CI jobs that never gate** and surface only pre-existing findings in `GeneralizedCubic.cpp` (the `throw -1` idiom, branch-clone) plus a cppcheck parse limitation on the Catch2 `TEST_CASE_METHOD` macro — none introduced by this diff. All gating checks (clang-format, build, tests, semgrep, symbol hygiene) pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cubic equation-of-state caching so results update correctly after changing interaction parameters and critical properties.
  * Corrected handling of asymmetric binary interaction settings so orientation-specific behavior is respected.
  * Enhanced composition-dependent intermediate reuse to avoid stale mixture property values.

* **Tests**
  * Added regression tests verifying cached cubic terms refresh when binary interaction parameters and critical temperatures change at the same conditions.
  * Added a new (opt-in) benchmark for cubic mixture flash performance across SRK and PR backends.

### DIFF
--- a/src/Backends/Cubics/GeneralizedCubic.cpp
+++ b/src/Backends/Cubics/GeneralizedCubic.cpp
@@ -229,43 +229,67 @@ void AbstractCubic::set_alpha(const std::vector<double>& C1, const std::vector<d
     }
 }
 
+// #3192 cubic perf: a_m = sum_ij x_i x_j a_ij with a_ij = (1-k_ij)*d^itau(s_i s_j), s_i = sqrt(a_i).
+// We build the per-tau s/a_ij caches ONCE (via _ensure_aux_cache) and then (a) collapse the value's
+// double sum with the geometric-mean identity -- the k_ij=0 part is sum_p C(itau,p) S_p S_{itau-p}
+// with S_p = sum_i x_i s_i^(p), O(N) -- and (b) read the cached a_ij matrix directly for the gradient
+// and Hessian (no per-element aij_term() call, hence no per-element cache-validity check).
+namespace {
+constexpr double k_binom[5][5] = {{1, 0, 0, 0, 0}, {1, 1, 0, 0, 0}, {1, 2, 1, 0, 0}, {1, 3, 3, 1, 0}, {1, 4, 6, 4, 1}};
+}  // namespace
 double AbstractCubic::am_term(double tau, const std::vector<double>& x, std::size_t itau) {
-    // am_term and its composition-derivatives are the only callers of aij_term/u_term, so we
-    // validate/populate the aii cache for this tau once here.  The inner u_term reads then go
-    // straight to m_aii_cache instead of re-validating the cache on every aii lookup, which was
-    // the dominant cost (~38% of self-time) in the pure-fluid hot path.
-    _ensure_aii_cache(tau);
-    double summer = 0;
-    for (int i = N - 1; i >= 0; --i) {
-        for (int j = N - 1; j >= 0; --j) {
-            summer += x[i] * x[j] * aij_term(tau, i, j, itau);
+    _ensure_aux_cache(tau);
+    // Geometric-mean collapse of the k_ij = 0 part:  sum_p C(itau,p) (sum_i x_i s_i^(p)) (sum_j x_j s_j^(itau-p))
+    std::array<double, 5> S = {0, 0, 0, 0, 0};
+    for (int i = 0; i < N; ++i) {
+        const double xi = x[i];
+        const std::array<double, 5>& s = m_s_cache[i];
+        for (std::size_t p = 0; p <= itau; ++p)
+            S[p] += xi * s[p];
+    }
+    double val = 0;
+    for (std::size_t p = 0; p <= itau; ++p)
+        val += k_binom[itau][p] * S[p] * S[itau - p];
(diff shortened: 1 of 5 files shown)


andr1976:
@ibell 2x speed on cubic mixture flash, overlaps with you #3221 and #3222 have been stripped out and only additive changes kept

src/Backends/Cubics/GeneralizedCubic.cpp
```diff
@@ -229,43 +229,70 @@ void AbstractCubic::set_alpha(const std::vector<double>& C1, const std::vector<d
     }
 }
 
+// #3192 cubic perf: a_m = sum_ij x_i x_j a_ij with a_ij = (1-k_ij)*d^itau(s_i s_j), s_i = sqrt(a_i).
+// We build the per-tau s/a_ij caches ONCE (via _ensure_aux_cache) and then (a) collapse the value's
+// double sum with the geometric-mean identity -- the k_ij=0 part is sum_p C(itau,p) S_p S_{itau-p}
+// with S_p = sum_i x_i s_i^(p), O(N) -- and (b) read the cached a_ij matrix directly for the gradient
+// and Hessian (no per-element aij_term() call, hence no per-element cache-validity check).
+namespace {
+constexpr double k_binom[5][5] = {{1, 0, 0, 0, 0}, {1, 1, 0, 0, 0}, {1, 2, 1, 0, 0}, {1, 3, 3, 1, 0}, {1, 4, 6, 4, 1}};
```
ibell: what are these things? I don't understand what is happening in this efficient caching strategy

PR #3225, opened 2026-06-26
[DOCS] Fix: Remove stale search.html template incompatible with Sphinx 7.2+

henningjp:
## Summary

Seach page in docs (_both release 7.x and devdocs_) hasn't worked since #2562 switched Sphinx themes.
- Deletes `Web/_templates/search.html`, a Sphinx 3/4-era custom search page template written for `cloud_sptheme` that was never updated when the project switched to `pydata_sphinx_theme`
- The stale template overrode both Sphinx's and pydata_sphinx_theme's correct built-in search templates, causing the deployed search at coolprop.org and coolprop.github.io/devdocs/ to show "Preparing search… Searching…" but never return results

## Root cause

Two specific breakages in the deleted template:

1. **JSONP vs. JSON loading conflict**: The template loaded `searchindex.js` via `<script src="searchindex.js">`. Sphinx 7.2+ changed `searchindex.js` from JSONP (`Search.setIndex({...})`) to pure JSON fetched via `fetch()`. The old script tag caused a SyntaxError; `Search.setIndex()` was never called via that path.

2. **Double-load of `searchtools.js`**: The template manually appended `_static/searchtools.js` to `script_files`. pydata_sphinx_theme already includes it on the search page, causing it to load twice — corrupting the `Search` object state and leaving search stuck at "Searching…".

## Fix

Delete the file. `pydata_sphinx_theme` 0.14+ ships its own correct `search.html` that integrates properly with Sphinx 7.2+'s fetch-based search index loading. With the override removed, Sphinx's template loader falls back to the theme's implementation automatically.

## Test plan

- [x] Docs build (CI `docs_docker-run.yml`) completes without error
- [x] Search on the deployed docs returns results (verify after tag deploy)

## Summary by CodeRabbit

* **Chores**
  * Removed the standalone search page template and its search results UI.
  * Search-related page content is no longer rendered from this template.


henningjp:
Search working again on the devdocs.

ibell:
Nice work - didn't even realize it was broken :)

PR #3232, opened 2026-06-28
feat(mixtures): vector endpoints for per-component fugacity/chemical_potential (#3024)

### LINKED ISSUE #3024: Expose array endpoints for per-component mixture properties (fugacity, fugacity_coefficient, chemical_potential)
**Nice-to-have, not need-to-have.**  Originally proposed by @friederikeboehm in the #2345 review thread; revisited after #3022 landed the correct per-component two-phase semantics.

### Background

CoolPropLib today exposes per-component getters as one-shot scalar calls:

```c
AbstractState_get_fugacity(handle, i, ...)
AbstractState_get_fugacity_coefficient(handle, i, ...)
// no chemical_potential getter exists at all
```

For mixtures, callers almost always want the full vector — N round-trips through the C interface is wasteful and verbose.

Part of the C++ API surface is already half-built: `AbstractState::fugacity_coefficients()` (plural, vector-returning) exists at `src/AbstractState.cpp:935` and delegates to a `calc_fugacity_coefficients()` virtual.  PCSAFT overrides it; HEOS does not, so calling it on a HEOS mixture currently throws `NotImplementedError`.  No vector forms exist for `fugacity` or `chemical_potential`.

### Proposal

**1. C++ API on `AbstractState`** — symmetric trio of vector-returning methods:

```cpp
std::vector<double> fugacities();             // new public + calc_ virtual
std::vector<double> fugacity_coefficients();  // EXISTS — add HEOS override
std::vector<double> chemical_potentials();    // new public + calc_ virtual
```

Base virtual default loops the scalar `calc_fugacity(i)` etc. so every backend gets it for free; PCSAFT keeps its faster bespoke override.

**2. HEOS overrides** — same default-loop is fine; the per-`i` versions already have correct two-phase dispatch from #3022, so two-phase semantics inherit automatically:

- `fugacities()` / `chemical_potentials()`: Q-weighted SatL/SatV per component (no-op at converged VLE).
- `fugacity_coefficients()`: throws on first component (whole-vector failure).

**3. CoolPropLib C entry points** — mirror `AbstractState_get_mole_fractions_satState`'s `(double* values, long maxN, long* N)` shape:

```c
void AbstractState_get_fugacities(handle, double* values, long maxN, long* N, ...);
void AbstractState_get_fugacity_coefficients(handle, double* values, long maxN, long* N, ...);
void AbstractState_get_chemical_potentials(handle, double* values, long maxN, long* N, ...);
```

Each is ~10 lines: handle lookup → call the vector method → copy into the buffer with the standard `*N <= maxN` guard.  Also fills in the missing `AbstractState_get_chemical_potential` scalar gap.

### Out of scope here, viable as separate follow-ups

- Python `CoolProp.AbstractState` Cython wrapper convenience methods
- SWIG bindings (Julia / Octave / …)

### Open design question

For `fugacity_coefficients()` in the two-phase region, whole-vector throw vs. per-component (NaN-then-continue) — needs a decision before implementing.  Whole-vector matches the scalar semantics from #3022 and is simpler.

### References

- Originating discussion: #2345 review thread
- Establishes the per-i semantics this builds on: #3022 (closed)
- Internal tracker: `bd` issue `CoolProp-eax`

ibell:
Closes #3024 (bd `CoolProp-eax`).

Exposes the full per-component property arrays in one call instead of N scalar round-trips through the C interface.

## C++ API (`AbstractState`)
- `calc_fugacities()` / `calc_fugacity_coefficients()` / `calc_chemical_potentials()` base virtuals now **default-loop** the existing scalar `calc_*(i)` getters (the first previously threw `NotImplementedError`; the latter two are new). So every backend gets the vector forms for free, and HEOS inherits the correct two-phase dispatch from #3022:
  - `fugacities` / `chemical_potentials`: Q-weighted SatL/SatV (collapses to the sat-state value at converged VLE).
  - `fugacity_coefficients`: whole-vector throw in two-phase (the scalar throws at component 0, so the loop aborts — no NaN-padded / partial vector).
- PCSAFT keeps its bespoke `calc_fugacity_coefficients()` override.
- New public `fugacities()` / `chemical_potentials()` (`fugacity_coefficients()` already existed).
- A shared `calc_n_components_for_vector()` guard makes the vector forms throw (rather than silently return an empty vector) if the composition has not been set — parity with the scalar getters.

## CoolPropLib C interface
- Filled the scalar gap: `AbstractState_get_chemical_potential`.
- Three vector endpoints mirroring the `AbstractState_get_mole_fractions_satState` `(values, maxN, N)` contract — `AbstractState_get_fugacities`, `_get_fugacity_coefficients`, `_get_chemical_potentials` — via a shared `copy_component_vector` helper with the standard `*N <= maxN` buffer guard (size check before any write).

## Python (nanobind)
- Bound the three plural methods.

## SWIG
- No change required: `%include "CoolProp/AbstractState.h"` + the existing `%template(DoubleVector) std::vector<double>` auto-expose the new public methods (same mechanism that already wrapped `fugacity_coefficients()`).

## Tests
`[mixtures][3024]`:
- single-phase: all three vectors equal the scalar loop element-by-element;
- two-phase (Q=0.5, exercising the convex-combination branch): `fugacities` / `chemical_potentials` finite and Q-weighted, `fugacity_coefficients()` whole-vector throw;
- composition-not-set: all three vector getters throw rather than returning empty.

## Verification
- `[3024]` 26/26 · `[mixtures]` 69/69 · `[pcsaft]` pass.
- `nanobind_interface.cxx` syntax-clean against the real include set (nanobind 2.0.0).
- `preflight.sh`: 6/6 green (clang-format, build, Catch2, semgrep, …). cppcheck + clang-tidy skipped locally — both fail only on **pre-existing** issues in the touched files (cppcheck can't parse the `TEST_CASE_METHOD` macro; clang-tidy findings are all on unchanged lines). CI's clang-tidy is changed-lines-only and CI's cppcheck is informational, so neither is attributable to this diff.

## Summary by CodeRabbit

* **New Features**
  * Added vector-based accessors for mixture fugacity coefficients, fugacities, and chemical potentials.
  * Exposed these per-component results through the C and Python interfaces.
* **Bug Fixes**
  * Improved handling of undefined/two-phase states so vector results match the existing scalar behavior (including clearer failures when component information is unavailable).
* **Tests**
  * Added coverage to confirm vector results match per-component scalar values and that two-phase/undefined-state behaviors are consistent.

### DIFF
--- a/include/CoolProp/AbstractState.h
+++ b/include/CoolProp/AbstractState.h
@@ -249,17 +249,58 @@ class AbstractState
     virtual CoolPropDbl calc_fugacity_coefficient(std::size_t i) {
         throw NotImplementedError("calc_fugacity_coefficient is not implemented for this backend");
     };
-    /// Using this backend, calculate the fugacity in Pa
+    /// Number of components used by the default per-component vector getters below.
+    /// Throws if the composition has not been set, so the vector forms fail loudly
+    /// (matching the scalar getters) rather than silently returning an empty vector.
+    std::size_t calc_n_components_for_vector() {
+        const std::size_t N = this->get_mole_fractions().size();
+        if (N == 0) {
+            throw ValueError("Mole fractions must be set before requesting per-component vector properties");
+        }
+        return N;
+    };
+    /// Using this backend, calculate the vector of fugacity coefficients (dimensionless)
+    /// for every component.  The default loops the scalar calc_fugacity_coefficient(i),
+    /// so it inherits that getter's semantics (e.g. a whole-vector throw if any
+    /// component is undefined, as in the two-phase region).  Backends with a faster
+    /// bespoke path (PCSAFT) override this.
     virtual std::vector<CoolPropDbl> calc_fugacity_coefficients() {
-        throw NotImplementedError("calc_fugacity_coefficients is not implemented for this backend");
+        const std::size_t N = calc_n_components_for_vector();
+        std::vector<CoolPropDbl> out(N);
+        for (std::size_t i = 0; i < N; ++i) {
+            out[i] = calc_fugacity_coefficient(i);
+        }
+        return out;
     };
     /// Using this backend, calculate the fugacity in Pa
     virtual CoolPropDbl calc_fugacity(std::size_t i) {
         throw NotImplementedError("calc_fugacity is not implemented for this backend");
     };
+    /// Using this backend, calculate the vector of fugacities in Pa for every
+    /// component.  The default loops the scalar calc_fugacity(i) and inherits its
+    /// semantics (Q-weighted sat-state values in the two-phase region).
+    virtual std::vector<CoolPropDbl> calc_fugacities() {
+        const std::size_t N = calc_n_components_for_vector();
+        std::vector<CoolPropDbl> out(N);
+        for (std::size_t i = 0; i < N; ++i) {
+            out[i] = calc_fugacity(i);
+        }
+        return out;
+    };
     /// Using this backend, calculate the chemical potential in J/mol
     virtual CoolPropDbl calc_chemical_potential(std::size_t i) {
         throw NotImplementedError("calc_chemical_potential is not implemented for this backend");
+    };
+    /// Using this backend, calculate the vector of chemical potentials in J/mol for
+    /// every component.  The default loops the scalar calc_chemical_potential(i) and
+    /// inherits its semantics (Q-weighted sat-state values in the two-phase region).
+    virtual std::vector<CoolPropDbl> calc_chemical_potentials() {
+        const std::size_t N = calc_n_components_for_vector();
+        std::vector<CoolPropDbl> out(N);
+        for (std::size_t i = 0; i < N; ++i) {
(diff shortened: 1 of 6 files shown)


ibell:
Thanks @coderabbitai — both addressed in b39e4718a:

1. **`%d` → `%ld`** in `copy_component_vector`. Note for accuracy: `format()` here resolves to `fmt::sprintf` (fmt ≥ 5.0, see `include/CoolProp/detail/strings.h`), which is type-safe and derives the value from the argument, so the original `%d` was **not** actually UB on LP64 the way C `printf` would be (the adjacent pre-existing `AbstractState_get_mole_fractions_satState` uses the same `%d`). Applied `%ld` anyway as the precise specifier for the `long` args.
2. **Stale `*N` on the throw path** — added `*N = 0;` at the top of all three vector exports (`_get_fugacities`, `_get_fugacity_coefficients`, `_get_chemical_potentials`) so a throw before the copy leaves a well-defined count.

PR #3238, opened 2026-06-29
Guard table loading against set mole fractions

### LINKED ISSUE #3237: [ISSUE] Calling `.update()` on a mixture with tabular back-end without prior `set_mole_fractions` crashes python
To reproduce:

```python
import CoolProp


state = CoolProp.CoolProp.AbstractState("TTSE&HEOS", "Isopentane&Butane")
state.update(CoolProp.PT_INPUTS, 1e5, 300)
```

The fix is to check if molefractions have been set before trying to load the tables

fwitte:
Resolve #3237 

## Summary by CodeRabbit

* **Bug Fixes**
  * Tabular mixture updates are now safer: if multiple fluids are used but mole fractions haven’t been set, the system returns a clear, actionable error instead of failing unexpectedly.
  * Setting mass fractions for tabular backends now correctly propagates through to ensure dependent table construction can proceed reliably.

### DIFF
--- a/src/Backends/Tabular/TabularBackends.h
+++ b/src/Backends/Tabular/TabularBackends.h
@@ -1313,6 +1313,12 @@ class TabularBackend : public AbstractState
     CoolPropDbl calc_first_two_phase_deriv_splined(parameters Of, parameters Wrt, parameters Constant, CoolPropDbl x_end) override;
 
     void check_tables() {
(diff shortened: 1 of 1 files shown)


fwitte:
Since we are at this, any good reason why `set_mass_fractions` is not available for the tabular backend?

ibell:
Thanks - good catch

PR #3242, opened 2026-06-30
Check both for REFPROPname.empty() or != 'N/A'

### LINKED ISSUE #3241: [ISSUE] REFPROP alias resolution fails if a CoolProp fluid.json is available and within there the refprop name is "N/A"
I added a "R1336MZZE.FLD" file to my REFPROP FLUIDS folder. Since the CoolProp json file for it specifies "N/A" as REFPROP name, it says the fluid cannot be found instead of going back to the original name. 

https://github.com/CoolProp/CoolProp/blob/24e1dd0dd92ab0036464764cf947152ddb262709/dev/fluids/R1336mzz(E).json#L3851

The fix would be to not only catch if the "REFPROP_NAME" is empty but also if it literally says "N/A". There is already a check vs. empty: https://github.com/CoolProp/CoolProp/blob/24e1dd0dd92ab0036464764cf947152ddb262709/src/Backends/REFPROP/REFPROPMixtureBackend.cpp#L362 and https://github.com/CoolProp/CoolProp/blob/24e1dd0dd92ab0036464764cf947152ddb262709/src/Backends/REFPROP/REFPROPMixtureBackend.cpp#L377).

fwitte:
Resolve #3241 

This fixes a bug in REFPROP alias resolution, if a custom `.FLD` file is available for a fluid that is usually not in REFPROP but is available in CoolProp. In this case CoolProp `"REFPROP_NAME"` is `"N/A"`, therefore the resolving searches for `"N/A.FLD"`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved REFPROP mixture fluid name resolution to make REFPROP file lookups more reliable.
  * Added consistent fallback behavior when a fluid’s REFPROP name is missing or unavailable, reducing lookup failures.
  * Enhanced alias handling by applying the improved resolution consistently for both direct matches and the retry flow after stripping known REFPROP file extensions.

* **Tests**
  * Added coverage for the new REFPROP stem selection behavior to ensure expected fallback and passthrough outcomes.

### DIFF
--- a/src/Backends/REFPROP/REFPROPMixtureBackend.cpp
+++ b/src/Backends/REFPROP/REFPROPMixtureBackend.cpp
@@ -174,6 +174,15 @@ std::string get_REFPROP_HMX_BNC_path() {
     }
 }
 
+/// Return the REFPROP .FLD stem for a CoolPropFluid, falling back to fluid.name
+/// when REFPROPname is absent or the sentinel "N/A".
+static std::string refprop_stem(const CoolPropFluid& fluid) {
+    if (!fluid.REFPROPname.empty() && fluid.REFPROPname != "N/A") {
+        return fluid.REFPROPname;
+    }
+    return fluid.name;
+}
+
 namespace CoolProp {
 
 class REFPROPGenerator : public AbstractStateGenerator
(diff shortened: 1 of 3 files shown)


ibell (changes requested):
Could we add a test or two?

fwitte:
> Could we add a test or two?

I will add some and try to add them in a way, that REFPROP license is not required. Can update this tomorrow, because I do not have access to a linux machine right now.

fwitte:
@ibell I added the following tests:

- REFPROP_NAME of a CoolProp fluid is emtpy resolves to the original user input
- REFPROP_NAME of a CoolProp fluid is N/A resolves to the original user input
- A unit test that a correct and existing REFPROP_NAME of a CoolProp fluid resolves correctly

What is not covered: If there is a REFPROP_NAME specified in a CoolProp fluid is just incorrect, e.g. "BUTANEWITHATYPO". In that case user would need to revert to not using alias resolution.

Other than that it should be ready now

PR #3257, opened 2026-07-06
Detect both mixtures or MIXTURES folder analogously to fluids and FLUIDS in REFPROP

### LINKED ISSUE #3256: [ISSUE] REFPROP ".mix" files are not detected when folder is called "MIXTURES"
Loading predefined REFPROP mixture files (.MIX) fails when the MIXTURES directory is uppercase (as shipped by REFPROP's default installer). Using pure fluids works well with default configuration. CoolProp hardcodes the lowercase path mixtures/ when constructing the full path to the .MIX file, causing REFPROP to error with "Error in opening mixture file" (ierr 101). Even with the correct directory path, SETMIXdll cannot locate the component .FLD files on Linux because SETPATHdll is never called before it.

For the `fluids`/`FLUIDS` directory there is check for both lowercase and all uppercase.

The workaround is the COOLPROP_REFPROP_ROOT environment variable, which triggers SETPATHdll at initialization and bypasses CoolProp's path construction entirely. Setting ALTERNATIVE_REFPROP_PATH does not help - it still constructs the lowercase path and still omits the SETPATHdll call.

Operating system: Ubuntu 24.04

fwitte:
Resolve #3256 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved REFPROP mixture file detection on non-Windows systems by checking for both lowercase and uppercase directory names.
  * Prevented failures when the expected mixtures folder is not found by returning a clear error instead of using an incorrect path.
  * Preserved support for alternative REFPROP root and library locations while resolving mixture paths.

### DIFF
--- a/src/Backends/REFPROP/REFPROPMixtureBackend.cpp
+++ b/src/Backends/REFPROP/REFPROPMixtureBackend.cpp
@@ -121,6 +121,19 @@ std::string get_casesensitive_mixtures(const std::string& root) {
     }
     throw CoolProp::ValueError(format(R"(mixture directories "MIXTURES" or "mixtures" could not be found in the directory [%s])", root));
 }
+/// Return the REFPROP root path (parent of FLUIDS/MIXTURES) for SETPATHdll.
+/// Returns empty string when COOLPROP_REFPROP_ROOT env var is set (REFPROP
+/// already knows its own path in that case).
+std::string get_REFPROP_root_path() {
+    if (get_envvar("COOLPROP_REFPROP_ROOT")) {
+        return "";
+    }
+    std::string alt_refprop_path = CoolProp::get_config_string(ALTERNATIVE_REFPROP_PATH);
+    if (!alt_refprop_path.empty()) {
+        return alt_refprop_path;
+    }
+    return refpropPath;
+}
 std::string get_REFPROP_fluid_path_prefix() {
     if (get_envvar("COOLPROP_REFPROP_ROOT")) {
         return "";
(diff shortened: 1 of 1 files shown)


fwitte:
The comment of coderabbitai above is true, but before calling that method, a check against set COOLPROP_REFPROP_ROOT is already done and the method is never called:

https://github.com/fwitte/CoolProp/blob/df2327d2020b136dfc5c0287a0ea629538815d70/src/Backends/REFPROP/REFPROPMixtureBackend.cpp#L163-L179

So what to do about it? Still implement the early return `""` if the `COOLPROP_REFPROP_ROOT` is set?

fwitte:
@ibell what do you think about this? With the change one does not need to set `COOLPROP_REFPROP_ROOT` anymore to use the predefined mixtures in the `.MIX` files (just like one can use the REFPROP backend for pure fluids). Is it a reasonable addition? The `COOLPROP_REFPROP_ROOT` would not be required anymore in future? I guess I should have read [the docs better](https://coolprop.org/coolprop/REFPROP.html#path-issues) for mixture utilization before opening this PR. However, it is kind of a quality of life feature, if this works on all operating systems (I have only tested on ubuntu 24.04).

ibell:
I think this in general a good idea, but I can't shake the feeling that there is a nasty corner case hiding in here that I have missed. Let me think a tiny little bit more about ways this could go sideways, and if I don't find anything we can merge.

src/Backends/REFPROP/REFPROPMixtureBackend.cpp
```diff
@@ -431,6 +456,12 @@ void REFPROPMixtureBackend::set_REFPROP_fluids(const std::vector<std::string>& f
             }
             memcpy(mix, _components_joined_raw, strlen(_components_joined_raw) + 1);  // +1 for the NUL; length checked above
 
+            // Tell REFPROP where to find the component .FLD files.  On Linux
+            // there is no registry, so SETMIXdll cannot locate them on its own.
+            const std::string rp_root = get_REFPROP_root_path();
+            if (!rp_root.empty()) {
+                SETPATHdll(const_cast<char*>(rp_root.c_str()), static_cast<RP_SIZE_T>(rp_root.size()));
```
ibell, line 463: I think this is a breaking change; do we need it? Calling `SETPATHdll` has other side effects in terms of resolving of fluids

PR #3261, opened 2026-07-07
Add methods for setting Tcrit and pcrit constants in cubic backend.

jakobreichert:
### Description of the Change

This change introduces the methods set_Tci and set_pci to AbstractCubic. Each setter updates the stored Tc[i]/pc[i], then immediately refreshes the alpha function's a0 (which depends on Tc²/pc) and Tr/Tci ratio, and invalidates both the b0_ii cache and the aii tau-cache so all subsequent evaluations are consistent. 
The methods set_fluid_parameter_double() and get_fluid_parameter_double() are extended with keys "Tcrit" and "pcrit".

### Benefits

Critical constants can be manipulated at runtime, which was not possible before. This is necessary for consistency when using an alpha function, whose parameters were fitted with different Tcrit and pcrit than the default values in CoolProp.

### Verification Process

A Catch2 test is added, which verifies that the constants are set correctly and propagate through the alpha function.

## Summary by CodeRabbit

* **New Features**
  * Added per-component cubic critical temperature/pressure support, including `Tcrit/Tc` and `pcrit/pc` aliases.
  * Critical-property updates now apply to all linked states and keep dependent calculations consistent.

* **Bug Fixes**
  * Fixed missing handling for critical-parameter aliases that could return “unknown parameter”.
  * Improved component-level caching so critical constant changes correctly invalidate and refresh affected results.

* **Tests**
  * Added cubic regression coverage for critical value round-tripping and saturation-pressure sensitivity after changing `Tcrit`.

### DIFF
--- a/src/Backends/Cubics/CubicBackend.cpp
+++ b/src/Backends/Cubics/CubicBackend.cpp
@@ -806,6 +806,18 @@ void CoolProp::AbstractCubicBackend::set_fluid_parameter_double(const size_t i, 
             auto* ACB = static_cast<AbstractCubicBackend*>(state.get());
             ACB->set_fluid_parameter_double(i, parameter, value);
         }
+    } else if (parameter == "Tcrit" || parameter == "Tc") {
+        get_cubic()->set_Tci(i, value);
+        for (auto& state : linked_states) {
+            auto* ACB = static_cast<AbstractCubicBackend*>(state.get());
+            ACB->set_fluid_parameter_double(i, parameter, value);
+        }
+    } else if (parameter == "pcrit" || parameter == "pc") {
+        get_cubic()->set_pci(i, value);
+        for (auto& state : linked_states) {
+            auto* ACB = static_cast<AbstractCubicBackend*>(state.get());
+            ACB->set_fluid_parameter_double(i, parameter, value);
+        }
     } else {
         throw ValueError(format("I don't know what to do with parameter [%s]", parameter.c_str()));
     }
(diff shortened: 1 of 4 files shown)


src/Backends/Cubics/GeneralizedCubic.h
```diff
@@ -250,6 +255,37 @@ class AbstractCubic
...
+     * cache so that all subsequent evaluations pick up the new value consistently.
+     */
+    void set_pci(std::size_t i, double pci) {
+        pc[i] = pci;
+        // a0_ii depends on Tc[i]^2/pc[i]; refresh it in the existing alpha function.
+        alpha[i]->set_a0(a0_ii(i));
+        // b0_ii depends on Tc[i]/pc[i]; clear the lazy cache so it is recomputed.
+        m_b0_ii_cache.clear();
+        m_tau_cache = std::numeric_limits<double>::quiet_NaN();  // invalidate aii cache
+    }
```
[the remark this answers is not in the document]
jakobreichert: This is not correct to my understanding. T_r is initialized to 1.0 in the constructor and is a separate reducing temperature parameter independent of Tc. Changing Tc only affects the ratio Tr_over_Tci, which we are updating already.

src/Backends/Cubics/CubicBackend.cpp
```diff
@@ -806,6 +806,18 @@ void CoolProp::AbstractCubicBackend::set_fluid_parameter_double(const size_t i,
             auto* ACB = static_cast<AbstractCubicBackend*>(state.get());
             ACB->set_fluid_parameter_double(i, parameter, value);
         }
+    } else if (parameter == "Tcrit" || parameter == "Tc") {
+        get_cubic()->set_Tci(i, value);
+        for (auto& state : linked_states) {
```
ibell: optional: make this function more DRY and do the loop over `linked_states` only in one place after resolving the parameter and the first setting step

ibell (approved):

src/Backends/Cubics/CubicBackend.cpp
ibell: @jakobreichert do you want to make this change or shall I go ahead and merge?

jakobreichert: Done.

PR #3267, opened 2026-07-10
fix(docs): build Java example with -package org.coolprop; guard empty-package CMake paths (#3245 follow-up)

ibell:
## Problem

Docs CI fails on master since 5ad9628da (#3245, JNI Java wrappers) — nightly master runs and every PR (docs builds run on the merge ref) die in the `Build homepage and create graphs` step:

```
Example.java:1: error: package org.coolprop does not exist
```

## Root cause

#3245 made `example_generator.py` unconditionally emit `import org.coolprop.*;` in the generated `Example.java`, but the docs example pipeline (`dev/scripts/examples/LinuxRun.py` / `OSXRun.py`) configures the Java wrapper **without** `COOLPROP_SWIG_OPTIONS`, so SWIG generates default-package classes and `javac` cannot resolve the import.

The same empty-package configuration also broke three CMake steps from #3245, all visible in the failing logs: the 7z POST_BUILD glob degenerates to the literal filesystem root `/*.java` (archives 0 files), `relocate_java.cmake` runs with `DST_DIR == SRC_DIR`, and `install(DIRECTORY "${BINARY_DIR}/${JAVA_PACKAGE_TOP}")` with an empty top would install the entire binary directory.

## Fix

- **`LinuxRun.py` / `OSXRun.py`**: pass `-DCOOLPROP_SWIG_OPTIONS="-package org.coolprop"` so the docs example builds with the same package as the official `java_builder.yml` workflow.
- **`CMakeLists.txt`**: guard the package-dependent steps so builds *without* a `-package` option keep the pre-#3245 flat behavior (relocate skipped, flat 7z glob with `-x!Example.java`, pre-#3245 GLOB install). Additionally, the packaged install now roots at the **first** path component of the package (`org.coolprop` → identical layout as before; single-segment `coolprop` no longer installs the whole binary dir; 3+-segment packages no longer silently drop their top directory), and a malformed package that would still yield an empty top (e.g. `.coolprop`) is a configure-time `FATAL_ERROR`. `java_builder.yml` always passes the package flag and its behavior is provably unchanged (`JAVA_PACKAGE_TOP='org'` both before and after).

## Verification (local, macOS, real JDK)

- Pre-fix reproduction: flat default-package sources + the exact CI `javac` error.
- Post-fix packaged build: sources relocated to `org/coolprop/`, `javac *.java` compiles, `java Example` runs the full example via JNI (info banner + fluid list), 7z archive contains the 40 packaged sources; install lands at `Java/platform-independent/org/coolprop/*.java`.
- Flat no-package build: sources stay flat, 7z archives 40 files (was 0), install is the flat pre-#3245 layout, no build internals leak.
- Single-segment `-package coolprop`: installs `platform-independent/coolprop/`, not the binary dir; `-package .coolprop` fails loud at configure.
- Preflight: 4 passed / 0 failed / 5 skipped (no C++ changes). Adversarial review + delta re-review: one blocking finding (the single-segment install) — fixed and re-verified; residual edge closed with the configure-time guard.

## Follow-ups filed (pre-existing #3245 regressions, out of scope here)

- **CoolProp-qrn3**: the unified Java release artifact ships without wrapper sources (`install(FILES platform-independent.7z)` dropped by #3245 + fail-open `|| true` copy in the `java_builder.yml` merge step).
- **CoolProp-jrvp**: `Web/coolprop/wrappers/Java/index.rst` still documents the flat no-package build whose own `Javatestbuild` ctest now fails for the same import reason.

Closes bd issue CoolProp-7c3b.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Java wrapper packaging and installation when no SWIG package is specified.
  * Prevented unintended relocation or installation of unrelated build files.
  * Added validation for invalid Java package paths.

* **Build Improvements**
  * Example Java builds on Linux and macOS now use the standard `org.coolprop` package consistently.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1785,14 +1785,19 @@ if(COOLPROP_JAVA_MODULE)
   set_target_properties(CoolProp PROPERTIES OUTPUT_NAME "CoolPropJava")
   coolprop_hide_json_symbols(CoolProp)
 
-  # Relocate SWIG-generated Java files into the package directory (cross-platform)
-  add_custom_command(
-      TARGET CoolProp
-      POST_BUILD
-      COMMAND ${CMAKE_COMMAND} -D SRC_DIR="${CMAKE_CURRENT_BINARY_DIR}"
-                               -D DST_DIR="${CMAKE_CURRENT_BINARY_DIR}/${JAVA_PACKAGE_DIR}"
-                               -P "${CMAKE_CURRENT_SOURCE_DIR}/dev/cmake/relocate_java.cmake"
-  )
+  # Relocate SWIG-generated Java files into the package directory (cross-platform).
+  # Only when a -package was requested via COOLPROP_SWIG_OPTIONS: without one the
+  # generated sources belong in the default package (flat in the binary dir) and
+  # relocating would run with DST_DIR == SRC_DIR.
+  if(NOT JAVA_PACKAGE_DIR STREQUAL "")
+    add_custom_command(
+        TARGET CoolProp
+        POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -D SRC_DIR="${CMAKE_CURRENT_BINARY_DIR}"
+                                 -D DST_DIR="${CMAKE_CURRENT_BINARY_DIR}/${JAVA_PACKAGE_DIR}"
+                                 -P "${CMAKE_CURRENT_SOURCE_DIR}/dev/cmake/relocate_java.cmake"
+    )
+  endif()
 
   if(WIN32)
     set_target_properties(CoolProp PROPERTIES PREFIX "")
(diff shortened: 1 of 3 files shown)


ibell:
@pkirkham can you take a look at this PR?

ibell:
**Note on this PR's red docs check:** the docs run on this PR's merge ref gets past the previously-failing Java example step (validating this fix) and then fails at `Web/coolprop/Tabular.rst:250` with the pre-existing empty-`ValueError` mixture-table bug — the independent breakage fixed by #3255. The two docs failures interlock: neither fix can produce a green docs run against master alone, so the PRs are stacked (**merge order: #3267 → #3255 → #3266**) and #3255's merge ref (= both fixes) is the one expected to go green.

ibell:
**Decomposition of this PR's three failing checks** (none caused by this diff — it touches only the Java-example CMake section and two runner scripts):

1. **docs `build`** — expected: gets past the Java step this PR fixes, then hits the pre-existing `Tabular.rst` mixture-table `ValueError` (#3239 regression, fixed by stacked #3255). See previous comment.
2. **`nanobind wheels`** — same root cause via a second harness: `test_doc_examples.py::test_doc_page_examples[coolprop/Tabular.rst]` executes the doc-page example and hits the identical `ValueError`. Also resolves in the stack at #3255.
3. **`address sanitizer`** — systemic, unrelated: the ASAN job times out at its 90-min cap on **every** PR since July 7. The last green run (July 6) finished in 1h29m56s — four seconds under the limit — and master didn't change in the July 6→7 window (same runner image), so accumulated test growth has permanently crossed the fixed budget. Evidence logged in bd CoolProp-xuu (options there: drop `ASAN_OPTIONS=verbosity=1`, which emits ~255k poisoning log lines per run; raise the timeout; or the planned tag split).

PR #3273, opened 2026-07-16
perf(flash): delta-only α^r for the HEOS mixture density solve (~3×)

andr1976:
## Summary

Speeds up the **HEOS mixture flash** by ~3× on density-solve-dominated states, with **bit-identical** results.

Profiling a 5-component natural-gas `PT` flash showed ~85 % of the time is residual-Helmholtz-energy (α^r) evaluation, dominated by the spinodal density search (`solver_dpdrho0_Tp`). That search's `dpdrho_resid` was computing the **full 15-derivative** α^r set **~3× per Halley iteration** — once in `update_DmolarT_direct` and again in each of `calc_d3alphar_dDelta3` / `calc_d4alphar_dDelta4`, which re-run the whole `all()`.

But those residuals only need **δ-derivatives** (pressure, dp/dρ, d²p/dρ², d³p/dρ³) — never τ-derivatives. This PR adds a δ-only α^r evaluation and routes the density-root residuals through it.

## Changes

- **`ResidualHelmholtzGeneralizedExponential::all_deltaonly`** (`src/Helmholtz.cpp`): `all()` with every τ-derivative and mixed term deleted — the `u`-value (all six contributions, so `ndteu` is identical), the δ-derivatives of `u` up to 4th order, the `B_delta…B_delta4` chain, and the α^r / dα^r_dδ / … / d⁴α^r_dδ⁴ accumulation + scaling. Every surviving expression is **term-for-term identical** to `all()`, so the five populated δ-fields are **bit-for-bit equal**.
- **Plumbing**: a `virtual all_deltaonly()` on `BaseHelmholtzTerm` defaulting to the full `all()` (correct for every term type; only the generalized-exponential term overrides it), threaded through `ResidualHelmholtzContainer`, `CorrespondingStatesTerm`, `DepartureFunction`/`ExcessTerm` (the GERG departure reuses the same generalized-exponential term, so it gets the speedup too), and the mixture `ResidualHelmholtz`.
- **`calc_residual_deltaonly`** + a δ-only route in `calc_alphar_deriv_nocache` (any `nTau==0, nDelta≤4` request; bit-exact, so transport-property pressure evals benefit too).
- **`dpdrho_resid`** now computes the δ-only derivatives **once per ρ** in `call()` and reuses them in `deriv()`/`second_deriv()` (Halley always calls those at the same ρ), eliminating both the τ-derivative work and the redundant recomputes. It no longer writes the full state; the one downstream reader of that side effect (`this->p()` in the heavy slow-path) now reads a stored `p_last`. `SolverTPResid` deliberately keeps `update_DmolarT_direct` because `solver_rho_Tp` reads `first_partial_deriv` from the post-solve state.

## Correctness

- New **`[deltaonly]`** test (`CoolProp-Tests-DeltaOnly.cpp`): bit-exactness of `all_deltaonly` vs full `all()` over a (τ,δ) grid for a pure fluid (generalized-exponential term), a GERG mixture (departure/excess term) and CO₂/Water (non-analytic term) — **1218 assertions, 0 mismatches**.
- `[flash]`, `[mixture]`, `[Michelsen]`, and `[consistency]` (**24 991 assertions**) all pass.
- Density roots unchanged to the last digit (a 100-flash NG-mixture sweep gives a bit-identical accumulator before/after).

## Performance

Controlled interleaved A/B (same machine, driver loops `update(PT_INPUTS)` over a representative NG P-T grid, only the library differs), 5×40 flashes:

| | ms/flash | ratio |
|---|---|---|
| master | 283 | — |
| this PR | 90 | **3.1×** |

(Absolute numbers scale with machine load; the interleaved ratio is the robust figure and is consistent with the mechanism: 3 full `all()` → 1 ~half-cost δ-only eval per Halley iteration.)

## Scope

This is the self-contained, zero-algorithmic-risk optimization (it changes *how cheaply* α^r derivatives are computed, not *which* density root is found). Two further levers — a warm-start fix for `solve_trial_rho_warm`, and a stable-root strategy that avoids the global spinodal scan — are deliberately left as follow-ups since they can change which root is selected.

## Summary by CodeRabbit

* **Performance Improvements**
  * Added a “delta-only” evaluation mode for residual Helmholtz derivatives used in density- and pressure-related calculations.
  * Avoids computing unused tau/mixed derivative components during targeted derivative solving.

* **Reliability**
  * Ensures delta-only residual derivative results match the existing full evaluation path exactly (bit-for-bit).

* **Testing**
  * Added a new Catch2 test runner unit and a dedicated DeltaOnly test covering pure fluids and mixtures over a grid of evaluation points.

### DIFF
--- a/src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
+++ b/src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
@@ -2662,6 +2662,14 @@ HelmholtzEOSBackend::StationaryPointReturnFlag HelmholtzEOSMixtureBackend::solve
        public:
         HelmholtzEOSMixtureBackend* HEOS;
         CoolPropDbl T, p, delta, rhor, tau, R_u;
+        // Residual-alphar delta-derivatives at the current rho, plus the pressure there.  call()
+        // computes them once (delta-only: no tau-derivatives, no state write) and deriv()/
+        // second_deriv() reuse them -- the Halley solver always invokes those at the same rho right
+        // after call().  The previous path wrote the full state and recomputed the FULL derivative
+        // set in each accessor; p_last preserves the pressure the heavy slow-path below needs
+        // (it read this->p(), which used to be a side effect of the dropped state write).
+        HelmholtzDerivatives d;
+        CoolPropDbl p_last;
 
         dpdrho_resid(HelmholtzEOSMixtureBackend* HEOS, CoolPropDbl T, CoolPropDbl p)
           : HEOS(HEOS),
@@ -2670,21 +2678,22 @@ HelmholtzEOSBackend::StationaryPointReturnFlag HelmholtzEOSMixtureBackend::solve
             delta(_HUGE),
             rhor(HEOS->get_reducing_state().rhomolar),
             tau(HEOS->get_reducing_state().T / T),
-            R_u(HEOS->gas_constant()) {}
+            R_u(HEOS->gas_constant()),
+            p_last(_HUGE) {}
         double call(double rhomolar) override {
             delta = rhomolar / rhor;  // needed for derivative
-            HEOS->update_DmolarT_direct(rhomolar, T);
+            d = HEOS->calc_residual_deltaonly(tau, delta);
+            p_last = rhomolar * R_u * T * (1 + delta * d.dalphar_ddelta);
             // dp/drho|T
-            return R_u * T * (1 + 2 * delta * HEOS->dalphar_dDelta() + POW2(delta) * HEOS->d2alphar_dDelta2());
+            return R_u * T * (1 + 2 * delta * d.dalphar_ddelta + POW2(delta) * d.d2alphar_ddelta2);
         };
         double deriv(double rhomolar) override {
             // d2p/drho2|T
-            return R_u * T / rhor * (2 * HEOS->dalphar_dDelta() + 4 * delta * HEOS->d2alphar_dDelta2() + POW2(delta) * HEOS->calc_d3alphar_dDelta3());
+            return R_u * T / rhor * (2 * d.dalphar_ddelta + 4 * delta * d.d2alphar_ddelta2 + POW2(delta) * d.d3alphar_ddelta3);
         };
         double second_deriv(double rhomolar) override {
             // d3p/drho3|T
-            return R_u * T / POW2(rhor)
-                   * (6 * HEOS->d2alphar_dDelta2() + 6 * delta * HEOS->d3alphar_dDelta3() + POW2(delta) * HEOS->calc_d4alphar_dDelta4());
+            return R_u * T / POW2(rhor) * (6 * d.d2alphar_ddelta2 + 6 * delta * d.d3alphar_ddelta3 + POW2(delta) * d.d4alphar_ddelta4);
         };
     };
     dpdrho_resid resid(this, T, p);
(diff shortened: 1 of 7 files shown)


ibell:
This is awesome and was on my speed optimization TODO list. Nice work. I'll need to review closely before merging, but conceptually this seems good.

src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
```diff
@@ -3583,10 +3592,22 @@ void HelmholtzEOSMixtureBackend::calc_all_alphar_deriv_cache(const std::vector<C
...
+    // needs a tau-derivative (nTau > 0) or an out-of-range order falls through to the full path.
+    if (nTau == 0 && nDelta >= 0 && nDelta <= 4) {
+        HelmholtzDerivatives derivs = residual_helmholtz->all_deltaonly(*this, mole_fractions, tau, delta);
+        return derivs.get(0, nDelta);
+    }
     bool cache_values = false;
     HelmholtzDerivatives derivs = residual_helmholtz->all(*this, mole_fractions, tau, delta, cache_values);
     return derivs.get(nTau, nDelta);
 }
+HelmholtzDerivatives HelmholtzEOSMixtureBackend::calc_residual_deltaonly(const CoolPropDbl& tau, const CoolPropDbl& delta) {
```
ibell: I don't like this function name. Not very clear about what it is doing

ibell (changes requested):
There is a pretty big gap here: this approach is only applied to ResidualHelmholtzGeneralizedExponential and none of the other term classes get the same treatment. To be fair, that is ok and will already make a big difference, but for water or CO2, the nonanalytic term will not get the same benefit and that term is also costly

andr1976:
Thanks for the review @ibell! Addressed all three points in the latest commit:

**1. Non-analytic term (your main point).** You're right that it was the obvious gap — for water/CO₂ the non-analytic term is a real cost. Added `ResidualHelmholtzNonAnalytic::all_deltaonly`: it's `all()` with every τ-derivative and mixed term removed, each surviving δ-expression copied verbatim so the δ-fields are bit-for-bit identical. Extended the `[deltaonly]` bit-exactness test to **pure Water and CO₂** (which drive the non-analytic term) — 2030 assertions, 0 mismatches.

Isolated impact on a CO₂-rich CO₂/water PT flash (controlled interleaved A/B, density roots bit-identical): **~1.1× on top of the generalized-exponential change** (≈1.8–2× vs master), so the non-analytic term now gets the same benefit.

The remaining term classes (SAFT, cubic, Xiang-Deiters, Gao-B) still fall through to the default (full `all()`), which is correct but unaccelerated — happy to extend those too if you'd like, though they're empty for the fluids in the current benchmarks.

**2. Function name.** Renamed `calc_residual_deltaonly` → `calc_alphar_delta_derivs_nocache` to match the existing `calc_alphar_deriv_nocache` / `calc_all_alphar_deriv_cache` family and say what it returns.

**3. (@coderabbitai) `bits_of` long-double truncation.** Fixed — it now takes `CoolPropDbl` and compares the full object representation via `std::array<unsigned char, sizeof(CoolPropDbl)>`, so the bit-exactness check stays valid under the `long double` configuration.

`[flash]`, `[mixture]`, and `[consistency]` (24 991 assertions) all still pass.

src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
andr1976: Renamed to `calc_alphar_delta_derivs_nocache` in 3a9b3a8a — parallels the existing `calc_alphar_deriv_nocache` / `calc_all_alphar_deriv_cache` family and says it returns the residual α^r delta-derivatives (orders 0–4) without caching. Open to a different name if you'd prefer.

src/Tests/CoolProp-Tests-DeltaOnly.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+namespace {
+
+// Reinterpret a double's storage as a uint64 so two values compare bit-for-bit (memcpy avoids the
+// suspicious-memory-comparison pitfall of memcmp on floating-point).
+std::uint64_t bits_of(double d) {
+    std::uint64_t u = 0;
+    std::memcpy(&u, &d, sizeof(u));
+    return u;
+}
```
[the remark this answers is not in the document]
andr1976: Fixed in 3a9b3a8a: `bits_of` now takes `CoolPropDbl` and compares the full object representation via `std::array<unsigned char, sizeof(CoolPropDbl)>`, so the bit-exactness check stays valid under the long-double configuration. Thanks!

Issue #2558, opened 2025-06-14
querying `Tau` and `Delta` from multicomponent models fails


longemen3000:
MVE:
```julia
julia> PropsSI("Tau","Dmolar|gas",200.0,"T",300.0,"CarbonDioxide[0.5]&Hydrogen[0.5]")
ERROR: CoolProp: This input [22: "Tau"] is not valid for keyed_output : PropsSI("Tau","Dmolar|gas",200,"T",300,"CarbonDioxide[0.5]&Hydrogen[0.5]")
Stacktrace:
 [1] error(::String, ::String)
   @ Base .\error.jl:44
 [2] PropsSI(output::String, name1::String, value1::Float64, name2::String, value2::Float64, fluid::String)
   @ CoolProp C:\Users\andre\.julia\packages\CoolProp\4FVeh\src\CoolProp.jl:83
 [3] top-level scope
   @ REPL[181]:1

julia> PropsSI("Tau","Dmolar",200.0,"T",300.0,"Hydrogen")
ERROR: CoolProp: This input [22: "Tau"] is not valid for keyed_output : PropsSI("Tau","Dmolar",200,"T",300,"Hydrogen")
Stacktrace:
 [1] error(::String, ::String)
   @ Base .\error.jl:44
 [2] PropsSI(output::String, name1::String, value1::Float64, name2::String, value2::Float64, fluid::String)
   @ CoolProp C:\Users\andre\.julia\packages\CoolProp\4FVeh\src\CoolProp.jl:83
 [3] top-level scope
   @ REPL[182]:1

julia> PropsSI("Delta","Dmolar|gas",200.0,"T",300.0,"CarbonDioxide[0.5]&Hydrogen[0.5]")
ERROR: CoolProp: This input [23: "Delta"] is not valid for keyed_output : PropsSI("Delta","Dmolar|gas",200,"T",300,"CarbonDioxide[0.5]&Hydrogen[0.5]")
Stacktrace:
 [1] error(::String, ::String)
   @ Base .\error.jl:44
 [2] PropsSI(output::String, name1::String, value1::Float64, name2::String, value2::Float64, fluid::String)
   @ CoolProp C:\Users\andre\.julia\packages\CoolProp\4FVeh\src\CoolProp.jl:83
 [3] top-level scope
   @ REPL[183]:1

julia> PropsSI("Delta","Dmolar",200.0,"T",300.0,"Hydrogen")
[...]

Issue #2675, opened 2026-02-08
2026 equation of state for Carbon dioxide


longemen3000:
https://www.nist.gov/publications/alternative-fundamental-equation-state-fluid-carbon-dioxide

What is the stance on changing a pure fluid EoS? i know that methanol was changed before, but CO2 is one of the more used multiparameter equations in CoolProp.

longemen3000:
I will see what i can do. I'm somewhat familiarized with the evaluation routines in CoolProp, but the parsing and configuration is another different area hahah

ibell:
I think we want to allow an interface something like this: `PropsSI('T','P',101325,'Q',0,'Water{"EOS":"Wagner-JPCRD-2002"}')` where the stuff in the `{}` is a JSON-compatible string, and allows you to select a given model based on the key of the model. Ideally we add the same for other kinds of models, but `EOS` is a good start.

ibell:
There is now a mechanism for adding additional flags when loading a model: https://coolprop.github.io/devdocs/coolprop/BackendOptions.html . Such a thing could be used to define a grammar for selecting the EOS to be used at model-load time (like this case!)

PR #3278, opened 2026-07-19
feat(mixtures): add Beckmüller-2021 H2 binary models (H2 + CH4/N2/CO/CO2)

### LINKED ISSUE #2263: new departure functions for H2 + (CH4, N2, CO, CO2)
Do you want to ask a question? Are you looking for technical support? For any kind of question regarding CoolProp and its usage, you can ask the [CoolProp user group](https://goo.gl/Pa7FBT)

If you have an issue to report, please continue and fill out the applicable sections below.  The details provided will help to resolve this issue as quickly as possible.

### Prerequisites

Before posting an issue, please:  
    * Check the [CoolProp User Group](https://goo.gl/Pa7FBT) for common solutions.  
    * Check that your issue isn't already filed under [CoolProp Issues](https://github.com/CoolProp/CoolProp/issues).  
    * Check the CoolProp [latest release](http://www.coolprop.org) and [development version](http://www.coolprop.org/dev) documentation.  

### Description

There are new departure functions for Hydrogen mixtures +  (CH4, N2, CO, CO2), published here : [
https://doi.org/10.1063/5.0040533
](
https://doi.org/10.1063/5.0040533
)

And it seems that CoolProp does not use it. is it ok to add those to the departure JSON? (and link those departures to the appropiated component pairs in the mixtures JSON

andr1976:
## Summary

Adds the binary hydrogen mixture models of **Beckmüller, Thol, Bell, Lemmon & Span**, *"New Equations of State for Binary Hydrogen Mixtures Containing Methane, Nitrogen, Carbon Monoxide, and Carbon Dioxide"*, J. Phys. Chem. Ref. Data **50**, 013102 (2021), [doi:10.1063/5.0040533](https://doi.org/10.1063/5.0040533), for the pairs **H₂ + {CH₄, N₂, CO, CO₂}**.

Closes #2263.

## Why this is a pure data addition

The models use the GERG-2008 mathematical form (composition-dependent reducing functions + binary-specific departure functions with polynomial + exponential + modified Gaussian-bell-shaped terms), which CoolProp already implements as the **`Gaussian+Exponential`** departure type. So **no C++ changes** are needed — only mixture data.

## Changes

- **`mixture_departure_functions.json`** — 4 new `Gaussian+Exponential` departure functions (Supplementary Material Table S3).
- **`mixture_binary_pairs.json`** — the 4 H₂ pairs get `F = 1`, the new reducing parameters (Table S2), and a link to the new departure functions. Previously **only H₂-CH₄ had a departure function**; H₂-N₂/CO/CO₂ were reducing-only. For **H₂-CO**, which CoolProp stores in reversed (H₂, CO) order, the reducing betas are inverted (`β_ij = 1/β_ji`; γ unchanged) per the reducing-function symmetry.
- **`CoolPropBibTeXLibrary.bib`** — new `Beckmueller-JPCRD-2021` entry, so the auto-generated binary-pairs documentation table (`Web/fluid_properties/Mixtures.rst`, built from the library data) resolves the citation for these pairs.
- **`dev/mixtures/add_h2_beckmuller_2021.py`** — reproducible generator for the two JSON edits (documents the source data and the ordering/inversion logic).
- **`CoolProp-Tests-H2Mixtures.cpp`** — reproduces the paper's **full** single-phase computer-implementation test table (Supplementary Table S6, x_H₂ = 0.4).

## Validation

Reproduces **all 16 × 6 tabulated properties** of Table S6 — pressure, isobaric heat capacity, speed of sound, enthalpy, entropy, and Helmholtz energy — **to ~1–4 ppm** (96 assertions):

- `p`, `cp`, `w` are reference-state-independent.
- `h`, `s`, `a` match too because CoolProp uses the same pure-fluid reference states (Table S1) as the paper — including CoolProp's updated normal-hydrogen reference (Tc = 33.145 K, ρc = 15.508 mol/dm³, verified identical).

The residual ~1–4 ppm is the **gas-constant convention** (the paper uses GERG-2008 R = 8.314472 J/mol/K for all components; CoolProp uses each pure fluid's native reference-EOS R — 8.31451 for CH₄/N₂/CO₂). Consistent with this, the CO+H₂ mixture — where both fluids use R = 8.314472 — reproduces to the tightest ~1.1 ppm. The test tolerance (2e-5) accommodates this convention plus the limited digits of the published parameters.

## Note

This **changes results** for existing H₂ + {CH₄, N₂, CO, CO₂} mixture calculations (an accuracy improvement over GERG-2008), so it is **not bit-identical** with prior output. No existing test asserts fixed reference values for these pairs (the one H₂-CO₂ test, #2558, only checks finiteness).

## Summary by CodeRabbit

- **New Features**
  - Added Beckmüller (2021) hydrogen binary-mixture models for methane, nitrogen, carbon dioxide, and carbon monoxide.
  - These models are now used by default; earlier published models are retained but not selectable through the API.
  - Added validation against published hydrogen-mixture reference data.

- **Bug Fixes**
  - Improved handling of duplicate interaction parameters and unsupported runtime overrides.

- **Documentation**
  - Documented model precedence, record ordering, and handling of alternate interaction-parameter models.


dev/mixtures/mixture_binary_pairs.json
ibell, line 1: In general I think that we should not overwrite the coefficients, rather they should be added as a prefix or suffix (I cannot remember what the overwriting rules are when the same component pairs appear more than once). I am thinking of the world where we enable GERG-2008, EOS-CG-2021, current defaults, etc.

src/Tests/CoolProp-Tests-H2Mixtures.cpp
```diff
@@ -0,0 +1,81 @@
+// Regression test for the binary hydrogen mixture models of Beckmueller, Thol, Bell, Lemmon & Span,
+// "New Equations of State for Binary Hydrogen-Mixtures Containing Methane, Nitrogen, Carbon Monoxide,
+// and Carbon Dioxide", J. Phys. Chem. Ref. Data 50, 013102 (2021); GitHub issue #2263.
+//
+// The models add binary-specific departure functions (GERG-2008 "Gaussian+Exponential" form) and
+// updated reducing parameters for H2 + {CH4, N2, CO, CO2}.  This test reproduces the single-phase
+// computer-implementation test values from Table S6 of the Supplementary Material (x_H2 = 0.4).
+//
+// Tolerance: the paper's reference values use the GERG-2008 gas constant R = 8.314472 J/mol/K for all
```
ibell: I don't think this is correct. CoolProp by default uses the CODATA value for gas constant for mixtures, but that can be overridden by config to use a mole-fraction-weighted average of the values of the pures, but that comes with its own issues because the derivatives of the gas constant of the mixture are not included in the derivatives. Normally that is a tiny correction, except for in the critical region.

ibell (changes requested):
Thank you for taking this on! A few comments, but very close to being able to merge. These updates make the binary pairs so much more accurate that I would propose they become the default.

andr1976:
Thanks Ian — glad it's close, and agreed these should become the default.

On **"don't overwrite the coefficients"**: I dug into the loader and I think your **eb7caaa (#2675, EOS selection via JSON suffix)** is exactly the framework to build on. State on the binary side today:

- `m_binary_pair_map` already stores `CAS → vector<Dictionary>`, so the *storage* for multiple models per pair exists — but the loader collapses it to a single element (last-wins under `OVERWRITE_BINARY_INTERACTION`, otherwise it throws on a duplicate CAS pair), and everything downstream reads `v[0]`.
- Departure functions already coexist by unique `Name` (the new `*-Beckmueller2021` entries are *added*, not overwritten); it's only the binary-*pair* record (reducing params + the `function` link) that's single-valued.
- The selection layer from eb7caaa (`selected_EOS_index`, `select_eos_by_index` / `select_eos_by_bibtex`, the `Fluid{"EOS":...}` suffix) is pure-fluid only — there's no binary analogue yet.

So the clean way to get "GERG-2008 / EOS-CG-2021 / current defaults all available" is to mirror eb7caaa for binary interactions: keep multiple entries in the per-pair vector, add `selected_binary_index` + `select_binary_by_bibtex`, and a hook so a `...{"binary":"Beckmueller-JPCRD-2021"}`-style suffix (or a config key) selects the model — then this PR (and the EOS-CG PR #3281, which uses the same overwrite pattern) become purely *additive*: GERG stays, Beckmüller / EOS-CG are added and one is marked default.

Two ways to sequence it:

1. A small **precursor PR** extending eb7caaa to binary interactions, after which #3278 / #3281 are clean data-only adds.
2. Merge the data now as the new default (overwrite) and I open a **follow-up issue** to build the binary-selection layer on your framework (GERG-2008 stays reachable via the dedicated backend in the meantime).

Any preference? I'm happy to take on the framework extension either way — it looks modest given the vector storage already exists and eb7caaa is the blueprint.

Separately I'll address the rest: the gas-constant note in the test is wrong (mixture `R` defaults to **CODATA** via `NORMALIZE_GAS_CONSTANTS`, not the pure-fluid values — I'll correct the comment), plus the CodeQL file-close findings and the CodeRabbit nit in the generator script, and I'll rebase to clear the `CMakeLists.txt` conflict.

ibell:
I think that https://github.com/CoolProp/CoolProp/commit/eb7caaa3d6bff7308f999a405687dee0619fd29f is a bad idea. It was a spike that petered out once we started building it. 

I think it would be best to just add the binary parameters in the right place so that they become default. I added the https://github.com/CoolProp/CoolProp/pull/3314 with GERG-2004 and GERG-2008 strict EOS and mixture models, so the concern about overwriting is a bit weaker. I still am not sure what to do about EOS-CG-2021.

andr1976:
@ibell so in the right place means not updating Kunz-JCED-2012 as done ?

ibell:
Yes that is correct. It should be default, but not overwrite the previous
parameters.

andr1976:
@ibell done — I've restored the four Kunz-JCED-2012 records, so the data diff is now purely additive (52 insertions, 0 deletions) and Beckmüller becomes the default rather than replacing anything.

Since you weren't sure what the overwriting rules were, here is what I found: a repeated CAS pair was a hard load error. `MixtureBinaryPairLibrary::load_from_JSON` threw unless `OVERWRITE_BINARY_INTERACTION` was set, and that key *erased* the earlier record. So two published models genuinely could not coexist, and "default but not overwrite" wasn't expressible in the data file alone — it needed a small change in the loader.

The storage for it turned out to already exist, unused: `m_binary_pair_map` is `CAS -> vector<Dictionary>`, but the loader only ever built a one-element vector and all four consumers read `[0]`. A repeated pair in the built-in library now appends instead, keeping `[0]` — the first record in document order — as the model in force. A newer correlation therefore becomes the default by being listed *ahead of* the one it supersedes, without deleting it. No selection API, no new config key, no name-suffix parsing: I dropped the eb7caaa direction entirely, as you asked.

I deliberately scoped that latitude to the built-in load. `set_interaction_parameters` and the REFPROP `HMX.BNC` import share the same loader, and filing a caller's parameters away as an unread alternate would look exactly like success while changing nothing — so run-time data still errors on a known pair, as documented. I hit that fail-open while implementing this, and both halves are now covered by tests; neither path had coverage before.

To be straight about the limitation: nothing *reads* an alternate, so the retained Kunz parameters are preserved rather than reachable. For these four pairs `GERG2008::` from #3314 gives them in full, which I think is what you meant when you said the concern was weaker.

I also rebased onto master, which clears the `CMakeLists.txt` conflict, and documented the ordering rule in `Web/fluid_properties/Mixtures.rst` so the next person adding a superseding model doesn't have to rediscover it.

Verification: the four H2 pairs are pinned three ways — the Table S6 values, the default record's BibTeX/`gammaT`, and a check that both records are present in the right order. I mutation-tested both new guards: deleting a Kunz record fails the retention test naming the pair, and moving a Kunz record ahead of its Beckmüller counterpart fails three test cases across 29 assertions. Full non-slow suite is green (181,429 assertions).

andr1976:
CodeRabbit's finding on `MixtureParameters.cpp` was correct and I've fixed it in 22ffbed5. It was a real bug, not a nit: the shipped library loads lazily behind a `std::call_once`, and `Web/fluid_properties/Mixtures.rst` tells users to set `OVERWRITE_BINARY_INTERACTION` *before* their first mixture call. Because I tested that key ahead of the shipped-load branch, such a caller was in overwrite mode while the shipped records were still being read — so each superseded record replaced *and erased* the newer one it is meant to sit behind. Reproduced before fixing: with the key set first, `Methane+Hydrogen` came back as `Kunz-JCED-2012` and the Beckmüller record was gone. Precedence among shipped records is now document order and never consults that key.

Since you'll want to agree to these rather than discover them, here is everything this PR changes about **pre-existing** behaviour.

**1. Results change for the four H2 pairs.** This is the intent, but the size is worth seeing — and three of the four previously had no departure function at all (`F = 0`), so they gain an excess term they never had:

| pair | F before | F after | gammaT | gammaV |
|---|---|---|---|---|
| Methane+Hydrogen | 1.0 | 1.0 | 1.352643115 → 1.44 | 1.018702573 → 0.804 |
| Nitrogen+Hydrogen | **0** | 1.0 | 1.175696583 → 1.24 | 0.970115357 → 0.773 |
| CarbonDioxide+Hydrogen | **0** | 1.0 | 1.782924792 → 1.961 | 1.15279255 → 0.842 |
| Hydrogen+CarbonMonoxide | **0** | 1.0 | 1.377504607 → 1.105 | 1.121416201 → 1.04 |

Anyone currently computing CO2/H2 or N2/H2 will see materially different numbers.

**2. A repeated CAS pair in the shipped library is no longer an error.** Before, `load_from_JSON` threw on the second occurrence, so the shipped file could never contain duplicates at all. The trade-off you should weigh: an *accidental* duplicate is now a silent no-op instead of a loud failure. The four intentional pairs are pinned by a test asserting exactly two records in the right order, but a new accidental duplicate elsewhere would not be caught. If you'd rather keep it loud, the alternative is an explicit allowlist of pairs permitted to repeat — happy to add that instead.

**3. The precedence direction is the reverse of the only pre-existing multi-record behaviour.** Your overwrite path was *last*-in-document-order wins, with the earlier record erased. Mine is *first* wins, with later records retained. I chose first because all four consumers already read `[0]` and the code there calls it "the first matching binary pair", and because it reads naturally to list a newer model ahead of the one it supersedes. But it does mean appending a record for an existing pair has no effect, which is the trap in this direction. Say the word and I'll flip it to last-wins — the newer record then inserts at `[0]` and placement stops mattering.

**4. `OVERWRITE_BINARY_INTERACTION` is now scoped to run-time input only.** No observable change before this PR, since the shipped file had no duplicates for it to act on, but its meaning is narrower now and the docs say so.

**5. Deliberately unchanged:** `set_interaction_parameters` and the REFPROP `HMX.BNC` import still throw on an already-known pair unless that key is set. I kept that on purpose — filing a caller's parameters away as an unread alternate would look exactly like success while changing nothing.

Full non-slow suite is green (549 cases, 181,429 assertions). The two older CodeRabbit findings on `dev/mixtures/add_h2_beckmuller_2021.py` are moot — that script was untracked in cd3459ac.

andr1976:
Following up on point 3 of my last comment — I've gone ahead and flipped it, so please read that point as settled rather than as an open question.

Precedence among several shipped records for a pair is now **last-in-document-order wins**: later data supersedes earlier (f59e38d2). That matches the direction the code always had for a repeated pair — your `OVERWRITE_BINARY_INTERACTION` path was last-wins — so there is no longer one rule for run-time input and the opposite rule for the shipped library. The one difference remains the point of the exercise: that path *erased* the superseded record, whereas this keeps it.

It also removes a trap I'd built in. Under first-wins, appending a record for an existing pair had no effect at all — a contributor could add a model and ship nothing. Under last-wins a stray record changes results and the tests move, so it fails loudly instead of silently.

Practically: **adding a superseding correlation is now just an append.** The four H2 pairs are laid out that way — the Kunz-JCED-2012 records are back in their original positions with the Beckmüller records directly after them — so the data diff against master is still purely additive, 52 insertions and 0 deletions.

Two notes on what this changed elsewhere:

- I dropped the fresh-process regression test I'd added for the `OVERWRITE_BINARY_INTERACTION` leak. Under this layout the leak would erase the earlier record and keep the later one, which is still the correct model in force, so the test could no longer observe anything and would have passed vacuously. The branch ordering that prevents the leak is still there — it protects the *retained* records — but that invariant isn't reachable through the API, so it carries a comment rather than a test. I'd rather tell you that than leave a green test that checks nothing.
- The remaining two guards were re-verified non-vacuous against the new layout: deleting a Kunz record fails the retention test naming the pair, and moving a Beckmüller record ahead of its Kunz counterpart fails three test cases across 29 assertions.

Everything else in my previous comment stands, in particular the results change for the four pairs and the fact that three of them previously had `F = 0`. Full non-slow suite is green (549 cases, 181,429 assertions).

andr1976:
Both of CodeRabbit's findings on f59e38d2 were valid and are fixed in 9aee7894.

**The stale contract.** The `@param allow_multiple_models` block still said the *first* record is the model in force — I updated the branch comment in the last-wins flip and missed the parameter doc immediately above it. Corrected to state last-wins, and that the vector order is deliberately the reverse of document order.

**The ordering one, which was the real find.** `load_defaults_if_needed()` is reached only from `binary_pair_map()` and `apply_simple_mixing_rule()` — never from `set_interaction_parameters()` or the REFPROP `#MXM` path. Calling either as the first library action therefore dropped the caller's records into an *empty* map as brand-new entries; the shipped records then arrived afterwards and, being newer in document order, were prepended and won. The caller's parameters became an unread alternate: no error, no effect. I reproduced it in a fresh process before touching anything — no exception, and the pair reported `Kunz-JCED-2012`.

`load_from_JSON` now loads the shipped library first whenever `allow_multiple_models` is false, so run-time data always merges into the complete map and is either rejected or honoured per `OVERWRITE_BINARY_INTERACTION`. It is guarded on that flag so the shipped load cannot re-enter the `std::call_once` and deadlock; the new test drives exactly that path, so it would hang rather than pass if it could.

**Worth flagging separately, because it is not mine:** this also closes a pre-existing silent loss on master. Following the override workflow in `Mixtures.rst` — set `OVERWRITE_BINARY_INTERACTION`, then hand in parameters — but reaching `set_interaction_parameters()` before any property call, the caller's record is discarded today. I measured it both ways rather than inferring it:

| | gammaT in force | BibTeX |
|---|---|---|
| without the fix | 0.979273013 | `Kunz-JCED-2012` — caller's record silently lost |
| with the fix | 1.234 | `test-not-a-real-model` — caller's record applies |

Same applies to importing an `HMX.BNC` before first use. Happy to split that into its own PR if you'd rather keep this one to the H2 data; it is a two-line change sitting in the same function, which is why I did it here.

One note on the new guard, since I'd rather say it than let a green tick imply more than it should: it is order-sensitive. It is strongest as the first library action in a process and merely duplicates the adjacent refusal case once the library is loaded, so it carries no shared tag and is run deliberately — `CatchTestRunner "[runtime_before_defaults]"`. Sharing `[mixtures]`/`[hydrogen]` would have swept it into runs where it passes without exercising anything, which is a trap I walked into earlier in this PR and removed.

Full non-slow suite green (549 cases, 181,429 assertions).

andr1976:
@coderabbitai review

andr1976:
@coderabbitai review

andr1976:
@coderabbitai full review

Issue #3171, opened 2026-06-11
Michelsen stability false-positive: methanol/benzene at 308.15 K / 1 atm flagged two-phase (single-phase liquid)


andr1976:
### Summary

With `MIXTURE_STABILITY_ALGORITHM = 1` (Michelsen, the v8 default), the blind PT
flash classifies **methanol & benzene** (HEOS) at **308.15 K, 101325 Pa,
x_methanol ≈ 0.54** as **two-phase**, but at these conditions the mixture is
subcooled **single-phase liquid** (bubble pressure ≈ 34 kPa ≪ 101 kPa). The
stability test (`StabilityEvaluationClass::is_stable()` → `false`) routes the
feed into the two-phase phase-split solver, which then chases a **spurious
split** in which *both* phases are liquid-like (vapour denser than liquid):

```
beta ≈ 0.799   x ≈ [0.238, 0.762]   y ≈ [0.616, 0.384]
rho_liq ≈ 12634 mol/m³   rho_vap ≈ 16638 mol/m³   (rho_vap > rho_liq — unphysical for VLE)
```

So this is a **stability false-positive**, not a phase-split problem.

### Why it is isolated / suspicious

The neighbouring compositions in the same series (x_methanol = 0.56, 0.58, 0.76,
0.78, 0.80) are all correctly classified as single-phase liquid — only **0.54**
is flagged unstable. An isolated false-positive at one composition points to the
Michelsen TPD stability search (`minimize_tpd` / `check_stability_michelsen`)
locating a spurious negative-TPD stationary point for this near-ideal but
associating (methanol) system — most likely a second liquid-like root being
found as the "vapour-like" trial.

### Reproduction

```python
import CoolProp.CoolProp as CP
AS = CP.AbstractState("HEOS", "methanol&benzene")
AS.set_mole_fractions([0.54, 0.46])
AS.update(CP.PT_INPUTS, 101325, 308.15)
print(AS.phase(), AS.Q(), AS.rhomolar())   # flagged two-phase pre-#3170
```

### Relationship to #3168 / #3170

Found during the #3168 phase-split convergence work. **PR #3170 makes the flash
recover correctly**: its phase-split convergence gate refuses to publish the
non-converging spurious split and falls back to a single-phase density solve, so
`update()` now returns the correct single-phase liquid (consistent with the
x = 0.56 neighbour). The wrong-answer symptom is therefore resolved by #3170 —
but the **underlying stability misclassification remains**: the stability
[...]

PR #3284, opened 2026-07-20
fix(flash): mixture PT flash returns spurious middle-branch root for subcooled liquid (#3283)

### LINKED ISSUE #3283: HEOS PT flash returns spurious middle-branch root for subcooled-liquid mixture (Methane/Ethane/Propane, knife-edge in T)
### Summary

The blind (no-envelope) HEOS **PT flash** returns a **spurious middle-branch root** for a
subcooled-liquid mixture at scattered "knife-edge" temperatures. For
**Methane & Ethane & Propane**, z = [0.05, 0.90, 0.05] (90 % ethane), at
**p = 2.0 MPa (20 bar)**, the bubble pressure at 223.15 K is ≈ 8.98 bar, so 20 bar is
**subcooled single-phase liquid** — the correct root is liquid (ρ ≈ 16 400 mol/m³).
Instead the flash returns ρ ≈ 6751 mol/m³ classified as *gas*, with a nonsensical
enthalpy (h ≈ −666 kJ/mol).

### Reproduction

```python
AS = CP.AbstractState("HEOS", "Methane&Ethane&Propane")
AS.set_mole_fractions([0.05, 0.90, 0.05])
AS.update(CP.PT_INPUTS, 2.0e6, 223.15)
print(AS.phase(), AS.rhomolar(), AS.hmolar()/1000)   # -> gas, 6750.9, -666.2   (WRONG)

# correct root:
AS2 = CP.AbstractState("HEOS", "Methane&Ethane&Propane")
AS2.set_mole_fractions([0.05, 0.90, 0.05]); AS2.specify_phase(CP.iphase_liquid)
AS2.update(CP.PT_INPUTS, 2.0e6, 223.15)
print(AS2.rhomolar(), AS2.hmolar()/1000)             # -> 16357.5, +3.20   (correct liquid)
```

### Observed vs. expected (p = 20 bar, z = [0.05, 0.90, 0.05])

| T [K] | ρ [mol/m³] | h [kJ/mol] | phase | |
|------:|-----------:|-----------:|:------|:--|
| 222.5 | 16389.7 | 3.15 | liquid | ok |
| **223.15** | **6750.9** | **−666.2** | **gas** | **spurious** |
| 225.0 | 16265.1 | 3.35 | liquid | ok |
| 212.5 | 6853.8 | −1940.7 | — | spurious |
| 235.0 | 3982.3 | −53.5 | gas | spurious |
| 248.15 / 193.15 | 14982 / 17729 | +5.3 / +0.9 | liquid | ok (controls) |

Deterministic on a **fresh** `AbstractState` (not warm-start dependent), and **scattered in T**
(222.5 and 225 K are fine; 223.15, 212.5, 235 K fail) — the fingerprint of a PT
density-solver root-selection / initial-guess problem, not a phase-envelope error. The
spurious ρ ≈ 6751 sits between the real liquid (~16400) and vapour (~1200) branches.

### Root cause

In `FlashRoutines::PT_flash_mixtures` → the blind-flash `solve_single_phase` lambda:

1. At 20 bar the SRK pre-screen has a **single** real root (~15500); `solver_rho_Tp_SRK`
   returns it for **both** the gas and liquid phase requests.
2. The code treats `gas_ok && liq_ok` as "two roots exist" and solves **both** HEOS
   branches. The `iphase_gas` branch is seeded from that liquid-like SRK density and, on
   the GERG mixture isotherm's spurious secondary loop at these temperatures, converges to
   ρ ≈ 6751 — a mechanically-stable (dp/dρ > 0) but thermodynamically-artifact root whose
   computed Gibbs energy (≈ −19039 J/mol) is *lower* than the true liquid's (≈ −1044 J/mol).
3. The Gibbs-minimising selection therefore picks the artifact root.

The stability test correctly reports **stable** here, so this is distinct from the
Michelsen stability false-positive of #3171 (that path already recovers via #3170; this one
does not).

### Environment

Confirmed on current `master` (`a540d910`) built from source, and on the released pip
`8.0.0b1` — i.e. **not** a stale checkout. `MIXTURE_STABILITY_ALGORITHM` = default.

andr1976:
## Summary

Fixes a wrong-root bug in the blind (no-envelope) HEOS **PT flash** for mixtures: at scattered "knife-edge" temperatures it returned a **spurious middle-branch density root** for a subcooled-liquid state.

Closes #3283.

Reproduction (Methane&Ethane&Propane, z=[0.05, 0.90, 0.05], p = 2.0 MPa):

```python
AS = CP.AbstractState("HEOS", "Methane&Ethane&Propane")
AS.set_mole_fractions([0.05, 0.90, 0.05])
AS.update(CP.PT_INPUTS, 2.0e6, 223.15)
print(AS.phase(), AS.rhomolar(), AS.hmolar()/1000)   # before: gas, 6750.9, -666.2   (WRONG)
                                                      # after:  liquid, 16357.5, +3.20
```

At 20 bar the bubble pressure is ≈ 8.98 bar, so the state is subcooled single-phase **liquid** (ρ ≈ 16 400). Before the fix, 223.15/212.5/235 K returned a bogus ρ ≈ 6751 "gas" root (h ≈ −666 kJ/mol); neighbouring 222.5/225 K were fine — the fingerprint of a density-solver root-selection problem.

## Root cause

In `FlashRoutines::PT_flash_mixtures` → `solve_single_phase`: at these conditions the SRK pre-screen has a **single** real root, and `solver_rho_Tp_SRK` returns that same value for **both** the gas and liquid phase requests. The code treated `gas_ok && liq_ok` as "two roots exist" and solved **both** HEOS branches. The gas branch, seeded from the liquid-like SRK density, converged onto a **secondary loop** of the GERG mixture isotherm at ρ ≈ 6751 — a mechanically-stable (dp/dρ > 0) but thermodynamically-artifact root whose computed Gibbs energy (≈ −19039 J/mol) is *lower* than the true liquid's (≈ −1044), so the Gibbs-minimising selection picked it. Knife-edge in T because that spurious loop only yields a convergeable root at scattered temperatures.

## Fix

Only run the two-branch Gibbs comparison when the SRK roots are **genuinely distinct** (`|ρ_srk_gas − ρ_srk_liq| > 1e-3·max(...)`, a real vapour+liquid pair). A single SRK root ⇒ single phase ⇒ solve only the branch whose density class matches the SRK seed (`ρ_srk > ρ_reducing` → liquid, else gas), with fallback to the other branch on throw. The `srk_distinct` two-branch block is unchanged from before.

Note: this also slightly changes the single-valid-root selection to classify by density (dense → liquid) rather than always preferring the "ok" phase — an improvement that matches the root's density class; the fallback-on-throw still protects a misclassified state.

## Relationship to #3171

Distinct. #3171 is a Michelsen stability **false-positive** (whose wrong-answer symptom is already handled by #3170's fallback); there the stability test is wrong. Here the stability test correctly reports **stable** — the bug is one layer downstream, in the single-phase root selection. Verified empirically: my fix does not change the #3171 case (still correct liquid) and does not alter the stability verdict.

## Validation

New regression test `[PTflash]` (`CoolProp-Tests-PTFlashMiddleRoot.cpp`) covers the reported knife-edge temperatures + passing-neighbour controls, comparing the blind flash against the imposed-liquid reference root. It **fails on unmodified master** (8/35 assertions) and **passes with the fix**. No regressions: `[mixture]` (24), `[flash]` (2097) and `[consistency]` (24991) assertions all pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved pressure–temperature flash calculations for certain subcooled mixtures.
  - Prevented selection of an incorrect intermediate-density root.
  - Improved liquid-phase density selection when multiple candidate solutions are available.
  - Increased reliability of liquid-state results in challenging mixture conditions.

- **Tests**
  - Added regression coverage for methane, ethane, and propane mixtures across affected temperature conditions.
  - Verified correct liquid-phase identification and density results.
  - Expanded automated test builds to include the new regression coverage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2374,6 +2374,8 @@ if(COOLPROP_CATCH_MODULE)
   list(APPEND APP_SOURCES
        "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests-PXFlash.cpp")
   list(APPEND APP_SOURCES
+       "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests-PTFlashMiddleRoot.cpp")
+  list(APPEND APP_SOURCES
        "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests-CubicAlpha.cpp")
   list(APPEND APP_SOURCES
        "${CMAKE_CURRENT_SOURCE_DIR}/src/Tests/CoolProp-Tests-CubicU.cpp")
--- a/src/Backends/Helmholtz/FlashRoutines.cpp
+++ b/src/Backends/Helmholtz/FlashRoutines.cpp
@@ -126,9 +126,18 @@ void FlashRoutines::PT_flash_mixtures(HelmholtzEOSMixtureBackend& HEOS) {
             double rho_srk_liq = HEOS.solver_rho_Tp_SRK(T_saved, p_saved, iphase_liquid);
             bool gas_ok = rho_srk_gas > 0 && ValidNumber(rho_srk_gas);
             bool liq_ok = rho_srk_liq > 0 && ValidNumber(rho_srk_liq);
+            // solver_rho_Tp_SRK returns the SAME single root for both phase requests when the SRK
+            // cubic has only one real root (a single-phase state).  Only Gibbs-compare two HEOS
+            // branches when the SRK roots are genuinely DISTINCT (a real vapor+liquid pair).
+            // Otherwise the "gas" HEOS solve is seeded from a liquid-like SRK density and, on a
+            // pathological multiparameter mixture isotherm, can converge to a spurious middle root
+            // whose unphysical Gibbs energy is lower than the true liquid root's, so the Gibbs test
+            // selects it (GH #3283: Methane/Ethane/Propane subcooled liquid at 20 bar returning a
+            // bogus rho~6751 "gas" root at knife-edge temperatures 223.15/212.5/235 K).
+            bool srk_distinct = gas_ok && liq_ok && std::abs(rho_srk_gas - rho_srk_liq) > 1e-3 * std::max(rho_srk_gas, rho_srk_liq);
 
-            if (gas_ok && liq_ok) {
-                // Both SRK roots valid — solve both HEOS roots, pick lower Gibbs
+            if (srk_distinct) {
+                // Both SRK roots valid and distinct — solve both HEOS roots, pick lower Gibbs
                 double rho_gas = -1, rho_liq = -1;
                 HEOS.specify_phase(iphase_gas);
                 try {
@@ -152,9 +161,13 @@ void FlashRoutines::PT_flash_mixtures(HelmholtzEOSMixtureBackend& HEOS) {
                     throw ValueError("Unable to obtain either HEOS density root in PT_flash_mixtures");
                 }
             } else {
-                // Only one SRK root valid (or neither) — solve that branch,
-                // fall back to the other if HEOS throws.
-                phases primary = gas_ok ? iphase_gas : (liq_ok ? iphase_liquid : iphase_gas);
+                // A single SRK root (both requests returned the same value), or only one/none
+                // valid => a single phase.  Solve the branch whose density class matches the SRK
+                // seed (dense => liquid, else gas); this seeds solver_rho_Tp on the correct side
+                // and avoids converging to the spurious middle root.  Fall back to the other
+                // branch if HEOS throws.
+                double rho_srk = liq_ok ? rho_srk_liq : rho_srk_gas;
+                phases primary = (gas_ok || liq_ok) ? ((rho_srk > HEOS.rhomolar_reducing()) ? iphase_liquid : iphase_gas) : iphase_gas;
                 phases fallback = (primary == iphase_gas) ? iphase_liquid : iphase_gas;
                 HEOS.specify_phase(primary);
                 try {
--- a/src/Tests/CoolProp-Tests-PTFlashMiddleRoot.cpp
+++ b/src/Tests/CoolProp-Tests-PTFlashMiddleRoot.cpp
@@ -1,0 +1,64 @@
+// Focused regression test for the blind (no-envelope) mixture PT flash returning a
+// spurious middle-branch density root (GitHub #3283).
+//
+// Runs in the default suite (tag [PTflash], NOT [.]-hidden).  Run explicitly:
+//   ./CatchTestRunner "[PTflash]"
+//
+// Background (#3283): for Methane&Ethane&Propane at z=[0.05,0.90,0.05], p=2.0 MPa the
+// state is subcooled single-phase LIQUID (bubble p ~ 8.98 bar << 20 bar).  At scattered
+// "knife-edge" temperatures the flash returned a spurious rho ~ 6751 mol/m^3 classified
+// as gas (h ~ -666 kJ/mol) instead of the liquid root (rho ~ 16400, h ~ +3 kJ/mol).
+//
+// Root cause: in FlashRoutines::PT_flash_mixtures the SRK pre-screen has a single real
+// root at these conditions (solver_rho_Tp_SRK returns the same value for both phase
+// requests), but solve_single_phase treated gas_ok && liq_ok as "two roots", solved a
+// spurious gas HEOS branch seeded from the liquid-like SRK density (converging onto a
+// secondary loop of the GERG isotherm), and the Gibbs-minimising selection picked that
+// artifact root because its unphysical Gibbs energy was lower than the true liquid's.
+// The fix only Gibbs-compares when the SRK roots are genuinely distinct.
+
+#if defined(ENABLE_CATCH)
+
+#    include <catch2/catch_all.hpp>
+
+#    include <memory>
+#    include <string>
+#    include <vector>
+
+#    include "CoolProp/AbstractState.h"
+#    include "CoolProp/DataStructures.h"
+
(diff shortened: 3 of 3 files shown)


ibell:
Thanks for this — I reviewed it against current `master`.

The premise checks out. I confirmed in `solver_rho_Tp_SRK` that when the cubic has one real root (`Nsolns == 1`, and likewise when only one root is positive) the function returns before the `switch (phase)`, so both phase requests genuinely do get the same value. So `srk_distinct` is a valid detector for the condition you describe, and the fix does close the reported path.

Two things I'd like to see addressed, and one process request.

**1. Classifying the SRK root against `rhomolar_reducing()` mixes two equations of state**

```cpp
phases primary = (gas_ok || liq_ok) ? ((rho_srk > HEOS.rhomolar_reducing()) ? iphase_liquid : iphase_gas) : iphase_gas;
```

`rho_srk` comes from the SRK cubic; `rhomolar_reducing()` is the multiparameter/GERG reducing density. SRK systematically underpredicts liquid density, so there is a band near the critical region where the SRK root falls below `rhomolar_reducing()` while the true HEOS root is liquid. In that case `primary` becomes `iphase_gas`, `solver_rho_Tp` is seeded on the gas side, and it can converge onto exactly the middle-branch root this PR is removing — and unlike the `srk_distinct` branch, there is no Gibbs cross-check here to catch it, so it gets published directly.

The reported #3283 temperatures sit far above the threshold, so the new test won't exercise this. Could the branch be chosen from something intrinsic to the SRK solve (e.g. which root of the cubic it was, or comparing against the SRK covolume `b`) rather than against a reducing density from a different EOS?

**2. The root cause is a missing mechanical-stability check, and the `srk_distinct` path still lacks one**

What actually went wrong in #3283 is that a mechanically unstable root was allowed to win a Gibbs comparison. This PR blocks the path where the two SRK roots coincide, but leaves the comparison itself unguarded:

```cpp
rho = (G_liq <= G_gas) ? rho_liq : rho_gas;
```

At a state where the SRK cubic *does* have two distinct roots but the HEOS gas solve converges onto a secondary branch, that line still selects the artifact whenever its unphysical Gibbs energy is lower — the same #3283 signature reached through the other branch. A `dp/drho > 0` check on each candidate before the Gibbs comparison would test the actual invariant instead of using SRK root multiplicity as a proxy for it, and would make the `srk_distinct` split less load-bearing.

I'm happy to take the proxy fix now and the stability guard as a follow-up if you'd rather keep this PR tight — but I'd like the invariant written down somewhere either way.

**3. Please rebase**

The branch is 66 commits behind `master` and the only checks that have run here are CodeQL, CodeRabbit and the CLA — the build/test matrix never ran on this PR. `CoolProp-Tests-PTFlashMiddleRoot.cpp` has therefore never been compiled or executed in CI, including its hard-coded reference bounds:

```cpp
REQUIRE(rho_liquid > 14000.0);
REQUIRE(rho_liquid < 19000.0);
```

It does merge-test clean against current `master`, so a rebase should be uneventful — I'd just like to see the full matrix green before merging.

The test itself is nicely scoped: using an imposed-liquid reference rather than a hard-coded density, and including the passing neighbours as controls, is the right shape.


---

andr1976:
Thanks @ianhbell — rebased onto current master and addressed all three.

**1. Intrinsic SRK classification.** The single-root branch now classifies `rho_srk` against SRK's *own* critical density instead of `rhomolar_reducing()`. SRK has `Z_c = 1/3` and `b = Ω_b·R·Tc/pc`, so `rho_c,SRK = 1/v_c = 3·Ω_b/b`, computed from `SRK_covolume()`. No cross-EOS comparison anymore, so the near-critical band you flagged — SRK root below the GERG reducing density while the true HEOS root is liquid — is closed: a subcritical single SRK root sits on SRK's own liquid branch (above its own critical density) and seeds the liquid side. (`b` is built with the same `Ω_b`, so the coefficient cancels and the threshold reduces to `3/Σ xᵢ R Tcᵢ/pcᵢ` — self-consistent by construction.)

**2. Mechanical-stability invariant.** I probed the #3283 isotherm to settle whether the artifact is mechanically unstable — it isn't. At the spurious root (Methane/Ethane/Propane, 2 MPa, 223.15 K, ρ≈6760) `dp/dρ ≈ +2.5e5 > 0`, mechanically **stable**, sitting just above a negative-pressure secondary loop; the true liquid root (ρ≈16360) is also `dp/dρ > 0`. So a plain `dp/dρ > 0` guard can't separate them — it would be false comfort and could reject correct roots. I've written the invariant *and* this finding into a comment at the `srk_distinct` Gibbs comparison, and tracked a follow-up for a stronger guard: reject a candidate adjacent to a negative-pressure segment of the isotherm, or select the lowest-Gibbs mechanically-valid root from a spinodal-bracketed global scan (`solver_rho_Tp_global`-style). Note #3283 itself takes the single-root branch (SRK returns one root, `srk_distinct == false`), so the distinct-root path has no concrete reproducer yet — happy to open a GitHub issue for the follow-up if you'd like it visible on the tracker.

**3. Rebased** onto current master (was 49 behind); the fix replays cleanly, so the full matrix now runs. `CoolProp-Tests-PTFlashMiddleRoot.cpp` compiles and executes — its reference bounds (`rho_liquid ∈ (14000, 19000)`) hold across every GENERATE'd temperature. Local preflight: clang-format/build/tests/semgrep green; `[PTflash]` (35 assertions), `[flash]` (4127), `[mixture]` (2054) all pass; cppcheck/clang-tidy report only pre-existing whole-file findings (none on this diff).

On the near-critical band in point 1: I don't have a concrete reproducer to pin as a regression test — the fix is principled (SRK-intrinsic) and the existing #3283 test guards against regression. Happy to add a near-critical test if you can point at a case, or hunt one down as a follow-up.

PR #3292, opened 2026-07-27
[split into #3294–#3298] Incompressible fluids: Tbase fix, Chebyshev caloric fits, GUI fluid-add, runtime registration

jowr:
**Superseded — this PR has been split into five reviewable PRs. Please review those instead; this one is kept open only as a tracking issue and should not be merged.**

At 154 files this was too large to review (CodeRabbit skipped it outright at its 150-file cap, and the single biggest commit was 126 generated data files). The work is unchanged — the tip of the stack is byte-identical to this branch apart from one unrelated `.coderabbit.yaml` tweak that was dropped as out of scope.

### The stack

Each PR is based on the previous one, so they should be reviewed and merged in order.

| PR | Contents | Files | Hand-written |
|---|---|---|---|
| #3294 | Exact `T == Tbase` derivative; stop shipping unfitted placeholder coefficients | 27 | 11 |
| #3295 | Simplify the fitting pipeline; audit the reference data | 23 | 23 |
| #3296 | Chebyshev caloric fitter + generated `json/` data | 134 | **7** |
| #3297 | C++ Chebyshev backend + runtime INCOMP registration | 6 | 6 |
| #3298 | Desktop GUI "add incompressible fluid" | 15 | 15 |

#3294 is independent and can merge on its own.

### Why this grouping

Two orderings are forced rather than chosen:

* The `IncompressibleFluid` move-semantics fix repairs `std::move` calls that only exist once the runtime-add refactor has landed, so it belongs with #3297 — not with the early correctness fixes.
* The `LBE` Chebyshev backfill *is* `_cheb` data, so it cannot precede the Chebyshev feature; it lives in #3296.

The generated `json/` data is isolated in #3296 because it accounts for 126 of this PR's files while containing no reviewable logic. Verified that it can safely land before its C++ consumer: there is no schema validation on incompressible JSON and unknown keys are never looked up, so the `*_cheb` entries are inert until #3297 — and #3296's branch builds and passes the full incompressible suite with the data present and no support for it.

### Review history preserved here

The adversarial review that ran against this branch (16 actionable findings + 11 nitpicks) is in the comments above, and every finding is fixed in the stack. One was declined on the merits and documented rather than patched: synchronizing `add_one` with a write-only lock would be worse than none, since every reader walks the same containers unlocked and would keep racing while appearing protected — see #3297.


jowr:
@coderabbitai review

---

PR #3293, opened 2026-07-27
fix(beads): bootstrap bd in ephemeral containers via a single hardened SessionStart hook

jowr:
### Description of the Change

`dev/ci/bootstrap-beads.sh` is a new `SessionStart` hook that makes `bd` (the beads issue tracker CLAUDE.md mandates for all task tracking) usable in Claude Code web/CI containers. Those containers start from a fresh clone with no `bd` binary and no Dolt database (`.beads/embeddeddolt/` is gitignored), so the committed `bd prime` hook was previously silently no-op'ing for an entire session.

The script:
- Installs `bd` via `go install` through the allowlisted Go module proxy (the upstream `curl|bash` installer's GitHub-release download is blocked by the agent proxy).
- Rehydrates the embedded Dolt DB from the committed `.beads/issues.jsonl`.
- Runs `bd prime` itself as its last step, and is the **only** `SessionStart` hook for beads — Claude Code runs same-event hooks concurrently rather than in array order, so a separate `bd prime` hook entry would race a `bd` this script hasn't finished installing yet.
- Is **opt-in**, not auto-detected: the install+hydrate steps only run when `BEADS_BOOTSTRAP=1` is set in an environment's persistent config. The cold-start install+hydrate can take a couple of minutes, and that cost should be a deliberate per-environment choice, not sprung on every session in every CI/remote container. Without the flag, the hook just primes an already-installed `bd` (near-zero cost) and exits — never installs, never hydrates.
- Checks hydration status via a `bd count` health probe rather than directory presence, since `bd init` creates the DB directory *before* importing — directory presence alone would falsely latch "hydrated" on a partial/failed import.
- Guards the hydration critical section with an `flock(1)` lock: atomic by construction (no TOCTOU window), and self-releasing on any process exit — including `SIGKILL`, which no shell trap can catch — so there's no separate staleness/steal logic to get wrong. A session that loses the lock race skips hydration *and* `bd prime` entirely for that run, rather than risk reading the DB mid-rebuild.
- Restores, via an `EXIT`/`INT`/`TERM` trap, the few tracked files (`.beads/config.yaml`, `.beads/.gitignore`, `.gitignore`) that `bd init --stealth` still normalizes even in stealth mode — but only when the tree was clean beforehand, never clobbering a developer's in-progress edits.

CLAUDE.md and `dev/ci/README.md` document the actual behavior (opt-in gating, idempotency, the concurrency lock, and the one known gap below) rather than an aspirational version.

### Benefits

- `bd` — and the beads-based issue-tracking workflow CLAUDE.md requires for all task tracking — becomes usable in every environment that opts in, instead of silently unavailable.
- No environment pays the cold-start cost unless it explicitly asked for beads tracking.
- Idempotent and strictly non-fatal by design: a warm container skips both install and hydration; any failure logs a warning and the session continues without bd, the same fail-open posture as the existing git hooks.

### Possible Drawbacks

- First cold start in an opted-in environment takes ~2-3 minutes (Go toolchain fetch + build) before the hook completes; an explicit 600s hook timeout is set so this isn't killed prematurely.
- Adds a dependency on the Go toolchain and module proxy being reachable in opted-in environments; if that's ever blocked too, bootstrapping fails the same non-fatal way (warns, session continues without bd) rather than a new failure mode.
- No shell trap can catch `SIGKILL` or survive a hard container teardown. If a hydration is killed that way, `.beads/config.yaml`, `.beads/.gitignore`, and `.gitignore` can be left modified in the working tree, and the *next* session's dirty-check will read that as a developer's in-progress edit and — by design — not auto-restore it either. This is called out explicitly in CLAUDE.md as a known, accepted gap (fixed by a manual `git checkout -- .beads/config.yaml .beads/.gitignore .gitignore`) rather than left as an implied guarantee.
- Opt-in means someone has to remember to set `BEADS_BOOTSTRAP=1` for a given environment; unlike the earlier auto-detected design, a forgotten flag means bd silently stays unavailable there (the original problem this PR set out to fix) — an accepted trade-off in exchange for never surprising an environment with an unwanted multi-minute install.

### Verification Process

- Ran `./dev/ci/preflight.sh` (this diff is shell + Markdown/JSON only, so clang-format/cppcheck/clang-tidy/semgrep auto-skip). The two remaining checks that failed (`json-symbols`, `install-headers`) are pre-existing failures on a clean `origin/master` checkout in this sandbox — confirmed unrelated to this diff by reproducing them against master before any of these changes existed.
- Ran the mandatory pre-PR adversarial-review subagent (per CLAUDE.md) on every substantive commit in this PR, including the opt-in gate rewrite.
- Live-tested extensively in a sandbox that happens to already match the target environment (no `bd` on `PATH`, no `.beads/embeddeddolt/`), by actually installing `bd` and running the script, inspecting real `git status`/`git diff` output for each scenario:
  - Cold hydrate under `BEADS_BOOTSTRAP=1` — install + DB hydration, tree clean afterward.
  - Opt-out default — confirmed install+hydrate is skipped even with `CI=true`/`CLAUDE_CODE_REMOTE=true` set, unless `BEADS_BOOTSTRAP=1` is also set; `BEADS_BOOTSTRAP=0` does not enable it (exact-match check).
  - Warm rerun — both steps no-op.
  - Dirty-tree guard — a pre-existing uncommitted edit to `.gitignore` is preserved, not clobbered.
  - `bd init --from-jsonl` failure path (corrupt JSONL) — warns, retries next session, doesn't leave a poisoned DB.
  - A real bug found *by* this testing: `bd prime` lazily creates an empty Dolt DB as an undocumented side effect, which then made a later `bd init --from-jsonl` abort with "already initialized". Fixed by clearing `.beads/embeddeddolt` before every hydration attempt, not just after a failure.
  - SIGTERM sent mid-hydration — confirmed dash does *not* terminate on a trapped signal unless the handler calls `exit` itself (a real bug caught this way and fixed); the process now exits promptly and the tree is left clean.
  - A genuine TOCTOU race in an earlier hand-rolled mkdir+pid-file lock (found by CodeRabbit's review, verified by hand-tracing) — replaced with `flock(1)`, and re-tested: lock contention against a real concurrent flock holder is skipped cleanly with no `bd prime` race, and release is automatic on any exit.
  - A real bug found while fixing the above: a bare `exec 9>file 2>/dev/null` was silently redirecting the *rest of the script's* stderr to `/dev/null`, not just that one line's — caught because a diagnostic message went missing in testing.
  - Forced `git checkout` failure during restore — now logs a per-file warning instead of silently leaving the tree dirty with no diagnostic trace.

### Applicable Issues

N/A — no linked issue. This fixes a self-discovered gap: `bd` was unavailable for an entire prior session because of exactly the problems fixed here.

## Summary by CodeRabbit

* **New Features**
  * Improved SessionStart automation in ephemeral web/CI containers to make the `bd` CLI available, bootstrap/hydrate the embedded issue database from committed data, and run `bd prime`.
  * Added an explicit execution timeout and ensured bootstrapping is opt-in via `BEADS_BOOTSTRAP=1`, with idempotent, concurrency-safe behavior and graceful fallbacks.

* **Documentation**
  * Expanded CI/Web bootstrap documentation with the install/bootstrap/prime flow, enablement guidance, cleanup/troubleshooting notes, and issue sync expectations.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -15,6 +15,15 @@
       {
         "hooks": [
           {
+            "command": "sh dev/ci/bootstrap-beads.sh",
+            "type": "command"
+          }
+        ],
+        "matcher": ""
+      },
+      {
+        "hooks": [
+          {
             "command": "bd prime",
             "type": "command"
           }
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -15,17 +15,9 @@
       {
         "hooks": [
           {
-            "command": "sh dev/ci/bootstrap-beads.sh",
-            "type": "command"
-          }
-        ],
-        "matcher": ""
-      },
-      {
-        "hooks": [
-          {
-            "command": "bd prime",
-            "type": "command"
+            "command": "sh \"${CLAUDE_PROJECT_DIR}/dev/ci/bootstrap-beads.sh\"",
+            "type": "command",
+            "timeout": 600
           }
         ],
         "matcher": ""
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -49,6 +49,23 @@ bd close <id>         # Complete work
 - If push fails, resolve and retry until it succeeds
 <!-- END BEADS INTEGRATION -->
 
+### `bd` in ephemeral (Claude Code web/CI) containers
+
+Fresh containers have no `bd` binary and no Dolt DB (the DB dir
+`.beads/embeddeddolt/` is gitignored). `dev/ci/bootstrap-beads.sh` — wired
+as the first `SessionStart` hook in `.claude/settings.json`, ahead of the
+beads-managed `bd prime` — installs `bd` via `go install` (the `curl|bash`
+installer's GitHub-release download is blocked by the agent proxy; the Go
+module proxy is allowed) and rehydrates the DB from the committed
+`.beads/issues.jsonl`. It is idempotent and non-fatal. First cold start
+takes ~2 min (Go toolchain fetch + build); warm containers skip both steps.
+To persist newly filed issues, append their `bd export --include-memories`
+lines to `.beads/issues.jsonl` and commit — do **not** overwrite the file
+wholesale (bd 1.1.0 re-serializes the `dependencies` field on unrelated
+issues, and a plain `bd export` drops the persisted memories). `bd dolt
+push` in the session-completion steps above is a no-op here — there is no
+Dolt remote in web containers; committing `issues.jsonl` is the sync.
+
 
 ## Build & Test
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -61,11 +61,15 @@ GitHub-release download is blocked by the agent proxy; the Go module proxy is
 allowed), rehydrates the DB from the committed `.beads/issues.jsonl`, and
 finally runs `bd prime` itself once both are ready.
 
-It only bootstraps in recognized ephemeral environments (`CI=true` or
-`CLAUDE_CODE_REMOTE=true`) — a developer workstation with `go` on `PATH` is
-left alone rather than getting an unprompted `go install` / toolchain fetch.
-Set `BEADS_BOOTSTRAP_FORCE=1` to opt in anywhere else (e.g. to exercise the
-script locally). It is idempotent — hydration is checked with a `bd count`
+It only bootstraps when explicitly opted in with `BEADS_BOOTSTRAP=1` — set once
+in an environment's persistent config, not auto-detected from `CI`/
+`CLAUDE_CODE_REMOTE`, since the cold-start install+hydrate can take a couple of
+minutes and that cost should be a deliberate per-environment choice, not
+sprung on every session in every ephemeral container. Without it, the hook
+only primes an already-installed `bd` (near-zero cost) and exits — never
+installs, never hydrates; a developer workstation with `go` on `PATH` is
+naturally left alone unless someone sets the variable there too. It is
+idempotent — hydration is checked with a `bd count`
 health probe rather than directory presence, so a partial or failed import is
 retried next session instead of latching "done" forever — and non-fatal: any
 failure warns on stderr and the session continues without bd. An `flock(1)`
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -68,10 +68,15 @@ Set `BEADS_BOOTSTRAP_FORCE=1` to opt in anywhere else (e.g. to exercise the
 script locally). It is idempotent — hydration is checked with a `bd count`
 health probe rather than directory presence, so a partial or failed import is
 retried next session instead of latching "done" forever — and non-fatal: any
-failure warns on stderr and the session continues without bd. A single-shot
-mkdir lock (with dead-PID stealing, not a busy-wait) guards the hydration
-critical section against two SessionStart hooks racing in the same container.
-On a tree that was clean beforehand, it also restores the few tracked files
+failure warns on stderr and the session continues without bd. An `flock(1)`
+lock (not a busy-wait) guards the hydration critical section against two
+SessionStart hooks racing in the same container; a session that loses the
+race skips hydration *and* `bd prime` entirely for that run rather than
+risk reading the DB mid-rebuild, and retries from scratch next session.
+flock's kernel-level exclusivity is what makes the acquire itself atomic
+and self-releasing on any exit (including SIGKILL, which no trap can
+catch) — no separate staleness/steal logic needed. On a tree that was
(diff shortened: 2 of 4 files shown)


jowr:
Thanks for the review, @coderabbitai — went through each finding against the actual runtime behavior (installed `bd`, ran the script for real, forced the concurrency race with a background `flock` holder) rather than just reading the diff. Pushed `3870c3d` addressing:

- **Stale-lock steal race (Critical)**: confirmed real — two processes could interleave `rm -rf`+`mkdir` such that one deletes the other's freshly-created lock. Replaced the whole mkdir+pid-file scheme with `flock(1)` instead of patching the steal logic: the kernel makes the acquire atomic and auto-releases the lock on any exit (including `SIGKILL`, which no shell trap can catch), so there's no staleness/steal logic left to get wrong.
- **`bd prime` racing a concurrent hydration**: agreed — a session that loses the lock now skips `bd prime` entirely and exits, instead of calling it against a DB that might be mid-rebuild.
- **`BEADS_BOOTSTRAP_FORCE` accepting any non-empty value**: fixed, now requires exactly `"1"` as documented.
- **Silent `hash -r` / inconsistent `bd prime` failure handling**: both now log a warning on failure, matching every other soft-failure path in the script.
- **One claim didn't hold up**: "the bootstrap path returns `bd prime`'s exit status" — it doesn't; both call sites already end in an unconditional `exit 0`. Kept the added failure-visibility anyway since it's useful for debugging, just not framed as an exit-code fix.

Also found one bug during live testing that wasn't in the review: a bare `exec 9>lockfile 2>/dev/null` silently redirected the *rest of the script's* stderr to `/dev/null` (bare `exec` applies redirects to the current shell persistently, not just that statement) — fixed, and reverified the lock-contention warning actually reaches stderr now.

To be clear on scope/intent: this hook only matters for AI-agent sessions bootstrapping `bd` in ephemeral containers — happy to adjust further or drop pieces of this if anyone has a different take on the approach, it's not something we need to force through as-is.

---

jowr:
Closing in favour of #3392, which carries this work forward on a branch that can actually be merged.

**Why this branch cannot be merged.** `jowr/install_beads` has *no merge base with `master`*:

```
jowr/install_beads   root b3847c75
origin/master        root 2694282f   (2026-08-15)
git merge-base       -> NONE
```

`master` was re-rooted on 2026-08-15, three weeks after this PR was opened. GitHub still reports `mergeable_state: "clean"`, which is misleading. Merging would need `--allow-unrelated-histories` and a conflict across the whole tree. Around 150 remote branches are still on the old root, so this is unlikely to be the only PR affected — worth checking the other pre-August ones before anyone tries to merge them.

**Why the approach changed too.** Two things came out of actually running this script in a target container rather than reading it:

1. **The `go install` path does not work here, at all.** A cgo build needs the ICU development headers `github.com/dolthub/go-icu-regex` compiles against, and these images ship libicu but not libicu-dev — it dies minutes in with `fatal error: unicode/uregex.h: No such file or directory`. Switching to `CGO_ENABLED=0` builds fine and then refuses to run: *"embedded Dolt requires a CGO build, but this bd binary was built with CGO_ENABLED=0"* — and `.beads/metadata.json` selects embedded mode. So even with `BEADS_BOOTSTRAP=1` set, this would have installed nothing. The npm package ships a prebuilt CGO-enabled binary from an allowlisted registry and installs in ~3 s.

2. **Opt-in only moves the problem.** It forces a choice between paying the cold-start cost in every session — including the many that never open the tracker — and not having `bd` at all. This PR's own Drawbacks section concedes the second case is *"the original problem this PR set out to fix"*. #3392 pays on first use instead: the SessionStart hook drops to ~9 ms and only the session that actually runs a `bd` command waits.

**What carried over.** All the hardening from your review with @coderabbitai survives: the `flock` that replaced the hand-rolled stale-lock protocol, the `bd count` health probe rather than directory presence, the trap restoring the tracked files `bd init --stealth` normalizes but only on an already-clean tree, and the strictly non-fatal posture. The recorded learning — non-fatal, exact-value gating, never `bd prime` against a rebuilding database — all still holds, and #3392 tightens the last one: readiness is now a marker written only after the import is verified, because `bd prime` creates an empty database as a side effect and directory presence latched that as "ready" forever.

One finding worth flagging back here, since it applies to the design in this PR as well: the git-hook stand-down keyed on `GIT_DIR`/`GIT_INDEX_FILE` does not work. Measured on git 2.43, `GIT_DIR` is not exported to hooks at all, and `GIT_INDEX_FILE` is set only for the index-touching ones — `post-checkout`, `post-merge` and `pre-push` get neither. `.beads/hooks/*` already export `BD_GIT_HOOK`, which is the reliable signal.

---

Issue #2488, opened 2025-03-01
Incompressible fitting code is no longer working


ibell:
Related to my efforts to debug #2447 , I found that something (and I can't figure out what) has changed in the incompressible fitting code. When the script is run, a lot of the models are changed/broken. Notably, viscosity and vapor pressure seem to always end up with messed up logexponential models, and I don't know why. 

I tried:

* To roll numpy back to version 1.24.0 thinking that the numpy might be too new. That didn't help. I noted that the ``np.NAN`` has been deprecated in numpy 2.0
* To debug into the fitting code, and walk through the vapor pressure and viscosity fitting code. Nothing jumped out at me, but there was some exception about the shape of the arrays.

@jowr any idea?

ibell:
Nevermind, I tracked it down

Issue #1331, opened 2016-11-13
Calculation BUG ASK FOR HELP


learnweierstrass:
I want to get the viscosity and thermal conductivity of LiBr aq with CoolProp
In Matlab I enter
 
CoolProp.PropsSI('V','T',373.15,'Q',0,'INCOMP::LiBr[0.55]') 
Then respond shows 
Ans= 
     1
 
Which is not consistent with the thermal physical data in the textbook(0.0015). 
Also 
 
CoolProp.PropsSI('L','T',373.15,'Q',0,'INCOMP::LiBr[0.55]')
Ans=
     0
 
Would you help me solve these questions, Thank you very much

ibell:
@jowr - any idea?

ibell:
This is clearly a bug somewhere.  Have you tried giving TP as inputs instead of TQ?

learnweierstrass:
But I can’t , If we just know the the equilibrium condition T and mass fraction, this will return a wrong value

 

发件人: bounces+848413-0dbf-learnweierstrass=163.com@sgmail.github.com [mailto:bounces+848413-0dbf-learnweierstrass=163.com@sgmail.github.com] 代表 Ian Bell
发送时间: 2016年11月12日 21:46
收件人: CoolProp/CoolProp <CoolProp@noreply.github.com>
抄送: learnweierstrass <learnweierstrass@163.com>; Author <author@noreply.github.com>
主题: Re: [CoolProp/CoolProp] Calculation BUG ASK FOR HELP (#1331)

 

This is clearly a bug somewhere. Have you tried giving TP as inputs instead of TQ?

ibell:
Definitely a bug...  I can replicate in python

learnweierstrass:
how can I solve this Bug dear

ibell:
I don't know right now, that's the whole point of it being a bug !  We have
to wait for the incompressible developer to chime in.

learnweierstrass:
thanks, keep in touch!

ibell:
The solid line at mu = 1 here is clearly LiBr:
http://www.coolprop.org/dev/_downloads/incompressibles_consistency.pdf

learnweierstrass:
Yes， definitely, I use experimental data to obtain a new MATLAB program, which can calculate the thermal conductivity. just do not use the 
CoolProp.PropsSI('L','T',373.15,'Q',0,'INCOMP::LiBr[0.55]')
in MATLAB. Phthon is running well

[...]

PR #3295, opened 2026-07-28
incompressible 2/5: simplify the fitting pipeline and audit the reference data

jowr:
Second of five PRs splitting up #3292. **1/5 (#3294) is merged, so this now targets `master` directly** — the branch was rebased onto it and this PR's diff is only its own 22 files.

Developer tooling only: nothing here changes the shipped library or any committed coefficient.

### Requirements

### Description of the Change

**Make the fitting pipeline approachable.** `dev/incompressible_liquids/` had accumulated ~15 years of dead Python-2 code and hard dependencies that made adding a fluid harder than the maths warrants. This removes the dead code, makes `matplotlib` and the built CoolProp wheel genuinely optional so fitting needs only `numpy` + `scipy`, splits `requirements.txt` into core / reports / dev so the install instructions match the file, and adds a step-by-step `README.md` for adding a new fluid.

**Audit every raw reference-data file** (`DATA_AUDIT.md`) for physical plausibility — no wrong-unit slips were found, and one real ordering bug was: the HFE-7100 tables ship with *descending* temperature (+64 down to −80 °C). Loaders now sort each grid by its own axis values rather than trusting file order, which also covers files that are shuffled rather than exactly reversed.

Also folds in the review findings on this area: the optimizer starting guesses are centralized on `IncompressibleData` (they were restated in three places, and a copy that drifted would stop recognising a failed fit — reopening the #1331/#2567 bug that 1/5 just fixed), `writeCsvTableToFile` no longer crashes on a bare relative path, and the golden-master regression test no longer passes silently when a fit failed.

**The tests are now actually run.** Both gates previously named `test_json_sanity.py` explicitly, so `test_data_sanity.py` (the grid-ordering contract) and `test_fitting_regression.py` (the golden master) — the two safeguards this PR adds — never executed anywhere. CI now runs the whole directory with `uvx --with numpy --with scipy`, and preflight's `incomp-sanity` check does the same; its stdlib-only fallback stays on `test_json_sanity.py`, the one module needing neither dependency.

### Benefits

* A contributor can add an incompressible fluid with `pip install -r requirements.txt` (numpy + scipy) and a documented recipe, without touching pipeline internals.
* The HFE-7100 grids are now loaded in the order the fitter assumes.
* The unfitted-placeholder guard from 1/5 can no longer be silently disarmed by a drifting copy of the starting-guess literals — and the check that enforces that is no longer tautological (see the rebase note).

### Possible Drawbacks

* Deletes a lot of long-dead code. It is dead (Python-2 syntax that cannot import on any supported interpreter), but it is a large deletion for reviewers to skim.
* The `incompressible JSON sanity` CI job now installs numpy and scipy, so it is slower than the stdlib-only version and depends on those resolving.

### Verification Process

* `pytest` on `dev/incompressible_liquids/` — 13 tests: data-file sanity (axis ordering across every SecCool grid, sentinel handling), JSON loader-contract and placeholder sanity, and a golden-master refit regression for three representative fluids (DowJ, LiBr, AKF) pinned against the committed JSON. Confirmed the same 13 run under CI's exact invocation.
* `test_data_sanity.py` asserts the HFE-7100 grids load ascending, which fails against the pre-fix loader. Shuffling a real grid's rows reproduces the sorted result exactly, so the fix is a sort by value and not a reversal.
* The golden-master test now **asserts** rather than skips when a property committed to `json/` comes back unfitted — previously a swallowed `TypeError`/`ValueError` in `fitFluidList` reported a green run for exactly the crash the test exists to catch. It covers all six fitted properties, including the exponential-family `saturation_pressure`/`T_freeze` fits that #2488 was about.
* The drift check between the stdlib fallback literals and the fitter's real starting guesses is verified to fail on a corrupted literal *and* on a guess constant added to the fitter but not to the literals.
* `matplotlib` genuinely optional: verified by running the full pipeline with it uninstalled — all 117 fluids plus 6 examples fitted and written.

**On regenerating `json/`:** this PR changes no committed coefficient, but "regeneration is byte-identical" would be too strong a claim to make. `writeFluidList` short-circuits on `CPIncomp/data/hashes.json`, which is **untracked** — so a regeneration run on a warm cache reports every fluid unchanged and proves nothing. With the cache removed, the current writer emits full-precision `repr` values where the committed files carry Python-2 `%e` formatting (`SolutionData.round()` exists but has no callers), so the files differ in formatting regardless of this branch. Byte-identity would need that formatting question settled first, which is out of scope here.

### Note on the rebase

1/5 was **squash-merged**, so none of its commits are ancestors of `master` and this branch's shared history was orphaned. It was rebased with `git rebase --onto origin/master f9d96cdf`, dropping the 8 now-redundant 1/5 commits and replaying this PR's own 8 on top of master. One conflict, in `writeCsvTableToFile`: 1/5 had added a docstring where this branch replaced the inline `os.makedirs` with `ensure_parent_directory()`. Resolved as both.

`test_json_sanity.py` merged with **no** conflict, which turned out to be the risky part. This branch's fitter-sourced `KNOWN_GUESSES` and 1/5's loader-contract checks combined into something that looked right and was not: the drift check compared the fitter to itself (so a corrupted literal passed), and it imported `pytest`, which broke the pytest-free preflight path 1/5 had added after this branch forked. Both are fixed, with the fitter side now enumerated dynamically so a *new* guess constant is caught too.

### Applicable Issues

Refs #2488 (the unpinned-numpy breakage that motivated pinning + the regression test).

### Review notes

Two earlier CodeRabbit threads were assessed and need no change:

* **`CPIncomp/PureFluids.py` vs `CoefficientFluids.py`** — the finding's premise is incorrect. Both modules exist and `CPIncomp/__init__.py` instantiates them separately via `getPureFluids()` and `getCoefficientFluids()`. `PureFluids.py` is the main pure-fluid module (734 lines, 26 classes: LBE, LiquidSodium, TherminolD12, …); `CoefficientFluids.py` holds a different category (229 lines, 8 classes: NitrateSalt, FoodProtein, FoodFat). The documentation pointing contributors at `PureFluids.py` is correct, and applying the suggestion would send them to the wrong module.
* **MD038 in `NOTES_mixing_models.md`** — does not reproduce; the only code span in that region has no internal whitespace. `markdownlint` is also not a repo gate (no reference in `.github/` or `.pre-commit-config.yaml`).

On the **`numpy>=2.4.6` / `scipy>=1.17.1` lower bounds**: left as ranges deliberately. The floors are recent, and the golden-master regression test — which now actually runs in CI — is the designed detector for a fitting change from a dependency upgrade. Upper bounds would make the pipeline uninstallable on otherwise-working environments in exchange for a risk that is already covered by a test.

The **"Docstring Coverage"** pre-merge warning measures every function in the changed files, not the ones this PR touches, so it cannot be cleared without documenting a large amount of untouched pipeline code. It is a warning, not a gate.

### Known, out of scope

`json/IceEA.json`, `IceNA.json` and `IcePG.json` ship conductivity and viscosity coefficients the pipeline can no longer reproduce — `SecCoolFluids.py` has those data loads commented out, so a refit clears both properties. This is the same class as the 1/5 placeholder bug, but no guard catches it because the values are neither zero nor optimizer guesses. Not introduced here (master behaves identically); worth its own issue and data fix.

---

## Summary by CodeRabbit

* **Documentation**
  * Expanded guidance for incompressible-mixture thermodynamics, reference temperatures, and entropy calculations.
  * Added fitting-pipeline instructions, data-audit notes, and research documentation on mixture modeling.
  * Clarified how to add pure fluids and solution mixtures.

* **Improvements**
  * Improved handling of unsorted property grids and generated output directories.
  * Example datasets now work when optional CoolProp support is unavailable.
  * Added more consistent fitting behavior and validation of generated data.

* **Tests**
  * Added broad sanity and regression coverage for fluid data, grids, and fitted coefficients.

### DIFF
--- a/.github/workflows/dev_checks.yml
+++ b/.github/workflows/dev_checks.yml
@@ -807,6 +807,11 @@ jobs:
     # twin: the `incomp-sanity` check in dev/ci/preflight.sh.
     if: github.event_name == 'pull_request' || github.event_name == 'push'
     runs-on: ubuntu-latest
+    # Least privilege: this job runs PR-authored test code, and needs nothing
+    # beyond reading the tree. Without an explicit block it would inherit the
+    # repository default, which is broader.
+    permissions:
+      contents: read
     steps:
       - uses: actions/checkout@v7
 
--- a/.github/workflows/dev_checks.yml
+++ b/.github/workflows/dev_checks.yml
@@ -822,7 +822,12 @@ jobs:
       # uvx pyright@1.1.409): unpinned, a fresh pytest release could turn this
       # required check red on unrelated PRs.
       - name: Run incompressible JSON sanity checks
-        run: uvx pytest@9.1.1 dev/incompressible_liquids/test_json_sanity.py -q --color=no
+        # The whole directory, with numpy+scipy: test_data_sanity (grid axis
+        # ordering) and test_fitting_regression (golden-master refit) are the two
+        # safeguards this area relies on, and both importorskip away without
+        # them -- so naming only test_json_sanity.py meant they never ran here.
+        # `uvx --with` matches the pyright job above.
+        run: uvx --with numpy --with scipy pytest@9.1.1 dev/incompressible_liquids/ -q --color=no
 
   installed-header-gate:
     name: Installed-header hygiene gate
--- a/Web/fluid_properties/Incompressibles.rst
+++ b/Web/fluid_properties/Incompressibles.rst
@@ -231,6 +231,26 @@ compositions as independent fluids. This should be kept in mind when comparing
 properties for different compositions. Setting the reference state for one
 composition will always affect all fluids consisting of the same components.
 
+.. warning::
+   Enthalpy and entropy for incompressible mixtures are obtained purely by
+   integrating the fitted heat capacity :math:`c(T,x)` in temperature; there is
+   **no enthalpy or entropy of mixing term** anywhere in this calculation. Two
+   states at different compositions that happen to be queried at the same
+   :math:`T` and :math:`p` do **not** have enthalpies that are comparable across
+   that composition change, and a mixing/energy balance across streams of
+   different concentration (for example, an absorption chiller's generator or
+   absorber) will silently omit the heat of mixing. This is a known,
+   long-standing limitation, not a bug to be patched incrementally — see
+   `#533 <https://github.com/CoolProp/CoolProp/issues/533>`_,
+   `#781 <https://github.com/CoolProp/CoolProp/issues/781>`_, and
+   `#1690 <https://github.com/CoolProp/CoolProp/issues/1690>`_. A proper fix
+   needs a composition-dependent excess-property formulation (e.g. a
+   Gibbs-energy-based model, as used for some published LiBr/H2O and NH3/H2O
+   correlations) rather than an incremental change to the current
+   heat-capacity-integration approach; see
+   ``dev/incompressible_liquids/NOTES_mixing_models.md`` for research notes
+   toward that future work.
+
 .. The approach described in textbooks like Cengel and Boles :cite:`Cengel2007`
 .. is that the internal energy :math:`u` only depends on temperature and does not
 .. change with pressure.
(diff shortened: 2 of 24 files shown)


jowr:
@coderabbitai review

---

PR #3296, opened 2026-07-28
incompressible 3/5: Chebyshev caloric fits — fitter and generated json/ data

jowr:
Third of five PRs splitting up #3292. **2/5 (#3295) is merged, so this now targets `master` directly** — the branch was rebased onto it and this PR&#39;s diff is only its own 6 commits.

⚠️ **This PR is 134 files, but only 7 are hand-written.** The other 127 are generated `dev/incompressible_liquids/json/*.json` coefficient files. Review the 7 code/doc files and the tests; the data is mechanically produced and re-running the generator is a verified no-op.

### Requirements

* All new code requires tests to ensure against regressions.

### Description of the Change

Adds a Chebyshev-basis fit (`density_cheb` / `specific_heat_cheb`) alongside the existing centred-polynomial fit for every fluid&#39;s density and heat capacity, and ships the generated entries.

**Why a second basis.** The current backend integrates the centred-monomial cp fit analytically: `h = ∫cp dT` and `s = ∫cp/T dT`, the latter via a binomial expansion with one `ln T` term. Both are exact, but the entropy path is numerically fragile at high order and is the same construction that bred the `T == Tbase` singularity class fixed in 1/5. In a Chebyshev basis the antiderivative of a series is another series (stable recurrence, no centring, no singularity), so `dh/dT == cp` and `ds/dT == cp/T` hold **by construction**. Entropy comes from re-expanding `cp(T)/T` at Chebyshev–Lobatto nodes, which converges geometrically because `T &gt; 0` on the whole domain. Fit degree is chosen by generalized cross-validation.

`NOTES_thermodynamic_consistency.md` records the assessment that motivated this, including the two independent entropy routes cross-checked against adaptive quadrature.

**Nothing consumes these entries yet** — that is 4/5. They are additive and inert here: the C++ loader in this PR knows nothing about `*_cheb`, there is no schema validation on incompressible JSON, and unknown keys are never looked up. That ordering is deliberate and verified (see below), so this PR is safe to land before the backend.

Two fluids move from a tabular refit to the exact basis conversion (ZMC, ZS55): their raw grids do not cover the advertised temperature range, so a refit would have shipped an unguarded extrapolation. LBE — added to master in #3193 after this work began — is backfilled so every fluid has the same coverage.

### Benefits

* An exact, singularity-free enthalpy **and** entropy integral becomes available, replacing an approximation-prone construction.
[...]

PR #3297, opened 2026-07-28
incompressible 4/5: C++ Chebyshev caloric backend and runtime INCOMP registration (#2384)

### LINKED ISSUE #2384: Adding incompressible fluids during runtime (Python)
I'm currently adding the new incompressible fluids as instructed in http://www.coolprop.org/fluid_properties/Incompressibles.html#adding-new-fluids which is rather tedious if I wanted to share the build (not impossible but would require packing another dependency and simultaneously not supporting quick experimentation with runtime additions). As Python wrapper HEOS backend supports runtime addition with add_fluids_as_JSON, I find it somewhat odd that same kind of functionality is not implemented for incompressible, unless I'm mistaken of course.

jowr:
Fourth of five PRs splitting up #3292. **3/5 (#3296) is merged, so this now targets `master` directly** — the branch was rebased onto it and this PR&#39;s diff is only its own 6 files.

### Requirements

### Description of the Change

The C++ consumer for the Chebyshev entries shipped in 3/5, plus runtime fluid registration.

**Chebyshev caloric type.** A new `INCOMPRESSIBLE_CHEBYSHEV` data type. At load time `validate()` builds the derived coefficient matrices once — `d/dT` (the density fit needs it for the pressure terms of `h` and `s`), `∫f dT` (enthalpy), and `∫f/T dT` (entropy, from a Chebyshev–Lobatto re-expansion of `f/T`). Integration constants are irrelevant because the backend always subtracts the same expression at the reference state. Evaluation collapses the composition direction to an effective 1D coefficient vector and runs Clenshaw — the same collapse-then-evaluate structure `Polynomial2DFrac` already uses.

The loader **prefers** `_cheb` when present and falls back to the classic polynomial otherwise, so this is backward-compatible with any fluid file that lacks the new entries. A `_cheb` entry that exists but fails to parse as `chebyshev` is a hard error rather than a silent fallback, so a typo in `type` cannot leave the author believing the Chebyshev fit is in use.

**Runtime registration.** `add_fluids_as_JSON(&#34;INCOMP&#34;, ...)` now works, closing #2384.

**On the relationship to the superancillaries:** CoolProp already has `superancillary::ChebyshevExpansion`, which backs the piecewise-Chebyshev saturation curves and uses the same Clenshaw recurrence. This deliberately does not build on it: that class is 1D-in-T, whereas the incompressible fit is 2D (T and composition), and it has no antiderivative — `do_derivs()` differentiates, but nothing integrates, and an exact enthalpy from an exact cp fit is the entire point here. The 1D Clenshaw evaluator in `IncompressibleFluid.cpp` is therefore a documented, intentional duplicate of `ChebyshevExpansion::eval`, kept file-local so the incompressible public header does not pull in the superancillary module for one method. If `ChebyshevExpansion` gains `derivative()`/`antiderivative()` upstream — wanted independently for the superancillary side — the 1D half of this path could delegate to it, leaving only the composition collapse and the cp/T re-expansion as incompressible-specific.

### Benefits

* Enthalpy and entropy are now derived the same exact way for every shipped fluid, with no `Tbase`-adjacent singularity class possible in this basis at all.
* Incompressible fluids can be registered at runtime, not only at build time (#2384).

### Possible Drawbacks

* `add_fluids_as_JSON(&#34;INCOMP&#34;, ...)` mutates the library without synchronization. This is documented in the code with the reasoning: a write-only lock would be *worse* than none, because every reader — and so every property evaluation — walks the same containers unlocked and would keep racing while appearing protected. Making it genuinely safe means locking the property-evaluation hot path, which is a backend-wide decision rather than something to smuggle in here. Until then the contract matches the pre-existing `HEOS`/`Cubics` runtime-add paths: register during start-up, before other threads query the library.

  Note the THREAD SAFETY comment states that start-up-only contract, while 5/5 adds a GUI that registers fluids at runtime from a Tauri command. The write ordering below narrows the racy read (a concurrent reader either misses the name or finds a fully-published fluid), but concurrent `std::map` insert versus traversal is still UB — this is not a thread-safety fix and does not claim to be.
* A second caloric code path now exists alongside the polynomial one. The loader preference is a single compile-time constant (`prefer_chebyshev_caloric`) so the two can be A/B&#39;d against the same library.

### Verification Process

* `CatchTestRunner` (Release): `[PolyMath]`, `[INCOMP]`, `[IncompressibleFluids]`, `[IncompressibleBackend]` — **1804 assertions, 9 cases**, all passing.
* A runtime `add_fluids_as_JSON(&#34;INCOMP&#34;, ...)` round-trip with hand-checkable coefficients: on T ∈ [280, 360], `u(T) = (T−320)/40`, `rho = 1000 − 5u`, `cp = 2000`, then `Dmass`/`Cpmass`/`Hmass`/`Smass` read back through the real backend.
* Negative tests for the fail-open cases found in review, each of which previously passed silently: an empty coefficient matrix (Clenshaw over no coefficients returned `0.0`, so `rho == 0` gave infinite enthalpy with **no error**, reachable at runtime via `&#34;coeffs&#34;: []`); duplicate names on re-registration; zombie map slots after a failed parse; a pure/solution flip; and names containing `,|[]:&amp;` which corrupt the comma-joined fluid lists and the `INCOMP::Name` / `Name[x]` parsing.
* The every-shipped-fluid `Tbase` sweep from 1/5 continues to pass with the Chebyshev path active.
* `clang-format` clean, and CI&#39;s diff-only `clang-tidy` reports nothing on any line this PR changes (verified by intersecting whole-file clang-tidy output against the 436 changed lines).

### Review findings addressed since the first push

Two commits here exist only to close review findings on this PR&#39;s own code:

* **Lint sweep.** The new `chebyshev` dispatch branch used the `!type.compare(...)` idiom, an implicit `int -> bool` conversion. Fixing only that branch would have left one arm of a six-branch chain reading differently from its five siblings, so the whole chain moved to `compare(...) == 0`. Verified free rather than assumed: clang-tidy reports exactly one finding on each sibling line — the same `readability-implicit-bool-conversion` — so the flip removes it and cannot surface anything new. The analogous `!res.compare(...)` chain in the interpolation-fraction parser is deliberately left alone: it is a separate function this PR does not otherwise touch, so it stays outside CI&#39;s diff-only scope.

* **A fail-quiet map access, fixed twice.** `add_one`&#39;s replace-in-place branch reached `fluid_map` through `operator[]`. `fluid_map` is a `std::map<std::size_t, IncompressibleFluid>`, so on a desync `operator[]` default-constructs an empty fluid and `is_pure()` is answered from it — and since a default fluid has `density.coeffs.cols() == 0` it reads as a *solution*, so the guard below would report a bogus classification mismatch or pass and overwrite the wrong slot.

  The first attempt used `at()`, which review showed made things worse: `std::out_of_range` does not derive from `CoolPropBaseError`, so it lands in the C wrapper&#39;s `catch (...)` arm in `CoolPropLib.cpp` — `errcode 3` with the message buffer never written. The Rust FFI consumer in 5/5 surfaces that buffer, so the user would have seen an error dialog with **no text**. It now uses `find()` plus an explicit `ValueError`, matching what `JSONIncompressibleLibrary::get()` already throws for the same missing-index condition, which crosses the boundary with its text intact and replaces two lookups with one.

  Review also showed the invariant the fix asserted did not hold: the append path published `string_to_index_map` *before* `fluid_map`, so &#34;name resolves to an index with no fluid&#34; was reachable from an ordinary allocation failure between the two writes — and because `fluid_map.size()` had not yet advanced, the *next* `add_one` for any other name computed the same index, so the poisoned name resolved to **a different fluid&#39;s data**: a silent wrong answer, not just a throw. `fluid_map` is now written first, closing both; the name vectors go last, so a throw there leaves the fluid resolvable but unlisted (inert) rather than listed but unresolvable.

  The pure/solution guard whose left-hand side that change rewrites is now covered by a test: re-registering the runtime-added fluid with a two-column density flips `is_pure()`, and the test asserts the thrown *message* rather than merely that something threw — the definition also passes through `validate()`, so a bare `CHECK_THROWS` would have passed on any unrelated validation error without touching the guard. Confirmed by disabling the guard, which fails the test.

### Known, not addressed here

* `add_obj` is **not** equivalent to the write ordering used in `add_one`: it calls the throwing `validate()` *between* its two map writes, never populates `name_vector_pure`/`name_vector_solution`, and skips the `,|[]:&amp;` name check. Same failure *direction* (an orphan `fluid_map` entry, not a poisoned name), so not worse, and its only call site is commented out — left for a separate change rather than audited as equivalent.
* Contiguity of `fluid_map`&#39;s keys is load-bearing (`index = fluid_map.size()`) with nothing enforcing it. Nothing erases today, so it holds; if an `erase` ever lands, that expression silently overwrites an existing fluid instead of appending.
* Replace-in-place move-assigns into a map node that `get()` may already have handed out by reference to a live `IncompressibleBackend`. Node stability keeps the pointer valid, and the classification guard is what stops the backend&#39;s cached `_fluid_type` going stale — but a solution→solution replacement that shifts `xmin`/`xmax` under a backend whose fractions are already set is not covered. Pre-existing to the replace feature.

### Applicable Issues

Closes #2384.

---

## Summary by CodeRabbit

* **New Features**
  * Added Chebyshev-based incompressible property fits (including density and specific heat), with support for derivatives, integrals, and temperature inversion.
  * Extended JSON-based fluid registration to support the incompressible backend.
* **Bug Fixes**
  * Hardened fluid loading/registration to prevent partially registered fluids when parsing or validation fails.
  * Improved validation/error reporting for Chebyshev metadata (including temperature-range coverage) and invalid fluid names.
* **Tests**
  * Expanded Chebyshev fit, integral/consistency, and runtime registration test coverage, including new negative-case assertions and stability adjustments.

### DIFF
--- a/include/CoolProp/fluids/IncompressibleFluid.h
+++ b/include/CoolProp/fluids/IncompressibleFluid.h
@@ -35,11 +35,27 @@ struct IncompressibleData
         INCOMPRESSIBLE_EXPPOLYNOMIAL,
         INCOMPRESSIBLE_EXPONENTIAL,
         INCOMPRESSIBLE_LOGEXPONENTIAL,
-        INCOMPRESSIBLE_POLYOFFSET
+        INCOMPRESSIBLE_POLYOFFSET,
+        INCOMPRESSIBLE_CHEBYSHEV
     };
     IncompressibleTypeEnum type;
     Eigen::MatrixXd coeffs;  //TODO: Can we store the Eigen::Matrix objects more efficiently?
-    //std::vector<std::vector<double> > coeffs;
+
+    /// Chebyshev fits only (type == INCOMPRESSIBLE_CHEBYSHEV). The rows of
+    /// coeffs are Chebyshev-in-T coefficients on [cheb_Tmin, cheb_Tmax]; the
+    /// columns multiply powers of (x - cheb_xbase). The derived matrices are
+    /// built once at load time by IncompressibleFluid::validate():
+    ///  - cheb_ddT:        coefficients of d/dT (the density fit needs this
+    ///                     for the pressure terms of h and s)
+    ///  - cheb_int_dT:     coefficients of the indefinite Int f dT (enthalpy)
+    ///  - cheb_int_dTdivT: coefficients of the indefinite Int f/T dT
+    ///                     (entropy; from a Chebyshev-Lobatto re-expansion of
+    ///                     f/T, which converges geometrically since T > 0)
+    /// Integration constants are irrelevant: the backend always subtracts the
+    /// same expression evaluated at the reference state.
+    double cheb_Tmin = 0.0, cheb_Tmax = 0.0, cheb_xbase = 0.0;
+    Eigen::MatrixXd cheb_ddT, cheb_int_dT, cheb_int_dTdivT;
+
     IncompressibleData()
       : type(INCOMPRESSIBLE_NOT_SET) {
 
--- a/include/CoolProp/fluids/IncompressibleFluid.h
+++ b/include/CoolProp/fluids/IncompressibleFluid.h
@@ -159,6 +159,14 @@ class IncompressibleFluid
       : strict(true), Tmin(_HUGE), Tmax(_HUGE), xmin(_HUGE), xmax(_HUGE), xid(IFRAC_UNDEFINED), TminPsat(_HUGE), xbase(_HUGE), Tbase(_HUGE) {
 
         };
+    // The virtual destructor below suppresses the implicitly-declared move
+    // constructor/assignment operator (even though it is only '= default'),
+    // so state them explicitly - otherwise callers that std::move() a fluid
+    // (e.g. JSONIncompressibleLibrary::add_one) silently fall back to a copy.
+    IncompressibleFluid(const IncompressibleFluid&) = default;
+    IncompressibleFluid(IncompressibleFluid&&) = default;
+    IncompressibleFluid& operator=(const IncompressibleFluid&) = default;
+    IncompressibleFluid& operator=(IncompressibleFluid&&) = default;
     virtual ~IncompressibleFluid() = default;
 
     std::string getName() const {
--- a/src/Backends/Incompressible/IncompressibleBackend.cpp
+++ b/src/Backends/Incompressible/IncompressibleBackend.cpp
@@ -939,7 +939,15 @@ TEST_CASE("Internal consistency checks and example use cases for the incompressi
         }
     }
     SECTION("INCOMP::ExamplePure") {
-        double acc = 0.0001;
+        // The golden values below were computed from the centered-polynomial
+        // fit of this example fluid's tabulated data. The Chebyshev caloric
+        // fits (preferred since the *_cheb entries shipped) reproduce the
+        // same data with an independently chosen order, so the two fits may
+        // legitimately differ at fit-quality level between the data points
+        // (~2e-4 relative for cp). The tolerance is fit-level, not
+        // bit-level; a wrong-fit-class regression is orders of magnitude
+        // larger.
+        double acc = 0.001;
         std::string fluid = std::string("INCOMP::ExamplePure");
         double T = +55 + 273.15;
         double p = 10e5;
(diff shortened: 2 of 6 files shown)


jowr:
@coderabbitai review

---

PR #3298, opened 2026-07-28
incompressible 5/5: add user-defined incompressible fluids from the Desktop GUI, and make the caloric path a runtime switch

jowr:
Last of five PRs splitting up #3292. **4/5 (#3297) is merged, so this now targets `master` directly** — the branch was rebased onto it and this PR&#39;s diff is its own 15 GUI files plus the 3 files of the runtime-flag change described below.

### Requirements

### Description of the Change

**An &#34;Add incompressible fluid&#34; dialog in the Tauri/React GUI.** The user pastes a property table (temperature plus any of ρ / cp / k / μ, SI units, header row recognised, blank cells meaning &#34;no data&#34;), the frontend fits it live, previews fit quality per property, and registers the fluid through the runtime-add path from 4/5. Definitions are persisted so they are re-registered on the next start.

The TypeScript fitter (`src/lib/incompfit.ts`) mirrors the offline Python pipeline: same Chebyshev basis for the caloric properties, same GCV degree selection, same positivity guard across the domain, and a centred monomial for conductivity / log-viscosity. It is cross-checked against the same `ExamplePure` golden value the C++ side uses.

A Rust FFI bridge exposes `add_incompressible_fluid` / `get_incompressible_fluids_list` as Tauri commands.

Because the two namespaces overlap — `Water` is a shipped fluid in **both** HEOS and INCOMP — the calculator&#39;s fluid dropdown now carries the backend explicitly in its value (`HEOS:Water` / `INCOMP:Water`). Previously a colliding INCOMP fluid could never be selected, so those entries were filtered out of the list and silently vanished, which made the shipped `INCOMP::Water` unreachable.

**The caloric-path selector becomes a runtime flag.** 4/5 chose between the Chebyshev and classic-polynomial caloric fits with a `static constexpr bool` hardcoded to `true`, which made the polynomial branch unreachable in any shipped binary and meant the documented &#34;flip to A/B&#34; workflow needed a recompile. It is now a standard configuration key, `INCOMPRESSIBLE_PREFER_CHEBYSHEV_CALORIC`, default `true` — so the Chebyshev fits remain the shipped behaviour and anyone can return to the old equations for density and specific heat without rebuilding.

The ordering is the part worth knowing, and it is documented on the key itself:

* the value is read **per fluid, as that fluid is parsed**;
* the shipped library is parsed once on first use, so changing the *shipped* fluids means setting the flag before any incompressible fluid is touched. The reliable route is the `COOLPROP_INCOMPRESSIBLE_PREFER_CHEBYSHEV_CALORIC` environment variable, which CoolProp applies when the configuration object is first constructed — earlier than the library load;
* fluids registered later through `add_fluids_as_JSON(&#34;INCOMP&#34;, ...)` always honour the then-current value, so `set_config_bool` works for those with no ordering caveat.

### Benefits

* A user can add an incompressible fluid without touching Python, the build system, or the fitting internals.
* Fit quality is shown before the fluid is registered, so a bad paste is caught immediately rather than becoming silently wrong properties.
* `INCOMP::Water` (and any other name shared with HEOS) is selectable in the GUI for the first time.
* The pre-Chebyshev caloric equations stay reachable in a shipped binary — both as an escape hatch if a Chebyshev fit misbehaves for someone, and to A/B the two paths against the same library.

### Possible Drawbacks

* **Pure fluids only.** Solutions need a 2-D T×composition table UI, which is not attempted here; filed as a follow-up.
* Fitting in the frontend duplicates the Python pipeline&#39;s logic in TypeScript. The cross-check against the shared golden value is what keeps the two honest; a shared implementation would mean shipping the fitter through the FFI, which is a much larger change.
* Registration inherits the thread-safety caveat documented in 4/5.
* The runtime flag cannot retroactively change fluids that are already parsed. That is inherent to a library parsed once at first use — the alternative would be supporting a full library reload, which is a much bigger change than the escape hatch warrants. The environment variable covers the case that matters.

### Verification Process

* `tsc --noEmit` clean; `vitest` — 33 tests, including the fitter cross-checked against the C++ side&#39;s `ExamplePure` golden value.
* `cargo test --release` — 18 tests, including an end-to-end round-trip that registers a fluid through the real FFI and reads back density / cp / enthalpy / entropy.
* `CatchTestRunner`: `[PolyMath]`, `[INCOMP]`, `[IncompressibleFluids]`, `[IncompressibleBackend]` — **1812 assertions, 10 cases**.
* **The flag&#39;s effect is observable, not inferred.** The test fluid carries *both* representations with deliberately different constant densities (polynomial 800, Chebyshev 900), so each section asserts a different value and neither would pass if the wrong path ran. The default is asserted to be `true`, and the previous value is restored by a scope guard so the global configuration cannot leak into another test — verified across 10 randomized-order seeds.
* Review findings addressed here, each a real fail-open: new names are validated against **both** namespaces (previously only INCOMP, so a new fluid could collide with a HEOS name and become unreachable); shadowed INCOMP entries are labelled rather than hidden; the exactly-determined fit fallback reported `rms`/`nrms` of `0`, which rendered as a confident &#34;0.0%&#34; for an interpolation whose accuracy is by definition unmeasured, and now reads &#34;n/a (interpolated)&#34;; `loadUserFluidDefinitions` no longer throws on a corrupt storage key that parses to a non-object.
* Accessibility: `role=&#34;dialog&#34;` / `aria-modal`, focus moved into the dialog on open, Escape closes it (overlay click already cancelled), and every visible label tied to its control.
* The runtime-add test is serialized on a guard — it mutates the process-wide INCOMP library while cargo runs tests on parallel threads, so it could previously race a concurrent enumeration.
* `build.rs` now emits `rerun-if-env-changed` for the whole supported `CPM_<Name>_SOURCE` set rather than only the variables that happen to be set, so adding a local dependency override invalidates a cached build. (This is also what makes the crate buildable behind a restrictive network policy, where the CoolProp cmake step would otherwise fail fetching its dependencies.)

### Applicable Issues

Completes the user-facing half of #2384.

---

## Summary by CodeRabbit

* **New Features**
  * Added GUI support for creating incompressible fluids from pasted SI property tables, with fit previews, validation, and warnings.
  * User-defined fluids are saved across restarts and available alongside built-in fluids for property calculations.
  * Added runtime registration of incompressible fluids through the API.
  * Added a setting to choose Chebyshev or polynomial caloric representations.

* **Bug Fixes**
  * Improved enthalpy and entropy calculations, avoiding issues at the reference temperature.

* **Tests**
  * Added coverage for table parsing, fitting, runtime registration, and caloric representation selection.


jowr:
@coderabbitai review

---

PR #2840, opened 2026-05-01
Add Qmass (mass-basis vapor quality) support across CoolProp backends

ibell:
## Summary

- Add full `Qmass` (mass-basis vapor quality) support: 8 new `*Qmass*_INPUTS` input pairs paralleling each existing Q-pair, a new `iQmass` keyed parameter, and a `Qmass()` accessor on `AbstractState` — symmetric with how `Smass`/`Smolar` and `Hmass`/`Hmolar` work today.
- Mixture support works from day one. HEOS uses a bracketed solver on Qmolar (5–8 iterations on typical refrigerant blends). REFPROP uses its native `kq=2` fast path in `TQFLSHdll`/`PQFLSHdll` for `QmassT_INPUTS`/`PQmass_INPUTS` (zero overhead vs. the molar siblings); the other 6 Qmass pairs fall through to the inherited solver.
- Pure / pseudo-pure fluids: `Qmass == Qmolar` exactly, so the input pair is rewritten to its molar sibling in `mass_to_molar_inputs` — no iteration.
- Conversion math lives in `include/qmass_conversions.h` (`CoolProp::detail::Qmolar_to_Qmass` / `Qmass_to_Qmolar` — pure free functions, no AbstractState dependency).
- Supersedes #2221 and #2838.

## Architecture

- Output direction (`Qmass()`): explicit. `_Q` (molar) + `MM_l` + `MM_v` from `calc_phase_molar_masses` (virtual; default = pure-fluid trivial; HEOS and REFPROP overrides handle mixtures).
- Input direction (`*Qmass*_INPUTS`): explicit for pure fluids; iterative for mixtures because at fixed `T`+`Qmass` the equilibrium phase compositions (and therefore `MM_v`) depend on the very Qmolar we're solving for.
- Default base implementation of `update_Qmass_pair` does the bounded solve. REFPROP overrides for the two pairs where it has a native flag.

## Files

- `include/qmass_conversions.h` (new) — free functions
- `include/DataStructures.h` — `iQmass` + 8 new `*_INPUTS` + `is_Qmass_pair` helper + 4 new `match_pair` branches in `generate_update_pair`
- `src/DataStructures.cpp` — parameter/pair table rows + 8 new cases in `split_input_pair` (the latter is needed so PropsSI's input/output overlap heuristic works for Qmass pairs)
[...]

PR #3308, opened 2026-08-03
Qmass pair with singlephase-like boundaries incl transport properties (more aligned with molar quality methods)

kaern:
### Description of the Change

The PQmass and QmassT (mass-quality two-phase) input pairs in the REFPROP backend have been updated to use the same flash strategy already applied to the molar PQ and QT inputs. At the phase boundaries (Q=0 bubble point, Q=1 dew point), the code now first calls SATTPdll (falling back to SATTdll/SATPdll on failure) to obtain saturated-phase densities, then calls THERMdll to compute the remaining thermodynamic properties. Caloric and transport property results from THERMdll are kept while the saturation pressure/temperature is preserved from the SATTP/SATT/SATP call to avoid small rounding discrepancies. For intermediate quality values (0 < Q < 1) the existing TQFLSHdll/PQFLSHdll routines are still used as a fallback.

### Benefits

Numerical consistency between the molar (PQ/QT) and mass-quality (PQmass/QmassT) paths at the saturation boundary. The saturation routines (SATTP/SATT/SATP) are more robust for bubble/dew-point calculations than the general flash routines, giving more reliable density and thermodynamic-property results right at Q=0 and Q=1.

### Verification Process

Added two example files that calls both PQmass and QmassT at Q=0 and Q=1 and compares results against the corresponding molar PQ/QT calls.

### Applicable Issues

#2221 
#3011 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved REFPROP mixture calculations at liquid and vapor quality boundaries.
  - Improved handling of mass- and molar-quality inputs for pressure, temperature, viscosity, and density calculations.
  - Added more robust recovery when saturation-property calculations encounter errors.
  - Ensured equilibrium pressure and phase-specific density results remain consistent across supported quality states.

- **Tests**
  - Added diagnostic coverage for R410A mixture properties across temperature-quality and pressure-quality conditions.

### DIFF
--- a/dev/Qmass_boundaries.cpp
+++ b/dev/Qmass_boundaries.cpp
@@ -1,0 +1,62 @@
+#include "AbstractState.h"
+#include "CoolProp.h"
+#include <iostream>
+#include <iomanip>
+
+using namespace CoolProp;
+
+int main() {
+    std::cout << "Testing mass quality calculations with R410A.MIX at 278.15 K and 0.5 quality" << std::endl;
+    std::cout << "=================================================" << std::endl;
+
+    try {
+        shared_ptr<AbstractState> AS(AbstractState::factory("REFPROP", "R410A.MIX"));
+
+        AS->update(QT_INPUTS, 0.5, 278.15);
+        std::cout << "  Viscosity: " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
+        std::cout << "  Pressure: " << std::setprecision(12) << AS->p() << " Pa" << std::endl;
+
+        AS->update(QT_INPUTS, 0.0, 278.15);
+        std::cout << "  Viscosity from QT (0.0, 278.15): " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
+
+        AS->update(QT_INPUTS, 1.0, 278.15);
+        std::cout << "  Viscosity from QT (1.0, 278.15): " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
+
+        AS->update(QmassT_INPUTS, 0.0, 278.15);
+        std::cout << "  Viscosity from QmassT (0.0, 278.15): " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
+
+        AS->update(QmassT_INPUTS, 1.0, 278.15);
+        std::cout << "  Viscosity from QmassT (1.0, 278.15): " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
+
+        AS->update(QmassT_INPUTS, 0.5, 278.15);
+        std::cout << "  Viscosity from QmassT (0.5, 278.15): " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
+
+        AS->update(QT_INPUTS, 0.5, 278.15);
+        std::cout << "  Viscosity from QT (0.5, 278.15): " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
+
+        AS->update(QmassT_INPUTS, 0.5, 278.15);
+        std::cout << "  Viscosity from QmassT (0.5, 278.15): " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
+
+        // -----------------------------------------
+
+        AS->update(PQ_INPUTS, 9e5, 0.5);
+        std::cout << "  Viscosity: " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
+
+        AS->update(PQ_INPUTS, 9e5, 0.0);
+        std::cout << "  Viscosity from PQ (9e5, 0.0): " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
+
+        AS->update(PQ_INPUTS, 9e5, 1.0);
+        std::cout << "  Viscosity from PQ (9e5, 1.0): " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
+
+        AS->update(PQmass_INPUTS, 9e5, 0.0);
+        std::cout << "  Viscosity from PQmass (9e5, 0.0): " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
+
+        AS->update(PQmass_INPUTS, 9e5, 1.0);
+        std::cout << "  Viscosity from PQmass (9e5, 1.0): " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
+
+        return 0;
+    } catch (const std::exception& e) {
+        std::cerr << "ERROR: " << e.what() << std::endl;
+        return 1;
+    }
+}
--- a/dev/Qmass_boundaries.cpp
+++ b/dev/Qmass_boundaries.cpp
@@ -13,46 +13,57 @@ int main() {
         shared_ptr<AbstractState> AS(AbstractState::factory("REFPROP", "R410A.MIX"));
 
         AS->update(QT_INPUTS, 0.5, 278.15);
-        std::cout << "  Viscosity: " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
         std::cout << "  Pressure: " << std::setprecision(12) << AS->p() << " Pa" << std::endl;
+        std::cout << "  Temperature: " << std::setprecision(12) << AS->T() << " K" << std::endl;
+        std::cout << "  Viscosity: " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
 
         AS->update(QT_INPUTS, 0.0, 278.15);
-        std::cout << "  Viscosity from QT (0.0, 278.15): " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
+        std::cout << "  Viscosity from QT (0.0, 278.15): " << std::setprecision(12) << AS->viscosity() << " Pa.s (" << std::setprecision(12)
+                  << AS->p() << " Pa, " << std::setprecision(12) << AS->T() << " K)" << std::endl;
 
         AS->update(QT_INPUTS, 1.0, 278.15);
-        std::cout << "  Viscosity from QT (1.0, 278.15): " << std::setprecision(12) << AS->viscosity() << " Pa.s" << std::endl;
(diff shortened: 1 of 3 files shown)


jowr:
Thank you for your contribution, @kaern. It seems like your email address in the commits does not match your GitHub account. The contributor license agreement can only be signed on GitHub, any chance you could change the email in the old commits?
While you are at it, maybe you can also address the CodeRabbit comments?

kaern:
Hi Jorrit, I was able to change the commits to the private github account. Hopefully that works. 

The example files i placed in dev, you can (should) delete. They were just for testing the functions, so i did not consider coderabbit suggestions on these. (unit test could be made on phase boundary states with molar/mass method that should give the same result (also for transport properties).

However, I tried to make the PQmass and QmassT as similar to the PQ and QT methods, so there are definitely identical code on the phase boundary between these methods. I did not understood why the Qmass pair needed special "update_Qmass_pair" function and early handle. I guess with the suggested single phase (phase boundary) additions, this special treatment maybe unnecessary, and the PQmass and QmassT handles could be placed at same if statement as PQ and QT. But at this stage, I did not want to touch too much more than demonstrating something wokring, and I let you decide what to do implementation wise.

kaern:
I remember Ian wanted to implement this with Claude, so Im thinking if Claude decided to add the update_Qmass_pair function, then I would definitely revert it and make Qmass implementations similar to Qmoler and reuse the code on the phase boundaries. just let me know what to do and I will make a unit test on phase boundaries as well

ibell:
Thanks @kaern — I dug into this with REFPROP 10 loaded locally and built master and your branch side by side. Two findings, and the second one is yours.

**The boundary rerouting is a no-op on current master.** `update_Qmass_pair` landed in #2840 on 2026-05-01, and with it the mass- and molar-quality paths are already bit-identical at Q=0 and Q=1 — viscosity included — across R410A.MIX, R407C.MIX and Methane/Ethane, with and without the `SATSPLN` splines active:

```
QT     Q=0 : p=9.362298710462663e+05 rho=1.583781339403651e+04 eta=1.518635715991902e-04
QmassT Q=0 : p=9.362298710462663e+05 rho=1.583781339403651e+04 eta=1.518635715991902e-04
```

So the `SATTPdll` path changes nothing numerically at the boundaries. Whatever you originally hit was almost certainly fixed by #2840 in between.

**Your `clear()` was the real bug.** The three lines you added with `//*! I think we need to clear here as well ?` — that hunch was right, and it turns out to be a live correctness bug in shipped master. `update_Qmass_pair` bypasses `update()`'s switch, so it never cleared the cache, and every lazily-cached derived property from the previous update survived:

```cpp
AS->update(PT_INPUTS, 5e6, 250.0);       // R32+R125 compressed liquid
AS->update(QmassT_INPUTS, 0.5, 278.15);  // two-phase
AS->viscosity();  // -> 2.4897e-04 Pa.s, the LIQUID value (~20x too high;
                  //    the correct two-phase value is 1.2732e-05)
```

Same for conductivity, and `tau`/`delta` were never set at all. That is exactly the "transport properties look wrong with Qmass" symptom you started from — just reached from a different direction than the phase boundary.

I've isolated it into #3320 (`clear()` + `check_status()` + `_tau`/`_delta`/`_gibbsmolar`, with a regression test that goes red on master). You found it; the credit is yours.

**One blocking issue in this branch, for the record.** The `PQmass` branch declares `int iFlsh = 0, iGuess = 0, ierr = 0;`, which shadows the outer `ierr` that the shared error check reads — CodeRabbit flagged it and it reproduces:

```
                        master                 this branch
PQmass 55MPa Q=0   throws (PQFLSH 141)   T=300 K, rho=6324.647, Q=0, eta=2.73e-05
PQmass 1Pa   Q=0   throws (PQFLSH 3)     T=116.5 K, rho=22811.9, Q=0
```

A failed flash silently returns a fabricated state (that `6.324647308314624` is REFPROP's leftover critical density from line 2 of `R410A.MIX`). Dropping `, ierr = 0` restores correct throwing — I verified the one-liner.

**On your question about `update_Qmass_pair` existing separately.** Good instinct. Rather than a third and fourth copy of the `SATTP` → `SATT`/`SATP` → `THERM` sequence, the right shape is the helper extraction CodeRabbit suggested (`sat_flash_TQ` / `sat_flash_PQ`) used by all four call sites — I've filed that so it doesn't get lost. Since the numbers don't move at the boundaries, it's a maintainability change, not a fix, so it can be its own PR without any pressure behind it.

So: happy to close this one in favour of #3320, or if you'd rather keep it, drop the boundary rewrite and the `dev/` files and it becomes the same small change. Either is fine by me — and thanks for the report, it caught something real.

kaern:
Thanks for your efforts :-) Let's use #3320 as you suggest and put update_Qmass_pair helpers on the todo list .. I will try to test this as well in the coming days, so keep it open and i'll get back with findings...

kaern:
I had a look again at the code and yes, it does now give the same results at the boundaries: 

  Pressure: 934815.109607 Pa
  Temperature: 278.15 K
  Viscosity: 1.26874845808e-05 Pa.s
  Viscosity from QT (0.0, 278.15): 0.000151863571599 Pa.s (936229.871046 Pa, 278.15 K)
  Viscosity from QT (1.0, 278.15): 1.2460510974e-05 Pa.s (933150.89911 Pa, 278.15 K)
  Viscosity from QmassT (0.0, 278.15): 0.000151863571599 Pa.s (936229.871046 Pa, 278.15 K)
  Viscosity from QmassT (1.0, 278.15): 1.2460510974e-05 Pa.s (933150.89911 Pa, 278.15 K)
  Pressure: 900000 Pa
  Temperature: 276.918730008 K
  Viscosity: 1.25978470159e-05 Pa.s
  Viscosity from PQ (9e5, 0.0): 0.000154276461459 Pa.s (900000 Pa, 276.869879844 K)
  Viscosity from PQ (9e5, 1.0): 1.24046311667e-05 Pa.s (900000 Pa, 276.976402434 K)
  Viscosity from PQmass (9e5, 0.0): 0.000154276461459 Pa.s (900000 Pa, 276.869879844 K)
  Viscosity from PQmass (9e5, 1.0): 1.24046311667e-05 Pa.s (900000 Pa, 276.976402434 K)
  
However, i am still concerned about the speed issues and why the Qmass and Qmolar methods are different. For now QmassT/PQmass @ 0.0 or 1.0 use TQFLSHdll and PQFLSHdll, whereas QT/PQ (molar) @ 0.0 or 1.0 use SATTPdll and THERMdll. Wouldn't it improve speed to choose the latter? To my understanding, drop the update_pair function, add PQmass and QmassT to the long list of "switch (input_pair) { case PT_INPUTS ..... }" make helper functions for the boundary rewrite to be used by both mass/molar methods ? (add drop my test files from dev).. ?

ibell:
Great question — I went and measured it, because it's exactly the kind of thing where the code *looks* like it should matter.

**Short version: the two entry points cost the same at the boundaries, so there's no speed to win there. But your instinct about the shape of the code is still right, and it's worth doing for its own sake.**

### The measurement

REFPROP 10, R32+R125 at R410A composition, 800 temperatures over 240–300 K, interleaved and min-of-4-rounds so caching and thermal drift can't favour one path:

```
QT     Q=0.0      40.3 us/call        QmassT Q=0.0      40.5 us/call
QT     Q=1.0      34.7 us/call        QmassT Q=1.0      35.1 us/call
QT     Q=0.5     376.6 us/call        QmassT Q=0.5     362.5 us/call
```

Same for the P,Q pairs (39.7 / 40.7 µs at Q=0 and 40.4 / 38.5 µs at Q=1). The differences are ~1%, comfortably inside run-to-run noise, and they don't even point consistently in the same direction.

And it isn't just that the timings are close — the results are identical. Over 200 temperatures × {Q=0, Q=1} and 200 pressures × {Q=0, Q=1}, comparing `p`, `rhomolar`, `hmolar`, `smolar`, `viscosity`, `conductivity` and `Q`:

```
QT vs QmassT boundary: 400/400 bit-identical
PQ vs PQmass boundary: 400/400 bit-identical
```

Bit-identical output *and* identical cost is the fingerprint of `TQFLSHdll`/`PQFLSHdll` noticing q = 0 or 1 and handing off to the same `SATT`/`SATP` solver underneath. So the `SATTPdll` route isn't skipping any work that the flash route is doing — there's simply nothing duplicated to remove.

The row I'd actually draw your attention to is the third one: **interior quality costs ~10× the boundary (≈370 µs vs ≈40 µs), and it's equally slow on both bases.** If something feels slow in practice, that's where it's coming from, and it's not Qmass-specific — molar `QT` pays it too.

### On the refactor

I do think it's worth doing, just framed as tidying rather than as a fix, and a bit narrower than "drop `update_Qmass_pair`":

- **The override has to stay in some form.** `is_Qmass_pair` covers 8 pairs, and REFPROP has a native `kq=2` flag for only two of them. The other six (`QmassSmolar`, `HmassQmass`, `DmassQmass`, …) fall through to the base-class TOMS748 root-find on Qmolar. So only the `QmassT`/`PQmass` arm can move into the `switch`; the rest still needs somewhere to land.
- **The interior path should keep `kq=2`.** That's the one regime where mass basis genuinely differs from molar, and REFPROP-native beats the TOMS748 fallback by a lot.
- **The boundary arm can just delegate.** Since `Qmass_to_Qmolar(0) = 0` and `(1) = 1` exactly, routing |q| < 1e-10 and |q−1| < 1e-10 straight to `QT_INPUTS`/`PQ_INPUTS` makes that bit-identity structural instead of a happy coincidence — and `_Qmass` still comes out right on its own via the lazy `calc_Qmass()`, so nothing extra to set. CodeRabbit's `sat_flash_TQ` / `sat_flash_PQ` extraction is the other way to get there; either one collapses the duplication you spotted.

### The unit test

Yes please, and it's worth having regardless of whether the refactor happens. There are `[Qmass][REFPROP]` tests already, but none of them assert `QT == QmassT` at the boundaries — which is precisely the invariant everything above rests on. If it ever silently stops holding, nothing currently notices.

So to summarise where I'd land: your `clear()` bug is fixed and shipped in #3320, the boundary rerouting doesn't buy speed and can be dropped, and the helper extraction + the boundary test are both genuinely worth having as a small follow-up whenever you feel like it. Thanks for pushing on this — the speed question was a fair one to ask, and now we have numbers on it instead of a guess.

kaern:
Thanks for providing this detailed response (as always) :-) You convinced me and I will rest my case, close this PR.. Thanks for the Qmass methods. This completes the Refprop backend in my opinion and I am very grateful for this feature.

ibell:
That's all Claude's doing :)

Issue #1360, opened 2016-11-30
Chemical reactions?


erlen07:
I think chemical reactions could also be handled with CoolProp if reference enthalpies and entropies were added to allow chemical calculations.
E.g. Combustion:  H2 + 1/2 O2 ->H2O
The reaction enthalpy at the reference temperature and pressure (298.15 K, 101325 Pa) should be -286 kJ/mol.
If current CoolPorp enthalpies are added up with thei stoichiometric coefficients, the result is wrong (-7 kJ/mol) because the enthalpies at refernce state do not correspond.
Kind regards Georg

jowr:
Have you tried the custom reference state functions? That might be the solution for your issue.

ibell:
@jowr has basically the right answer.  We need to add a reference state that sets the enthalpy and entropy to their standard values of formation at STP (298 K, 1 atm).  That's relatively fine for things that are gaseous, and gets more complex for water.  Nonetheless, this issue is intimately related: https://github.com/CoolProp/CoolProp/issues/58 .  If someone (hint, not me!) collected the standard molar enthalpies of formation for all the fluids in CoolProp, we could implement this without too much work.

ibell:
Bump.  Want to collect the standard molar enthalpies of formation?

alchemyst:
Just a note: Standard enthalpies of formation are calculated at 25 °C and 1 atm this is different from STP, which is [0 °C and 1 bar](https://en.wikipedia.org/wiki/Standard_conditions_for_temperature_and_pressure).

ibell:
Interesting - I should have known that given my current employment, but I
had always thought that STP was 298 K and 1 atm.  Evidently I was wrong.
Learn something new every day!

guipradow:
Hi, i'm interested in use CoolProp to handle chemical reactions. How can I set the reference state to STP (25ºC = 298.15 K and 1 atm)? Using set_reference_stateD(), how can I define the density at the reference point?

ibell:
Well, first thing is to do a calculation with pressure and temperature as
inputs, then that gives you density.

samiit:
Hi @ibell 

I checked that the issue still persists in CoolProp. I want to use CoolProp to evaluate dH, dS and dG of reactions, or even equilibrium constants and as of now cannot due to the enthalpy of formation issues.
If you have any other way to get dG and K_eq, I would appreciate?

Thanks in advance.

ibell:
[...]

Issue #58, opened 2014-06-26
Add enthalpy and entropy of formation


ibell:

ibell:
@pdewallef Do you still have an interest in this?  If you can collect the data I would be happy to integrate it into CoolProp

ibell:
@pdewallef - there is still interest in this - see #1360.  Can you, or one of your students, give me a list of standard molar enthalpies/entropies of formation for all fluids in CoolProp?

rodskades:
@ibell is there still interest in this? I'll need to gather these values for my "final essay" in mechanical engineering university, I could send them to you... I'll need one or two months to gather it all probably...

ibell:
Hi! Any progress on this?  It would still be a nice feature in CoolProp

ProfLeao:
I'm still interested in that. The data can certainly be collected from "The NBS Tables of Chemical Thermodynamic Properties" or even from the "CRC Handbook of Chemistry and Physics".

ibell:
Closing as part of an issue-tracker cleanup — this issue has had no activity in over two years.

*To be transparent: this comment and the close action are being performed by Claude (Anthropic's AI coding assistant), running on behalf of @ibell. Ian reviewed and approved the bulk-close criteria; I am the one executing the action.*

If you would like this issue reconsidered, please feel free to **re-open it** — a fresh comment with current behaviour against the latest CoolProp release is the most useful way to revive the discussion.

— Claude (acting for @ibell)

PR #3309, opened 2026-08-08
feat: standard enthalpy of formation from ATcT (HFORMATION) — Tier 1 of #1360

ibell:
Implements **Tier 1** of #1360 ("Chemical reactions?", open since 2016-11-30): stores the standard ideal-gas enthalpy of formation at 298.15 K as per-fluid metadata, and exposes it as the trivial parameter `HFORMATION`.

```python
>>> CoolProp.CoolProp.PropsSI("HFORMATION", "", 0, "", 0, "Methane")
-74513.0   # J/mol
```

## Why now

#1360 has had exactly one blocker for ten years: nobody produced the data. It was requested in 2016, 2019 and 2020, and the companion #58 was bulk-closed for inactivity. On 2026-07-27 **@Tobias-Reiter** attached a curated CSV of 75 ATcT values plus a REFPROP cross-check program, which unblocked it. **@matthewzyates** had earlier posted a working recipe validated against JANAF. This PR builds on both.

## Regenerated from source, not vendored

`dev/atct/fetch_atct_formation.py` derives every value directly from the [ATcT TN 1.220](https://doi.org/10.17038/CSE/2568691) published table, so no number in the repo rests on a hand-curated spreadsheet. That decision paid for itself twice:

- **Found a fluid the CSV missed** — R1132a (CAS 75-38-7, 1,1-Difluoroethene), −350660 ± 630 J/mol. 76 fluids covered, not 75.
- **Corrected a wrong value** — the CSV lists ortho- and para-hydrogen as `0.000 ± 0.000`. ATcT publishes a spin-resolved `H2 (g, para)` = **−0.058 kJ/mol** (ID `1333-74-0*2`). CoolProp's suffixed CAS (`1333-74-0p`) now binds to that row.

Coverage is 76 of 137 fluids. R134a, R1234yf and the other HFC/HFOs are **genuinely absent from ATcT** — no scraper improvement recovers them. Absent fluids get no block, and the getter throws.

## Why ATcT rather than REFPROP

REFPROP's `HEATFRMdll` is a *derived* quantity, not an independent measurement. Regressing its output against the gross heating value in each `.FLD` header recovers, with **max residual 0.000 kJ/mol** over 14 hydrocarbons:

| implied constant | fitted from REFPROP | CODATA |
|---|---|---|
| Δ<sub>f</sub>H(CO₂) | −393.5110 kJ/mol | −393.51 |
| Δ<sub>f</sub>H(H₂O, l) | −285.8300 kJ/mol | −285.83 |

Of 148 `.FLD` files, 82 hold `????` and 65 hold a bare number with **no citation**; exactly one (methane) cites ISO 6976:2016. Over the 41 overlapping species the two sources disagree by RMS 0.98 kJ/mol, worst **cyclobutene at 7.2σ**, with only 8/41 inside ATcT's 1σ. ATcT is version-stamped, per-species traceable and uncertainty-quantified, so it is the reference here. **No REFPROP code path is touched by this PR.**

## What reviewers should look at

**`kRevision` 19 → 20 in `SVDSurfaceSerializer.h`.** `iHmolar_formation` is inserted in the `// General parameters` block after `idipole_moment`, which renumbers every later entry. `SVDSurfaceSerializer.cpp` packs `CoolProp::parameters` as raw `int32` (`:142` write, `:337` read). A cache written under the old numbering would **pass the revision check and then silently decode property keys to the wrong parameter** — wrong values, no error. The bump forces rejection and rebuild. Verified by file mtimes: only surfaces the tests query were rebuilt, i.e. per-file rejection, not a directory redirect. CI could never have caught this (fresh `$HOME` every run); it would have hit only users upgrading with a warm cache.

**Hand-maintained enum mirrors.** `nanobind_interface.cxx` and `emscripten_interface.cxx` enumerate the enum member by member, so the new entry was added to both; the committed nanobind stub is regenerated with the pinned `nanobind==2.12.0` toolchain. The Cython constants and SWIG need no change — both are generated from `DataStructures.h`. Follow-up filed (`CoolProp-jxmc`): three artifacts must move together on any `parameters` change and only one of them fails loudly.

**Fluid-file key reordering.** `dev/fluids/*.json` has not been canonical since #2780, so writing any file with `package_json.json_options` reorders pre-existing keys. Commit `chore(fluids): canonicalize JSON key order…` does that alone for the 75 affected files — **pure reordering, every file verified parse-identical to its parent** — so the data commit that follows is 76 single-hunk additions and is actually reviewable.

## Disclosures

- The design spec's original acceptance criterion called for reproducing all 75 CSV values. That was **descoped**: committing the CSV as a fixture reintroduces the hand-curated artifact this design exists to eliminate. Replaced by a four-value spot-check (methane, water, CO₂, para-hydrogen) plus the coverage ledger. The spec records this.
- A **malformed** (not merely absent) `hmolar_formation` block aborts the whole `add_many` fluid-library load with a message to stdout. This is pre-existing behaviour shared with `parse_environmental`, not new fragility.
- ATcT's own redistribution terms could not be independently verified (the site blocks automated fetches). Values are cited with DOI, TN version and per-species ID throughout, consistent with field convention for thermochemical constants.

## Fail-loud by construction

`dev/atct/expected_coverage.json` pins every fluid's state. The regenerator **exits non-zero** if any fluid's coverage changes, so a future ATcT version that renames or drops a species is a failing run rather than a silent gap. Ambiguous CAS matches are errors, never a guess. `--update-ledger` now prints what it is about to overwrite.

## Scope

Enthalpy only. Deliberately **not** included, each needing its own design: standard entropy S°(298.15 K) — ATcT publishes only enthalpies, so it needs a different source and an explicit nuclear-spin convention decision; heating values HHV/LHV; a `"FORMATION"` reference state; reaction equilibrium (belongs downstream, per @matthewzyates). The `INFO.STANDARD_STATE` block attaches provenance **per quantity**, so `smolar` drops in later without a second migration through 137 files.

## Testing

- 29 offline pytest cases covering the three silent-failure parse traps (element rows reading `exact` with empty units; CAS shared across phases; species existing only as `(g, cis)`), plus a new build-free CI job so they run on every PR
- Catch2 `[formation]`: known values, absent-fluid throws, mixture throws, and a magnitude guard across all 76 fluids — 84 assertions
- `~[slow]`: 380 passed / 2 failed, both failures verified pre-existing on the base commit
- `./dev/ci/preflight.sh`: 6 passed / 2 failed / 2 skipped. The two failures are `cppcheck` and `clang-tidy`, which are file-scoped; parsing all 136 `file:line` findings and intersecting them against every line this branch adds gives **zero** attributable to this work.

Closes the Tier 1 portion of #1360.

## Summary by CodeRabbit

* **New Features**
  * Added standard molar enthalpy of formation data at 298.15 K for many fluids.
  * Exposed formation enthalpy through the public property interface, including source, uncertainty, and reference metadata.
  * Added validation checkpoints for fluid saturation properties.

* **Documentation**
  * Added guidance for regenerating, auditing, and updating thermochemical data.

* **Tests**
  * Added comprehensive offline parsing, coverage, data-validation, and API tests.
  * Added automated checks for the thermochemical data workflow.


ibell:
Thanks — worked through all five findings. Two fixed in 255d83a8, three answered below.

## Fixed

**CodeQL, `dev/atct/test_fetch_atct_formation.py`** — both valid. Dropped the unused `BindResult` import, and scoped the bare module import into the two tests that monkeypatch `__file__`, so the module is no longer imported at module level via both `import` and `from … import`. The local-import form matches those functions' existing `import json as _json`. 29 tests still pass.

## Not applicable — Tabular cache version bump (`DataStructures.h:87`, Major)

This one I checked carefully rather than dismissing, because it is exactly the failure mode I *did* find in the SVD cache and fix in this PR (`kRevision` 19→20). But the Tabular caches do not have it: **no Tabular structure persists a `parameters` value.**

- `PackablePhaseEnvelopeData` (`:117`) and `PureFluidSaturationTableData` (`:244`) have no `CoolProp::parameters` member at all.
- `SinglePhaseGriddedTableData` (`:592`) *does* declare `CoolProp::parameters xkey, ykey` — but its pack list deliberately excludes them:

  ```cpp
  MSGPACK_DEFINE(revision, matrices, xmin, xmax, ymin, ymax);   // TabularBackends.h:618
  ```

- `xkey`/`ykey` are assigned unconditionally in code per table type (`:804` `iHmolar`/`iP`, `:854` `iT`/`iP`), never read back from a cache.
- There is no `static_cast<CoolProp::parameters>` anywhere under `src/Backends/Tabular` — no enum integer is ever decoded from a Tabular blob.

So an old TTSE/BICUBIC cache cannot misdecode a renumbered enum, and there is nothing to version. The contrast with the SVD cache is the point: `SVDSurfaceSerializer.cpp` *does* pack the enum as raw `int32` (`:142` write, `:337` read), which is why that one needed the revision bump.

## Out of scope — duplicate `check_points` sample (`CarbonDioxide.json`, also `SulfurHexafluoride.json`)

Correct observation, but pre-existing upstream data, not introduced here. The `SUPERANCILLARY.check_points` arrays are byte-identical to `master`; this PR's `chore(fluids): canonicalize JSON key order…` commit only reorders keys (every file verified parse-identical to its parent). The duplicate 216.592 K sample simply became visible because reordering pulled those lines into the diff.

Filed separately rather than fixed here, to keep the canonicalization commit provably value-preserving.

## Already done — run `[formation]` and preflight (`CoolProp-Tests.cpp:7310`)

- `[formation]`: passing, 84 assertions, plausibility section checks all 76 fluids
- `~[slow]`: 380 passed / 2 failed, both verified pre-existing on the base commit
- `./dev/ci/preflight.sh`: 6 passed / 2 failed / 2 skipped

The two preflight failures are `cppcheck` and `clang-tidy`, which are file-scoped rather than line-scoped. Parsing all 136 `file:line` findings from both logs and intersecting them against every line this branch adds gives **zero** attributable to this work, and each flagged line is byte-identical to `origin/master`.

Issue #1125, opened 2016-06-15
Mixtures in PropsSI with cubic EOS


JonWel:
Formulations like this one crash with cubic (but works with HEOS or REFPROP):
```Julia
PropsSI("Dmolar", "T", 300, "P", 1e6, "PR::R125[0.17956]&R134a[0.43933]&R32[0.38111]")
```
Here are the log with `debug = 100`:
```
M:\Documents\GitHub\CoolProp\src\CoolProp.cpp:106: backend extracted. backend: PR. fluid: R125[0.17956]&R134a[0.43933]&R32[0.38111]
M:\Documents\GitHub\CoolProp\src\CoolProp.cpp:158: Detected fractions of [    0.180,    0.439,    0.381 ] for R125&R134a&R32.AbstractState::factory(PR,[ R125, R134a, R32 ])
M:\Documents\GitHub\CoolProp\src\Backends\Cubics\CubicBackend.cpp (148): update called with (9: (PT_INPUTS), 1e+06, 300)
```
And a resume of the error thrown:
```c++
Exception: EXCEPTION_ACCESS_VIOLATION at 0x7fece7d2724 -- unknown function (ip: 000007FECE7D2724)
```
To my understanding, the molar fractions are not initialized correctly.
The `State->get_mole_fractions().empty()` returns `0` (instead of `1` for HEOS and REFPROP) at the test [just before it should be initialized in the PropsSI procedure](https://github.com/CoolProp/CoolProp/blob/master/src/CoolProp.cpp#L237).
And in deed, the `mole_fractions` seems to be filled with `0` at that point, so not empty as with the other backends.

I however did not understood how and why this vector is initialized with zeros.
(I can shortcut that behavior by  adding `std::vector<CoolPropDbl> mole_fractions;` in [the protected part](https://github.com/CoolProp/CoolProp/blob/master/src/Backends/Cubics/CubicBackend.h#L32) of the `AbstractCubicBackend`. This is however strange for me, the `AbstractCubicBackend` is derived from `HelmholtzEOSMixtureBackend ` and should inherit the exact same definition as it is in it's protected part (and I copied the line from there) ).

JonWel:
@ibell Adding `std::vector<CoolPropDbl> mole_fractions;` in [the protected part](https://github.com/CoolProp/CoolProp/blob/master/src/Backends/Cubics/CubicBackend.h#L32) of the `AbstractCubicBackend` makes it.
[...]

Issue #1119, opened 2016-06-09
ValueError: _phase is unknown


gilbertsmink:
The second part in the following code does not work for me:
p = 2e5 gives "ValueError: _phase is unknown"

Or shouldn't I use the combination D and UMASS as input properties?

```
fluid = 'Hydrogen'
m = 1
T = 20+273

p = 1e5
u = CP.PropsSI('UMASS', 'P', p, 'T', T, fluid)
D = CP.PropsSI('D', 'P', p, 'T', T, fluid)
h = CP.PropsSI('HMASS', 'P', p, 'T', T, fluid)
# below does not work
p = CP.PropsSI('P', 'D', D, 'UMASS', u, fluid)

p = 2e5
u = CP.PropsSI('UMASS', 'P', p, 'T', T, fluid)
D = CP.PropsSI('D', 'P', p, 'T', T, fluid)
h = CP.PropsSI('HMASS', 'P', p, 'T', T, fluid)
# below does not work
p = CP.PropsSI('P', 'D', D, 'UMASS', u, fluid)
```

ibell:
That combination should be fine, but hydrogen is a nasty fluid.  The
consistency plots look pretty much fine:
http://www.coolprop.org/dev/fluid_properties/fluids/Hydrogen.html#consistency-plots
.  Where is that UD input located?

gilbertsmink:
Im trying to model the emptying of a tube trailer, filled with Hydrogen:

Might be I am making a essential thinking mistake in this way of solving?

```
##
T_tube = 20+273                                                         # [K] initial tubetrailer temperature
M_tube = 720                                                            # [kg] initial tubetrailer capacity
p_tube = 25.0e6                                                         # [Pa] initial tubetrailer pressure
D_tube = CP.PropsSI('D', 'P', p_tube, 'T', T_tube, fluid)               # [kg/m3] initial tubetrailer density
V_tube = M_tube / D_tube                                                # [m3] constant tube trailer volume  
# U_tube = M_tube * CP.PropsSI('UMASS', 'P', p_tube, 'T', T_tube, fluid)  # [J/kg] initial tubetrailer internal enerhy
s_tube = CP.PropsSI('SMASS', 'P', p_tube, 'T', T_tube, fluid) 
S_tube = s_tube*M_tube

#
p_min = 2.0e6               # [Pa] minimum tubetrailer pressure
m_flow = -250/3600          # [kg/s] massflow

t = 0                       # [s] initial time
[...]

PR #3311, opened 2026-08-11
Add modern CMake package support

### LINKED ISSUE #2144: Provide cmake config files to be used by users via find_package(CoolProp)

Tobias-Reiter:
## Summary

- Add an installable and relocatable CMake package
- Provide `CoolProp::CoolProp`, `CoolProp::Static`, and `CoolProp::Shared`
- Support building static and shared libraries simultaneously
- Preserve compatibility with existing single-library builds
- Avoid modifying a parent project's install prefix
- Add downstream C and C++ consumer tests
- Extend CI coverage for installed packages and relocated prefixes
- Document the CMake behavior and compatibility changes in the changelog

## Validation

- Built static and shared libraries together in MSVC Debug with
  `COOLPROP_MSVC_DEBUG=OFF`; verified `/MDd` for Static and `/MTd` for Shared
- Installed the Debug package and compiled/linked downstream C and C++
  consumers through the canonical, Static, and Shared targets
- Ran the installed shared-library C API smoke test successfully
- Verified mandatory Eigen and fmt license installation
- Verified that a shared-only package does not impose Threads on consumers,
  while the static export retains its link dependency
- Verified default and explicitly disabled `CPM_SOURCE_CACHE` behavior
- Parsed the updated workflow YAML and ran clang-format 18.1.8 plus
  `git diff --check`
- Confirmed that this PR contains no teqp/autodiff additions

## AI assistance and verification process

OpenAI Codex materially assisted with the CMake package design, implementation,
consumer fixtures, CI changes, documentation, and review follow-ups. Codex and
the repository's automated tooling performed the local verification listed
above in the contributor's workspace. Tobias Reiter remains responsible for
the submitted changes and has confirmed the project CLA. Every AI-shaped
commit also carries the required `Co-authored-by` trailer.

## CI note

The GUI workflow is path-triggered by changes to `CMakeLists.txt` and
`cmake/**`, so it runs for this CMake PR. Fork pull requests cannot receive the
updater-signing secrets: the previous jobs built the installers successfully
and then failed while signing updater artifacts. The workflow change disables
only updater artifacts for pull requests; trusted push and tag builds retain
the signed updater path.

Closes #2144

## Summary by CodeRabbit

* **New Features**
  * Added relocatable CMake packages for installed CoolProp libraries.
  * Added flexible static and shared builds with selectable linkage targets.
  * Added source-tree and installed-package integration for C and C++ projects.
  * Improved Windows build support, including calling conventions and architecture-specific configurations.

* **Documentation**
  * Added CMake build, installation, integration, and linking guidance.
  * Documented the CMake 3.15 requirement and updated packaging behavior.

* **Tests**
  * Added validation for package contents, relocatability, linkage, consumer projects, and C/C++ API usage.

### DIFF
--- a/.github/workflows/library_shared.yml
+++ b/.github/workflows/library_shared.yml
@@ -1,5 +1,8 @@
 name: CMake library builds
 
+permissions:
+  contents: read
+
 on:
   push:
     branches: [ 'master', 'main', 'develop', 'actions_shared' ]
--- a/.github/workflows/library_shared.yml
+++ b/.github/workflows/library_shared.yml
@@ -208,17 +208,38 @@ jobs:
         set -euo pipefail
 
         python - <<'PY'
+        import re
+        import xml.etree.ElementTree as ET
         from pathlib import Path
 
         expected = {
             "CoolProp_static.vcxproj": "MultiThreadedDebugDLL",
             "CoolProp_shared.vcxproj": "MultiThreadedDebug",
         }
+        namespace = {"msbuild": "http://schemas.microsoft.com/developer/msbuild/2003"}
+        debug_condition = re.compile(
+            r"\s*'\$\(Configuration\)\|\$\(Platform\)'\s*==\s*'Debug\|([^']+)'\s*"
+        )
         for project, runtime in expected.items():
-            content = (Path("build") / project).read_text(encoding="utf-8-sig")
-            setting = f"<RuntimeLibrary>{runtime}</RuntimeLibrary>"
-            if setting not in content:
-                raise SystemExit(f"{project} does not contain {setting}")
+            project_path = Path("build") / project
+            root = ET.parse(project_path).getroot()
+            debug_groups = []
+            for group in root.findall("msbuild:ItemDefinitionGroup", namespace):
+                match = debug_condition.fullmatch(group.get("Condition", ""))
+                if match:
+                    debug_groups.append((group, match.group(1)))
+            if not debug_groups:
+                raise SystemExit(f"{project} has no Debug ItemDefinitionGroup")
+            for group, platform in debug_groups:
+                setting = group.find(
+                    "msbuild:ClCompile/msbuild:RuntimeLibrary", namespace
+                )
+                actual = None if setting is None else setting.text
+                if actual != runtime:
+                    raise SystemExit(
+                        f"{project} Debug|{platform} runtime is {actual!r}; "
+                        f"expected {runtime!r}"
+                    )
         PY
 
         cmake --build build --config Debug --parallel 1
--- a/.github/workflows/library_shared.yml
+++ b/.github/workflows/library_shared.yml
@@ -1,4 +1,4 @@
-name: Library builds (shared)
+name: CMake library builds
 
 on:
   push:
@@ -28,6 +28,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [windows-latest, ubuntu-latest, macOS-latest]
+        linkage: [shared, static]
+        include:
+          - os: ubuntu-latest
+            linkage: both
+          - os: windows-latest
+            linkage: both
 
     steps:
     - uses: actions/checkout@v7
@@ -48,8 +54,7 @@ jobs:
         COOLPROP_VERSION=$(python dev/extract_version.py --cmake-only)
         echo COOLPROP_VERSION=$COOLPROP_VERSION >> $GITHUB_ENV
 
-    # cmake/dependencies.cmake fetches its CPM deps from network with no
-    # cache, so every OS leg here re-downloads the same dependency set.
+    # Reuse the pinned CPM dependency sources across builds on each OS.
     - name: Cache CPM packages
       uses: actions/cache@v6
       with:
@@ -59,7 +64,19 @@ jobs:
           cpm-${{ runner.os }}-
 
     - name: Configure CMake
-      run: cmake -B build -S . -DCMAKE_BUILD_TYPE:STRING=${{ env.BUILD_TYPE }} -DCOOLPROP_SHARED_LIBRARY:BOOL=ON
+      shell: bash
+      run: |
+        static=OFF
+        shared=OFF
+        case "${{ matrix.linkage }}" in
+          static) static=ON ;;
+          shared) shared=ON ;;
+          both) static=ON; shared=ON ;;
+        esac
+        cmake -B build -S . \
+          -DCMAKE_BUILD_TYPE:STRING=${{ env.BUILD_TYPE }} \
+          -DCOOLPROP_STATIC_LIBRARY:BOOL="$static" \
+          -DCOOLPROP_SHARED_LIBRARY:BOOL="$shared"
 
     - name: Build
       shell: bash
@@ -70,6 +87,11 @@ jobs:
       # under bash on all three runners. Mirrors csharp_builder.yml.
       run: |
         JOBS=$(nproc 2>/dev/null || sysctl -n hw.logicalcpu 2>/dev/null || echo 2)
+        if [ "${{ runner.os }}" = Windows ] && [ "${{ matrix.linkage }}" = both ]; then
+          # Each MSVC project already uses /MP. Building both projects in
+          # parallel can exhaust the compiler heap on hosted runners.
+          JOBS=1
+        fi
         echo "Building with -j$JOBS"
         cmake --build build --target install -j "$JOBS" --config ${{ env.BUILD_TYPE }}
 
@@ -78,7 +100,7 @@ jobs:
       # dynamic export table. Rides the shared build produced above (ELF on
       # ubuntu, Mach-O on macOS) — no separate build. nm-based, so the Windows
       # PE legs are skipped (exports are opt-in via src/CoolPropLib.def).
-      if: runner.os == 'Linux' || runner.os == 'macOS'
+      if: matrix.linkage != 'static' && (runner.os == 'Linux' || runner.os == 'macOS')
       shell: bash
       run: |
         set -euo pipefail
@@ -93,42 +115,165 @@ jobs:
         echo "Gating $lib"
         ./dev/ci/check-json-symbols.sh "$lib"
 
+    - name: Test installed CMake package
+      shell: bash
+      run: |
+        set -euo pipefail
+        prefix_a="$RUNNER_TEMP/coolprop-prefix-a-${{ matrix.linkage }}"
+        prefix_b="$RUNNER_TEMP/coolprop-prefix-b-${{ matrix.linkage }}"
+        cmake -E remove_directory "$prefix_a"
+        cmake -E remove_directory "$prefix_b"
+        cmake --install build --config "${{ env.BUILD_TYPE }}" --prefix "$prefix_a"
+        cmake -E rename "$prefix_a" "$prefix_b"
+
+        python - "$prefix_b" "$GITHUB_WORKSPACE" "$prefix_a" <<'PY'
+        from pathlib import Path
+        import sys
+
+        package_dir = Path(sys.argv[1]) / "lib" / "cmake" / "CoolProp"
+        required_files = {
+            package_dir / "CoolPropConfig.cmake",
+            package_dir / "CoolPropTargets.cmake",
+        }
+        missing = sorted(path for path in required_files if not path.is_file())
+        if missing:
+            raise SystemExit(f"CMake package files are missing: {missing}")
+
+        cmake_files = list(package_dir.glob("*.cmake"))
+        if not cmake_files:
+            raise SystemExit(f"No CMake package files found under {package_dir}")
+
+        forbidden = {
+            value.replace("\\", "/")
+            for value in (sys.argv[2], sys.argv[3])
+        }
+        for cmake_file in cmake_files:
+            content = cmake_file.read_text(encoding="utf-8").replace("\\", "/")
+            leaked = [value for value in forbidden if value and value in content]
+            if leaked:
+                raise SystemExit(f"{cmake_file} contains non-relocatable paths: {leaked}")
+        PY
+
+        expected=""
+        case "${{ matrix.linkage }}" in
+          static) expected=STATIC ;;
+          shared) expected=SHARED ;;
+        esac
+        cmake -S dev/ci/cmake-consumer -B consumer-build \
+          -DCMAKE_PREFIX_PATH="$prefix_b" \
+          -DCOOLPROP_EXPECTED_LINKAGE="$expected"
+        cmake --build consumer-build --config "${{ env.BUILD_TYPE }}"
+        ctest --test-dir consumer-build -C "${{ env.BUILD_TYPE }}" --output-on-failure
+
+        if [ "${{ matrix.linkage }}" != static ]; then
+          cmake -S dev/ci/cmake-consumer -B c-consumer-build \
+            -DCMAKE_PREFIX_PATH="$prefix_b" \
+            -DCOOLPROP_EXPECTED_LINKAGE=SHARED \
+            -DCOOLPROP_C_ONLY=ON
+          cmake --build c-consumer-build --config "${{ env.BUILD_TYPE }}"
+          ctest --test-dir c-consumer-build -C "${{ env.BUILD_TYPE }}" --output-on-failure
+        fi
+
+    - name: Test add_subdirectory consumer
+      if: runner.os == 'Linux' && matrix.linkage == 'static'
+      shell: bash
+      run: |
+        set -euo pipefail
+        cmake -S dev/ci/cmake-consumer -B source-consumer \
+          -DCOOLPROP_SOURCE_DIR="$GITHUB_WORKSPACE" \
+          -DCOOLPROP_EXPECTED_LINKAGE=STATIC
+        cmake --build source-consumer --config "${{ env.BUILD_TYPE }}"
+        ctest --test-dir source-consumer -C "${{ env.BUILD_TYPE }}" --output-on-failure
+
+    - name: Configure add_subdirectory consumer with CMP0091-old parent
+      if: runner.os == 'Windows' && matrix.linkage == 'static'
+      shell: bash
+      run: |
+        set -euo pipefail
+        cmake -S dev/ci/cmake-consumer -B source-consumer-windows \
+          -DCOOLPROP_SOURCE_DIR="$GITHUB_WORKSPACE" \
+          -DCOOLPROP_EXPECTED_LINKAGE=STATIC
+
     # Create a special case for Windows since we also need 2 flavours of 32bit binaries
     - name: Configure CMake for 32bit Windows stdcall
-      if: startsWith(matrix.os, 'windows')
+      if: startsWith(matrix.os, 'windows') && matrix.linkage == 'shared'
       run: cmake -B build_stdc  -S . -DCMAKE_BUILD_TYPE:STRING=${{ env.BUILD_TYPE }} -DCOOLPROP_SHARED_LIBRARY:BOOL=ON -DCOOLPROP_STDCALL_LIBRARY:BOOL=ON -A Win32
       
     - name: Build with CMake for 32bit Windows stdcall
-      if: startsWith(matrix.os, 'windows')
+      if: startsWith(matrix.os, 'windows') && matrix.linkage == 'shared'
       shell: bash
       run: |
         JOBS=$(nproc 2>/dev/null || sysctl -n hw.logicalcpu 2>/dev/null || echo 2)
         echo "Building with -j$JOBS"
         cmake --build build_stdc --target install -j "$JOBS" --config ${{ env.BUILD_TYPE }}
 
+    - name: Test 32bit Windows stdcall package
+      if: startsWith(matrix.os, 'windows') && matrix.linkage == 'shared'
+      shell: bash
+      run: |
+        set -euo pipefail
+        prefix="$RUNNER_TEMP/coolprop-stdcall"
+        cmake -E remove_directory "$prefix"
+        cmake --install build_stdc --config "${{ env.BUILD_TYPE }}" --prefix "$prefix"
+        cmake -S dev/ci/cmake-consumer -B consumer-stdcall -A Win32 \
+          -DCMAKE_PREFIX_PATH="$prefix" \
+          -DCOOLPROP_EXPECTED_LINKAGE=SHARED \
+          -DCOOLPROP_C_ONLY=ON
+        cmake --build consumer-stdcall --config "${{ env.BUILD_TYPE }}"
+        ctest --test-dir consumer-stdcall -C "${{ env.BUILD_TYPE }}" --output-on-failure
+
     - name: Configure CMake for 32bit Windows cdecl
-      if: startsWith(matrix.os, 'windows')
+      if: startsWith(matrix.os, 'windows') && matrix.linkage == 'shared'
       run: cmake -B build_cdecl -S . -DCMAKE_BUILD_TYPE:STRING=${{ env.BUILD_TYPE }} -DCOOLPROP_SHARED_LIBRARY:BOOL=ON -DCOOLPROP_CDECL_LIBRARY:BOOL=ON   -A Win32
       
     - name: Build with CMake for 32bit Windows cdecl
-      if: startsWith(matrix.os, 'windows')
+      if: startsWith(matrix.os, 'windows') && matrix.linkage == 'shared'
       shell: bash
       run: |
         JOBS=$(nproc 2>/dev/null || sysctl -n hw.logicalcpu 2>/dev/null || echo 2)
         echo "Building with -j$JOBS"
         cmake --build build_cdecl --target install -j "$JOBS" --config ${{ env.BUILD_TYPE }}
 
+    - name: Test 32bit Windows cdecl package
+      if: startsWith(matrix.os, 'windows') && matrix.linkage == 'shared'
+      shell: bash
+      run: |
+        set -euo pipefail
+        prefix="$RUNNER_TEMP/coolprop-cdecl"
+        cmake -E remove_directory "$prefix"
+        cmake --install build_cdecl --config "${{ env.BUILD_TYPE }}" --prefix "$prefix"
+        cmake -S dev/ci/cmake-consumer -B consumer-cdecl -A Win32 \
+          -DCMAKE_PREFIX_PATH="$prefix" \
+          -DCOOLPROP_EXPECTED_LINKAGE=SHARED \
+          -DCOOLPROP_C_ONLY=ON
+        cmake --build consumer-cdecl --config "${{ env.BUILD_TYPE }}"
+        ctest --test-dir consumer-cdecl -C "${{ env.BUILD_TYPE }}" --output-on-failure
+
     - name: Configure CMake for 64bit Windows arm64
-      if: startsWith(matrix.os, 'windows')
+      if: startsWith(matrix.os, 'windows') && matrix.linkage == 'shared'
       run: cmake -B build_arm64 -S . -DCMAKE_BUILD_TYPE:STRING=${{ env.BUILD_TYPE }} -DCOOLPROP_SHARED_LIBRARY:BOOL=ON -A ARM64
       
     - name: Build with CMake for 64bit Windows arm64
-      if: startsWith(matrix.os, 'windows')
+      if: startsWith(matrix.os, 'windows') && matrix.linkage == 'shared'
       shell: bash
       run: |
         JOBS=$(nproc 2>/dev/null || sysctl -n hw.logicalcpu 2>/dev/null || echo 2)
         echo "Building with -j$JOBS"
         cmake --build build_arm64 --target install -j "$JOBS" --config ${{ env.BUILD_TYPE }}
+
+    - name: Link CMake consumer for Windows arm64
+      if: startsWith(matrix.os, 'windows') && matrix.linkage == 'shared'
+      shell: bash
+      run: |
+        set -euo pipefail
+        prefix="$RUNNER_TEMP/coolprop-arm64"
+        cmake -E remove_directory "$prefix"
+        cmake --install build_arm64 --config "${{ env.BUILD_TYPE }}" --prefix "$prefix"
+        cmake -S dev/ci/cmake-consumer -B consumer-arm64 -A ARM64 \
+          -DCMAKE_PREFIX_PATH="$prefix" \
+          -DCOOLPROP_EXPECTED_LINKAGE=SHARED \
+          -DCOOLPROP_C_ONLY=ON
+        cmake --build consumer-arm64 --config "${{ env.BUILD_TYPE }}"
 
 
     # - name: Tar.gz the shared library to maintain case sensitivy and file permissions
@@ -139,10 +284,11 @@ jobs:
     #     tar -cvzf CoolProp-${{ env.COOLPROP_VERSION }}-shared-${{ matrix.os }}.tar.gz ./*
 
     - name: Archive artifacts
+      if: matrix.linkage == 'shared'
       uses: actions/upload-artifact@v7
       with:
           # Upload to unique artifact names tagged with suffix of matrix.os
-          # This is now requried by upload-artifact@v4, but will be merged later
+          # This is now required by upload-artifact@v4, but will be merged later
           name: shared_library-${{ matrix.os }}
           path: install_root/shared_library
 
--- a/.github/workflows/library_shared.yml
+++ b/.github/workflows/library_shared.yml
@@ -76,10 +76,17 @@ jobs:
           shared) shared=ON ;;
           both) static=ON; shared=ON ;;
         esac
+        extra_args=()
+        if [ "${{ runner.os }}" = Windows ] && [ "${{ matrix.linkage }}" = both ]; then
+          # Regression gate: Debug must still select the debug CRT even when
+          # the deprecated compatibility switch is explicitly disabled.
+          extra_args+=("-DCOOLPROP_MSVC_DEBUG:BOOL=OFF")
+        fi
         cmake -B build -S . \
           -DCMAKE_BUILD_TYPE:STRING=${{ env.BUILD_TYPE }} \
           -DCOOLPROP_STATIC_LIBRARY:BOOL="$static" \
-          -DCOOLPROP_SHARED_LIBRARY:BOOL="$shared"
+          -DCOOLPROP_SHARED_LIBRARY:BOOL="$shared" \
+          "${extra_args[@]}"
 
     - name: Build
       shell: bash
@@ -134,13 +141,16 @@ jobs:
         import sys
 
         package_dir = Path(sys.argv[1]) / "lib" / "cmake" / "CoolProp"
+        license_dir = Path(sys.argv[1]) / "share" / "licenses" / "CoolProp"
         required_files = {
             package_dir / "CoolPropConfig.cmake",
             package_dir / "CoolPropTargets.cmake",
+            license_dir / "Eigen" / "COPYING.MPL2",
+            license_dir / "fmt" / "LICENSE",
         }
         missing = sorted(path for path in required_files if not path.is_file())
         if missing:
-            raise SystemExit(f"CMake package files are missing: {missing}")
+            raise SystemExit(f"Required installed files are missing: {missing}")
 
         cmake_files = list(package_dir.glob("*.cmake"))
         if not cmake_files:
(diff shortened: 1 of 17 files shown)


Web/coolprop/wrappers/StaticLibrary/index.rst
```diff
@@ -31,13 +31,19 @@ If you are using CMake, the process is quite trivial to integrate the CoolProp s
...
+    add_subdirectory("${CMAKE_SOURCE_DIR}/externals/CoolProp" CoolProp)
+    add_executable(main "${CMAKE_SOURCE_DIR}/mycode.cpp")
+    target_link_libraries(main PRIVATE CoolProp::CoolProp)
+
+An already-installed CoolProp package can instead be used without embedding
+its source tree::
+
+    find_package(CoolProp 8 CONFIG REQUIRED COMPONENTS Static)
+    add_executable(main mycode.cpp)
+    target_link_libraries(main PRIVATE CoolProp::Static)
```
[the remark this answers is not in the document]
ibell: Is this fixed now?

CMakeLists.txt
```diff
@@ -1857,14 +1579,14 @@ if(COOLPROP_JAVA_MODULE)
     # the DIRECTORY form above would install the entire binary dir.
     install(
       CODE "file( GLOB _GeneratedJavaSources \"${CMAKE_CURRENT_BINARY_DIR}/*.java\" )"
-      CODE "file( INSTALL \${_GeneratedJavaSources} DESTINATION ${CMAKE_INSTALL_PREFIX}/Java/platform-independent )"
+      CODE "file( INSTALL \${_GeneratedJavaSources} DESTINATION \${CMAKE_INSTALL_PREFIX}/Java/platform-independent )"
     )
```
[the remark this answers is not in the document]
Tobias-Reiter: Thanks for flagging this. `file(INSTALL)` already applies `DESTDIR` internally via `cmFileInstaller::HandleInstallDestination()` ([CMake source, lines 380–436](https://github.com/Kitware/CMake/blob/v4.2.3/Source/cmFileInstaller.cxx#L380-L436)). `${CMAKE_INSTALL_PREFIX}` is intentionally deferred until install time so `cmake --install --prefix` is honored. Adding `$ENV{DESTDIR}` explicitly here would cause the staging prefix to be applied twice, so this should remain unchanged.

include/CoolProp/CoolPropLib.h
```diff
@@ -35,42 +37,50 @@
...
 #endif
 
 #ifndef __cplusplus
-#    if defined(__STDC_VERSION__)
-#        if (__STDC_VERSION__ >= 199901L)
-#            include <stdbool.h>
-#        endif
-#    endif
+#    include <stdbool.h>
 #endif
```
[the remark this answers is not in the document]
Tobias-Reiter: This is addressed by documenting the restriction in both `README.md` and the shared-library documentation. Separate consumer targets may select different variants, but one final binary must not link both. This is intentionally unsupported on Windows because the shared target propagates DLL-import semantics, so adding a mixed-link consumer test would test an unsupported configuration.

Tobias-Reiter:
Review follow-up: the GUI workflow change is directly tied to this CMake PR because that workflow is path-triggered by `CMakeLists.txt`, `cmake/**`, and `include/**`. The previous fork run (#1119) built the installers on all platforms and then failed only because updater-signing secrets are unavailable to fork pull requests. The PR-only path now omits updater artifacts, while trusted push/tag/manual builds retain signing; run #1125 passes on Windows, Linux, and macOS. I added this rationale to the PR description.

The remaining CodeRabbit threads were also reviewed: `file(INSTALL)` already applies `DESTDIR` internally, and the unsupported mixed Static/Shared final-binary case is documented. Both threads now contain the technical rationale and are resolved; no additional code changes were needed.

ibell (changes requested):
Thanks for this — the package layout, the relocatability check, and the downstream consumer tests are all genuinely good, and `find_package(CoolProp)` with real imported targets is long overdue (#2144).

Requesting changes on two silent MSVC regressions, one fail-open license install, and the fact that **no changelog entry accompanies this**. `Web/coolprop/changelog.rst` has both a top-level `Breaking Changes:` section and a per-release `**Behavior changes (potentially breaking):**` block, and neither is touched here. The PR description says "Preserve compatibility with existing single-library builds" — that holds for the *target name* (`target_link_libraries(main CoolProp)` still resolves), but not for several other axes listed below.

---

## Must fix

### 1. MSVC Debug configurations silently switch to the release CRT

`cmake/CoolPropLibrary.cmake`, `_coolprop_set_msvc_runtime()` picks `MultiThreaded` (shared) / `MultiThreadedDLL` (static) and only adds the `$<$<CONFIG:Debug>:Debug>` variant when `COOLPROP_MSVC_DEBUG` is ON.

The old path was `modify_msvc_flags("/MT")` (shared) / `("/MD")` (static), and `modify_msvc_flags` sets release to `flag_default` and debug to `flag_default` **+ `"d"`** whenever `COOLPROP_MSVC_REL`/`COOLPROP_MSVC_DBG` are `IGNORE` — which is the default (`CMakeLists.txt:308`).

| build | old Release | old Debug | new Release | new Debug |
|---|---|---|---|---|
| shared | `/MT` | `/MTd` | `/MT` | **`/MT`** |
| static | `/MD` | `/MDd` | `/MD` | **`/MD`** |

`DEBUG_POSTFIX d` is still applied, so `CoolPropd.dll` / `CoolPropd.lib` become release-CRT artifacts under a debug name. Consumers hit `LNK2038: mismatch detected for '_ITERATOR_DEBUG_LEVEL'`, or a cross-CRT heap if they force it through. This also inverts the meaning of `COOLPROP_MSVC_DEBUG`: it used to mean "also use the debug CRT when STATIC/DYNAMIC is pinned"; it is now the only way to get a debug CRT at all.

Suggested fix: emit `MultiThreaded$<$<CONFIG:Debug>:Debug>[DLL]` unconditionally, and keep `COOLPROP_MSVC_DEBUG` only as a back-compat no-op.

The CI matrix builds `Release` only, so nothing here catches it.

### 2. `cmake_policy(SET CMP0091 NEW)` fail-opens the surviving `modify_msvc_flags()` call sites

Three wrapper modules still call it — C# (`CMakeLists.txt:1559`), Java (`CMakeLists.txt:1805`), Mathematica (`CMakeLists.txt:2152`). The macro works by `string(REPLACE)`-ing `/MD` out of `CMAKE_CXX_FLAGS_<CONFIG>` and appending `/MT`. Under CMP0091 NEW, CMake no longer puts the runtime flag in those variables — it injects `/MD` from the default `MSVC_RUNTIME_LIBRARY` instead. So the *removal* half becomes a no-op and the compile line carries both `/MD` and `/MT`.

Those wrapper DLLs are deliberately `/MT` so C#/Java users don't need the VC redist. Nothing errors; this is exactly the "safeguard that silently no-ops" class. Please convert those three sites to `set_property(TARGET ... PROPERTY MSVC_RUNTIME_LIBRARY ...)` in the same pass that introduces the policy.

### 3. License installs fail open

```cmake
install(FILES "${Eigen_SOURCE_DIR}/COPYING.MPL2" ... OPTIONAL)
install(FILES "${fmt_SOURCE_DIR}/LICENSE"      ... OPTIONAL)
```

`_coolprop_install_standard_headers()` now *redistributes* Eigen (MPL-2.0) and fmt (MIT) inside every installed package. With `OPTIONAL`, a trimmed CPM cache that lacks either file ships the headers with no license text and the build still succeeds. Drop `OPTIONAL` so a missing license is a hard failure.

---

## Needs a changelog entry (behavior changes, potentially breaking)

Public header (`include/CoolProp/CoolPropLib.h`):

* **C89 consumers stop compiling.** `#include <stdbool.h>` is now unconditional in C mode; it was guarded by `__STDC_VERSION__ >= 199901L`. MSVC before VS2013 in C mode, and any C89 toolchain, now fail — on the header this PR advertises as the portable shared-library interface.
* **`EXTERNC` / `__powerpc__` no longer override `EXPORT_CODE`.** The old block did `#undef EXPORT_CODE` / `#define EXPORT_CODE extern "C"`, stripping `__declspec(dllexport)` *and* clobbering a user-supplied `EXPORT_CODE`. Now they compose under `#ifndef EXPORT_CODE`. Two consequences: on Windows, `-DCOOLPROP_EXTERNC_LIBRARY=ON` shared builds now emit `dllexport` where they previously relied on the `.def` list, so **the DLL export set changes**; and a consumer pre-defining both `EXPORT_CODE` and `EXTERNC` now keeps their own definition, reversing the old precedence. `wrappers/Fortran/detailed_example/CMakeLists.txt:8` sets `COOLPROP_EXTERNC_LIBRARY`, and the option also moved from target-private `COMPILE_FLAGS` to `PUBLIC`, so consumers now inherit `-DEXTERNC`.
* **Windows consumers of the new shared target get `__declspec(dllimport)` + `__stdcall` they never got before**, via the `INTERFACE COOLPROP_SHARED_LIBRARY_USE` definition. No pre-PR consumer path produced `dllimport`. Code mixing hand-written prototypes with the header, or built against a `.def`-limited `COOLPROP_LIBRARY_EXPORTS`, can now fail to link.
* **`cxx_std_17` is propagated `PUBLIC`.** Consumers are force-promoted to C++17 where they previously inherited nothing (CoolProp only set `CMAKE_CXX_STANDARD 17` for itself). A C++11/14 downstream is either silently bumped — which can break *their* code (`auto_ptr`, `register`, dynamic exception specs) — or errors out on a pinned lower standard.

Build/packaging:

* **`add_subdirectory`/FetchContent consumers lose all install rules** — `COOLPROP_INSTALL_CMAKE_PACKAGE` and `COOLPROP_INSTALL_LEGACY_LAYOUT` both default OFF when not top level. Defensible and probably right, but silent for anyone relying on it.
* **`add_subdirectory` consumers lose most of the propagated include path.** Was `PUBLIC ${APP_INCLUDE_DIRS}` (repo root, `include/`, `src/`, `dev/`, Eigen, fmt, msgpack, nlohmann, valijson, boost, IF97, REFPROP headers, incbin, miniz — `CMakeLists.txt:370-393`). Now `PUBLIC include/` plus the Eigen/fmt interface targets; the rest is `PRIVATE`.
* **Vendored Eigen + fmt land on every consumer's include path** via `PUBLIC`-linked SYSTEM interface targets pointing at `<prefix>/include/CoolProp/third_party/{eigen,fmt}`. A consumer with its own fmt now has two copies visible with order-dependent resolution between two SYSTEM dirs — a real version-skew/ODR hazard worth calling out explicitly in the docs, given the effort v8 spent making the JSON libs non-clashing.
* **`CMAKE_INSTALL_PREFIX` is now honored for top-level builds** (previously FORCEd to `<src>/install_root` unconditionally). Correct, but a script that passes a prefix *and* then reads `install_root/...` now finds nothing.
* **Default `install_root` contents change** — with `COOLPROP_INSTALL_CMAKE_PACKAGE` ON at top level it gains `bin/`, `lib/`, `lib/cmake/CoolProp/`, `include/` (tens of MB of Eigen+fmt) and `share/licenses/` alongside `shared_library/`.
* **MSVC CRT selection is no longer global** — `COOLPROP_MSVC_STATIC`/`COOLPROP_MSVC_DYNAMIC` previously rewrote `CMAKE_{C,CXX}_FLAGS_<CONFIG>` for *every* target in the build tree; now only the CoolProp library targets get `MSVC_RUNTIME_LIBRARY`.
* **New hard `Threads` dependency** (`find_package(Threads REQUIRED)` + `find_dependency(Threads)` in the installed config). Linked `PRIVATE`, so it still propagates as `$<LINK_ONLY:>` for the static archive. Cross/bare-metal toolchains with no thread model that previously configured now fail.
* **`cmake_minimum_required` 3.14 -> 3.15.**

---

## Smaller / please confirm

* **CI job identity changes.** The workflow display name goes `Library builds (shared)` -> `CMake library builds`, and the matrix jobs gain a linkage axis (`build (ubuntu-latest)` -> `build (ubuntu-latest, shared)`). If either is configured as a required status check, open PRs will block on a check that never reports. `release_all_files.yml` refers to the workflow by *filename*, so that path is fine. Worth verifying branch protection before merge.
* **Linux real-name changes on pre-release builds.** `VERSION` moved from `COOLPROP_VERSION` (`8.0.1dev`) to `COOLPROP_VERSION_NUMERIC` (`8.0.1`), so `libCoolProp.so.8.0.1` instead of `libCoolProp.so.8.0.1dev`. Identical for final releases, but the Debian branch (`CMakeLists.txt:926`) still uses `COOLPROP_VERSION`, so the two now disagree on dev builds. Please make them consistent.
* **Dual builds make `CoolProp` an `INTERFACE` library**, so `cmake --build . --target CoolProp` is no longer valid in that configuration. The Windows-package block (`CMakeLists.txt:1241`) uses exactly that invocation — safe today since it only ever configures single-variant, but it's a trap for later.
* **README/docs wording**: "`CoolProp::CoolProp` selects the value of `COOLPROP_DEFAULT_LIBRARY`" reads like a consumer knob; it's a producer-side choice baked into the installed package. Worth rephrasing.
* `cmake/dependencies.cmake`: `if(NOT DEFINED CPM_SOURCE_CACHE)` -> `if(NOT CPM_SOURCE_CACHE)` **plus `FORCE`** now overwrites a deliberately-falsy cached value with the in-tree default.

---

Happy to help with the changelog wording once the CRT items are sorted.

Tobias-Reiter:
Thanks for the detailed review. I have rebased onto current `master` and pushed
the fixes in `2b90db6a`.

The must-fix items are addressed:

- Debug configurations now always use the matching MSVC Debug CRT (`/MDd` for
  the historical static-library default and `/MTd` for shared). The legacy
  `COOLPROP_MSVC_DEBUG` cache option remains accepted as a compatibility no-op.
- C#, Java, and Mathematica now use the same target-local
  `MSVC_RUNTIME_LIBRARY` helper after their targets are created; no
  `modify_msvc_flags()` call sites remain.
- Eigen and fmt license installs are mandatory, and the package matrix asserts
  that both license files exist.
- The 8.0.1 changelog now documents the CMake minimum-version, target usage
  requirements, install/add_subdirectory behavior, vendored public-header
  dependencies, target-local CRT behavior, Threads dependency, and dual-target
  semantics.

I also incorporated the smaller findings: restored the legacy C89 stdbool
guard while supporting modern MSVC C, restored the prerelease Linux real-name,
made all four Windows installer sub-builds explicitly shared-only, clarified
that `COOLPROP_DEFAULT_LIBRARY` is producer-side, and fixed the CPM include
order so the default cache is established before CPM while explicit OFF/empty
values remain respected. A shared-only installed config no longer imposes
Threads on its consumers; static/dual packages still record the required link
dependency. The repository currently has no enforced required-check contexts,
so the clearer workflow display name does not strand an old required check.

Local validation included a full MSVC Debug static+shared build with
`COOLPROP_MSVC_DEBUG=OFF`, exact generated CRT-property checks, installation,
downstream Debug compilation/linking through `CoolProp::CoolProp`,
`CoolProp::Static`, and `CoolProp::Shared`, a successful shared C-API runtime
smoke, license checks, shared-only Threads-negative configuration, CPM
default/OFF checks, YAML parsing, clang-format 18.1.8, and `git diff --check`.
The updated CI now carries the corresponding regression gates.

The commits and PR description now also include the repository's required AI
assistance attribution. No teqp/autodiff changes are part of this PR.

Could you please re-review the updated branch when convenient?

cmake/CoolPropLibrary.cmake
```diff
@@ -0,0 +1,504 @@
...
+          DESTINATION "${_third_party_dir}/eigen/unsupported/Eigen")
+  install(DIRECTORY "${Eigen_SOURCE_DIR}/unsupported/Eigen/src/Polynomials"
+          DESTINATION "${_third_party_dir}/eigen/unsupported/Eigen/src")
+  install(DIRECTORY "${fmt_SOURCE_DIR}/include/fmt"
+          DESTINATION "${_third_party_dir}/fmt")
+
+  install(FILES "${Eigen_SOURCE_DIR}/COPYING.MPL2"
+          DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/licenses/CoolProp/Eigen")
+  install(FILES "${fmt_SOURCE_DIR}/LICENSE"
+          DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/licenses/CoolProp/fmt")
```
ibell, line 242: **Blocking — this breaks the `dev_checks` installed-header gate and `dev/ci/preflight.sh`.**

`dev_checks.yml` configures a top-level build with `-DCOOLPROP_SHARED_LIBRARY=ON`, so `COOLPROP_INSTALL_CMAKE_PACKAGE` defaults ON and `cmake --install` now writes **two** include roots: the legacy `shared_library/include/` and the new `include/` — the latter containing 361 Eigen `.h` files under `include/CoolProp/third_party/eigen/`.

`dev/ci/check-installed-headers.sh:106` picks its include root with:

```sh
INC_ROOT="$(find "$PREFIX" -path '*/include/CoolProp/CoolProp.h' | head -1)"
```

That is now ambiguous between the two roots, and `find` does not sort — so which one wins is readdir order. When it resolves to `<prefix>/include`, assertion 3 compiles every `*.h` beneath it standalone, including Eigen's `InternalHeaderCheck.h` files, each of which is nothing but:

```c
#ifndef EIGEN_CHOLMODSUPPORT_MODULE_H
#error "Please include Eigen/CholmodSupport instead of including headers inside the src directory directly."
#endif
```

Reproduced locally against a mock prefix built from the pinned Eigen 5.0.1 tarball: `find` returned `include/` first, and

```
printf '#include "CoolProp/third_party/eigen/Eigen/src/CholmodSupport/InternalHeaderCheck.h"\n' \
  | c++ -std=c++17 -fsyntax-only -I<prefix>/include -x c++ -
```

fails with that `#error`. The gate then exits 1 with `installed header is not self-contained`. Several `fmt` headers (`format-inl.h`) are not standalone-compilable either.

Two things worth fixing, since they are independent:
1. **The gate's ambiguity.** `INC_ROOT` should be pinned rather than discovered by `head -1` — otherwise it silently flips between two trees.
2. **The walk's scope.** Whichever root is used, the `find "$INC_ROOT" -name '*.h'` sweep needs to prune `*/third_party/*`; vendored headers are not CoolProp's self-containedness contract.

Because it hinges on directory order this may pass on one runner and fail on the next, which is worse than a deterministic failure — worth fixing even if CI happens to come back green.

CMakeLists.txt
```diff
@@ -1,27 +1,80 @@
...
+      CACHE PATH "CoolProp install path" FORCE)
+elseif(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR
+       AND CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+  # Preserve the historical in-tree release layout for top-level builds, but
+  # never overwrite a parent project's install prefix when used via
+  # add_subdirectory()/FetchContent.
+  set(CMAKE_INSTALL_PREFIX
+      "${CMAKE_CURRENT_SOURCE_DIR}/install_root"
+      CACHE PATH "CoolProp install path" FORCE)
 endif()
```
ibell: The `elseif` branch correctly guards on `CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR`, but the `if(DEFINED COOLPROP_INSTALL_PREFIX)` branch above it does not — it still `FORCE`-overwrites `CMAKE_INSTALL_PREFIX` in a nested build.

That contradicts the changelog entry added in this PR:

> Nested `add_subdirectory` / `FetchContent` builds ... no longer change the parent project's install prefix.

Concretely: a parent project that does `add_subdirectory(CoolProp)` and happens to carry `COOLPROP_INSTALL_PREFIX` in its cache (the legacy CoolProp knob — an older superbuild, or a user who set it once) has its own `CMAKE_INSTALL_PREFIX` silently rewritten, so `cmake --install` writes the *parent's* artifacts under CoolProp's prefix.

The new regression test can't catch it. `dev/ci/cmake-consumer/CMakeLists.txt:62` asserts the prefix is unchanged, but it never defines `COOLPROP_INSTALL_PREFIX`, so on this path the assertion passes vacuously.

Suggest adding the same top-level guard to the first branch (with a `message(WARNING)` when it's ignored in a nested build), and having the consumer fixture set `COOLPROP_INSTALL_PREFIX` so the guard is actually exercised.

cmake/CoolPropLibrary.cmake
```diff
@@ -0,0 +1,504 @@
...
+  # default empty and ignores MSVC_RUNTIME_LIBRARY. Append an explicit
+  # target-local option so add_subdirectory consumers still get the requested
+  # runtime without rewriting the parent's global CMAKE_CXX_FLAGS_* values.
+  if(_coolprop_needs_msvc_runtime_fallback)
+    target_compile_options(
+      "${target}"
+      PRIVATE "$<$<CONFIG:Debug>:${_runtime_flag}d>"
+              "$<$<NOT:$<CONFIG:Debug>>:${_runtime_flag}>")
+  endif()
+endfunction()
```
ibell, line 91: Making the MSVC runtime target-local is the right direction, but it is now applied *only* to the CoolProp library targets — and master's `modify_msvc_flags()` had directory-wide reach, so some in-tree consumers are left behind.

With `-DCOOLPROP_STATIC_LIBRARY=ON -DCOOLPROP_MSVC_STATIC=ON -DCOOLPROP_CPP_EXAMPLE_TEST=ON` on MSVC:

- the `CoolProp` target gets `MSVC_RUNTIME_LIBRARY=MultiThreaded` (`/MT`) from this function;
- `docuTest.exe` (`CMakeLists.txt:2170`) and the `Web/coolprop/snippets` executables (`CMakeLists.txt:2259`) link that library but get no `MSVC_RUNTIME_LIBRARY`, so under CMP0091 NEW they fall back to the default `MultiThreadedDLL` (`/MD`).

Result is `LNK2038: mismatch detected for 'RuntimeLibrary'`. Under master both sides got `/MT`, because `modify_msvc_flags("/MT")` rewrote `CMAKE_CXX_FLAGS_<CONFIG>` for the whole directory.

Either apply `_coolprop_set_msvc_runtime()` to those executables too, or set `CMAKE_MSVC_RUNTIME_LIBRARY` at directory scope so unmarked targets inherit the same choice. Narrow configuration, but it's a hard link error when hit.

.github/workflows/library_shared.yml
```diff
@@ -93,43 +125,211 @@ jobs:
...
+
+        expected = {
+            "CoolProp_static.vcxproj": "MultiThreadedDebugDLL",
+            "CoolProp_shared.vcxproj": "MultiThreadedDebug",
+        }
+        for project, runtime in expected.items():
+            content = (Path("build") / project).read_text(encoding="utf-8-sig")
+            setting = f"<RuntimeLibrary>{runtime}</RuntimeLibrary>"
+            if setting not in content:
+                raise SystemExit(f"{project} does not contain {setting}")
```
ibell: This gate can pass while the regression it exists to catch is present.

A generated `.vcxproj` carries one `<RuntimeLibrary>` element per configuration (Debug / Release / RelWithDebInfo / MinSizeRel), and `setting not in content` searches the whole file. So the assertion is "*some* configuration uses `MultiThreadedDebugDLL`", not "*Debug* uses it".

If the `$<$<CONFIG:Debug>:Debug>` genex in `_coolprop_set_msvc_runtime` regressed such that Debug emitted `MultiThreadedDLL` while RelWithDebInfo emitted `MultiThreadedDebugDLL`, the string is still present, the step goes green, and Debug links the release CRT — precisely the `_ITERATOR_DEBUG_LEVEL` / heap-ownership mismatch the comment at `cmake/CoolPropLibrary.cmake:68-70` exists to prevent.

Worth scoping the match to the `ItemDefinitionGroup` whose `Condition` names the Debug configuration, e.g. parse with `xml.etree.ElementTree` and check the element under `Condition="'$(Configuration)|$(Platform)'=='Debug|x64'"`, rather than a whole-file substring test.

(The `MultiThreadedDebug` / `MultiThreadedDebugDLL` pair itself is fine — the closing `</RuntimeLibrary>` in the search string keeps the shorter value from matching the longer one.)

ibell (commented):
Thanks for this — the packaging design is sound and the header rework fixes a real long-standing defect. Recording what I checked and four issues found.

**What holds up.** The `EXPORT_CODE`/`CONVENTION` rework is a genuine fix: the old scheme gave shared-library consumers C++-mangled `__cdecl` declarations against an `extern "C"`/`__stdcall` DLL, and the `COOLPROP_SHARED_LIBRARY_BUILD`/`_USE` split makes both sides agree while keeping `COOLPROP_LIB` working for the Tauri (`wrappers/GUI/src-tauri/build.rs`) and Mathcad builds. Restricting `coolprop_hide_json_symbols` to SHARED is correct — link options are inert on archives. The relocatability check and the `CMAKE_DISABLE_FIND_PACKAGE_Threads` shared-only gate are both well constructed.

I separately verified:
- `target_compile_features(... PUBLIC cxx_std_17)` does **not** break a C-only consumer (built a minimal repro).
- The installed `unsupported/Eigen/Polynomials` subset resolves its relative includes (`../../Eigen/Core`, `src/Polynomials/InternalHeaderCheck.h`) from the shipped tree.
- `Eigen/COPYING.MPL2` and `fmt/LICENSE` exist at the paths installed.
- Moving `project()` above the ASAN block correctly fixes `isMultiConfig`, and the `set_target_properties(... PROPERTIES APPEND_STRING PROPERTY ...)` → `set_property(... APPEND_STRING ...)` change fixes malformed syntax on master.

**Blocking:** the first inline comment — this PR as-is fails the `dev_checks` installed-header gate and `dev/ci/preflight.sh`. Reproduced locally.

Note that no CI has run on this PR yet (fork workflows are unapproved), so none of the new gates have been exercised.

<sub>Review assisted by Claude Opus 5; findings verified locally against the branch.</sub>

ibell (changes requested):
Formally requesting changes to gate this on the blocking item from my [previous review](https://github.com/CoolProp/CoolProp/pull/3311#pullrequestreview-5058083120) — no new findings here, and no need to re-read the whole thing.

**Blocker:** installing the Eigen and fmt trees under `include/CoolProp/third_party/` makes `dev/ci/check-installed-headers.sh` fail, which takes out both the `dev_checks` CI job and the local `dev/ci/preflight.sh` gate. The script's `find ... | head -1` now picks between two include roots non-deterministically, and on the `include/` root it standalone-compiles Eigen's `InternalHeaderCheck.h` files, which are bare `#error` directives. Reproduced locally against the pinned Eigen 5.0.1 tarball; details and a suggested two-part fix are in the inline comment on `cmake/CoolPropLibrary.cmake`.

Worth flagging that this is order-dependent, so it could come back green on one runner and red on the next — please don't read a passing run as evidence it's resolved.

The other three inline comments (unguarded `COOLPROP_INSTALL_PREFIX` branch, MSVC runtime not reaching in-tree executables, fail-open Debug-CRT assertion) are worth addressing but I'd take them as follow-ups if you'd rather land the blocker first.

To be clear about what this verdict is and isn't: the packaging design is good and the `EXPORT_CODE`/`CONVENTION` rework fixes a real long-standing defect. This is a CI-gate breakage, not a rethink.

One practical note — CI has never run on this PR (fork workflows are unapproved), so none of the new gates have actually been exercised. Once you push a fix I'll approve the workflow runs so we get real signal rather than local reproduction.

.github/workflows/library_shared.yml
Tobias-Reiter: Fixed in e76466e1. The workflow now parses each .vcxproj with xml.etree.ElementTree, selects ItemDefinitionGroup nodes whose condition exactly names Debug|<platform>, fails closed if none exist, and compares the ClCompile/RuntimeLibrary value exactly. The freshly generated x64 projects were verified as MultiThreadedDebugDLL for CoolProp_static and MultiThreadedDebug for CoolProp_shared.

cmake/CoolPropLibrary.cmake
Tobias-Reiter: Fixed in e76466e1. The gate now reads CMAKE_INSTALL_INCLUDEDIR directly from the configured CMakeCache.txt, validates the exact canonical CoolProp/CoolProp.h beneath that root, and prunes CoolProp/third_party from the standalone-compilation walk. The dev_checks job deliberately configures CMAKE_INSTALL_INCLUDEDIR=cp-headers, which has no path segment named include, so the former discovery logic cannot pass accidentally. A configure/install gate run compiled all 93 CoolProp-owned installed headers successfully.

CMakeLists.txt
Tobias-Reiter: Fixed in e76466e1. COOLPROP_INSTALL_PREFIX is now honored only for a top-level CoolProp build; a nested build warns and ignores the legacy knob without changing either the scoped or cached parent CMAKE_INSTALL_PREFIX. The source-consumer fixture now defines a deliberately different COOLPROP_INSTALL_PREFIX and asserts both values remain unchanged. I also verified top-level legacy-prefix, top-level modern-prefix, and nested-prefix configurations.

cmake/CoolPropLibrary.cmake
Tobias-Reiter: Fixed in e76466e1. A canonical-runtime helper maps the selected producer variant to the target-local CRT and is applied to docuTest.exe. The snippets block applies the same static CRT choice to its manually created CoolProp library and to every snippet executable. A four-case helper probe covered static-only, shared-only, and both dual-build defaults; generated Visual Studio snippet projects with COOLPROP_MSVC_STATIC=ON and COOLPROP_MSVC_DEBUG=OFF use MultiThreadedDebug in Debug and MultiThreaded in release configurations on both sides of the link.

cmake/CoolPropLibrary.cmake
```diff
@@ -0,0 +1,520 @@
...
+  # A parent with cmake_minimum_required(VERSION 3.14) may have enabled MSVC
+  # languages while CMP0091 was OLD. In that case CMake leaves this internal
+  # default empty and ignores MSVC_RUNTIME_LIBRARY. Append an explicit
+  # target-local option so add_subdirectory consumers still get the requested
+  # runtime without rewriting the parent's global CMAKE_CXX_FLAGS_* values.
+  if(_coolprop_needs_msvc_runtime_fallback)
+    target_compile_options(
+      "${target}"
+      PRIVATE "$<$<CONFIG:Debug>:${_runtime_flag}d>"
+              "$<$<NOT:$<CONFIG:Debug>>:${_runtime_flag}>")
```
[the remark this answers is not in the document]
Tobias-Reiter: Thanks — this is worth guarding. With the Visual Studio 17 generator I could not reproduce an immediate failure before the fix: CMake mapped the recognized CRT switch to ClCompile/RuntimeLibrary and emitted no ResourceCompile/AdditionalOptions. That mapping is generator-specific, though, while target_compile_options is otherwise language-agnostic. Commit 0700330e therefore adds the explicit COMPILE_LANGUAGE:C,CXX guard. I verified a nested cmake_minimum_required(VERSION 3.14) / CMP0091-OLD shared build containing CoolProp.rc: Debug ClCompile remains MultiThreadedDebug, ResourceCompile contains no CRT option, and the isolated ResourceCompile MSBuild target succeeds. The existing static fallback consumer also still configures with MultiThreadedDebugDLL.

ibell:
@jowr can you take a look at this PR? I think you know more about the intricacies of CMake than I do

jowr:
Thanks for this, and sorry for the slow reply. I have read it properly now. The CMake is good work and I would like to build on it rather than around it.

Three things before it can land, then one design question.

**1. It needs a rebase.** Master has moved 27 commits since you branched, and seven of the files you touch have changed, so it currently shows as conflicting.

**2. An empty prefix installs to the filesystem root.** `if(DEFINED COOLPROP_INSTALL_PREFIX)` is true for an empty value:

```
cmake -DCOOLPROP_INSTALL_PREFIX=''   ->   CMAKE_INSTALL_PREFIX=[]
```

Because this PR makes every `DESTINATION` relative, that sends the whole install to `/lib`, `/include` and `/bin`. `cmake -DCOOLPROP_INSTALL_PREFIX="$PREFIX"` with `PREFIX` unset is enough to hit it. Trailing slashes are handled correctly, so it is only the empty case.

**3. `Eigen_SOURCE_DIR` and `fmt_SOURCE_DIR` are used unguarded in `install()`** (`cmake/CoolPropLibrary.cmake:230-241`). With `CPM_USE_LOCAL_PACKAGES=ON` those are unset, configure still succeeds, and the install then fails:

```
file INSTALL cannot find "//Eigen": No such file or directory
```

The `$<INSTALL_INTERFACE:.../third_party/eigen>` on the exported targets would also point at a directory nothing populated.

**The design question.** Vendoring Eigen and fmt under `<prefix>/include/CoolProp/third_party/` is right for a relocatable SDK package, and it is the one thing a distribution package cannot have: a private Eigen copy in `/usr/include` is something Debian and Fedora will reject, and it pre-empts step 2 of #3388. Could it sit behind an option, say `COOLPROP_VENDOR_THIRD_PARTY` defaulting to ON, so both cases work? I would rather keep your design as the default than lose it.

One thing worth saying: your ordering, `project()` then the prefix block then the CPM include, is correct. I got exactly this wrong on #3393, where a dependency's own `project()` populated `CMAKE_INSTALL_PREFIX` before my block ran and pointed the whole install at `/usr/local`. Yours is immune by construction.

Happy to push the two fixes and test cases to your branch if you have maintainer edits enabled, or to leave them to you. If you are short of time, just say so and I will carry it forward with your commits intact.

Thanks again for the contribution.

jowr (dismissed):
1. Please rebase on current master. It has moved 27 commits since you branched, and seven of the files you touch have changed, so it currently shows as conflicting.
2. Guard against an empty prefix installallation to not pollute the filesystem root.
3. Eigen_SOURCE_DIR and fmt_SOURCE_DIR should be validated prior to using them
4. Please add an option COOLPROP_VENDOR_THIRD_PARTY defaulting to ON, so we can control whether we want to produce and SDK-like setup whether we want to rely on system packages like Linux distributions do.

Tobias-Reiter:
Implemented the CMake feedback from this comment and rebased onto upstream master `01342963`.

- Reject an empty `COOLPROP_INSTALL_PREFIX` during configuration; preserve the ordering after `project()` and before CPM.
- Add `COOLPROP_VENDOR_THIRD_PARTY` (default ON). ON obtains source headers/licenses even with `CPM_USE_LOCAL_PACKAGES=ON`; explicit `CPM_Eigen_SOURCE` / `CPM_fmt_SOURCE` overrides remain supported. OFF uses installed Eigen3/fmt targets and exports the corresponding `find_dependency()` calls, with no private header copies.
- Diagnose missing vendoring inputs during configuration; require Eigen >= 3.4 for system packages while accepting Eigen 5.
- Extend CI packaging tests and adapt the installed-header gate to the rebased upstream parallel sweep and optional legacy copies.

Local verification on Windows/MSVC Release: both static-library modes built and installed; both packages were relocated; all 8 downstream C-API/C++-API consumer tests passed. Empty-prefix and missing-dependency negative tests passed. Explicit local-source configuration passed with FetchContent disconnected. Exported package files were checked for source-path leaks, and workflow YAML/shell syntax checks passed.

The preflight was run for the CMake-only follow-up with its Catch build/tests skipped; the focused build/install/consumer tests above were run separately. Full cross-platform CI and the Linux header sweep are left to GitHub Actions, not claimed as locally verified. OpenAI Codex implemented and checked this follow-up.

jowr:
Thanks - I'll review it ASAP.

In the meantime, could you please make sure to either sign the CLA with you tugraz email or reauthor the commits?

jowr:
I ended up pushing a bunch of commits on top of yours. Could you please review my work and check whether that all still matches your intentions? If you approve my edits, I think we are good to merge the changes, but please do not hesitate to push back on things.

Tobias-Reiter:
Thanks @jowr, your follow-up changes through `ec9d0f12` match the intended design. I approve the edits and am happy for you to merge this PR.

The review and local verification were carried out with OpenAI Codex: Windows/MSVC Release builds and installs passed with `COOLPROP_VENDOR_THIRD_PARTY` both ON and OFF, all eight downstream C/C++ consumer tests passed, and empty install prefixes were correctly rejected. The current GitHub Actions workflows are also all green. No further changes requested from my side.

jowr:
Thank you for your help!

Tobias-Reiter:
thank you too!

Issue #1677, opened 2018-03-18
Really support mixtures with zero mole fractions


ibell:
Been a historical problem for a long time

ibell:
## Research summary: root cause and fix plan

### Reproduction
A **single** zero mole fraction mostly works today; the failure appears once **two or more** fractions are simultaneously zero — most visibly a pure fluid (or sub-mixture) expressed in a larger component list:

```python
AS = CoolProp.AbstractState("HEOS", "Methane&Ethane&Propane")
AS.set_mole_fractions([1.0, 0.0, 0.0])
AS.T_reducing()        # -> nan
AS.rhomolar_reducing() # -> nan
AS.update(PT_INPUTS, 101325, 300)  # -> "p is not a valid number"
```

By contrast `[0.5, 0.5, 0.0]` (one zero) is fine.

### Root cause
The GERG-2008 reducing function builds `Tr`/`rhor` from pairwise terms

```
f_Y_ij(x) = x_i * x_j * (x_i + x_j) / (beta^2 * x_i + x_j)
```

Since mole fractions are `>= 0` and `beta^2 > 0`, the denominator vanishes **only** at `x_i = x_j = 0`, where the numerator (always carrying an `x_i*x_j` factor) vanishes faster. So every singularity is a single, **removable** point — there is no genuine pole in the composition simplex. The `NaN` is purely `0/0` at the both-zero corner, and it propagates from `Tr`/`rhor` into density, pressure, fugacity, and every flash that consumes them. The same shape recurs in all ten pairwise derivative helpers and a few inline `XN_DEPENDENT`/`dbeta` terms in `ReducingFunctions.cpp`. A precedent guard already exists in `dYr_dbeta` (`ReducingFunctions.cpp:288`).

A second, independent site: `calc_excess_properties` computes `x_i*log(x_i)` unguarded, so `gibbsmolar_excess`/`smolar_excess` return `NaN` even for a **single** zero (e.g. `[0.6, 0.4, 0.0]`) via `log(0)`. (`calc_all_alpha0_derivs_nocache` is already guarded — no change needed there.)

### Why not the active-set / REFPROP approach
[...]

PR #3314, opened 2026-08-16
feat(gerg): strict GERG-2004 / GERG-2008 backends

ibell:
Adds GERG-2004 and GERG-2008 as two new **strict** backend families, `GERG2004::` and `GERG2008::`, implementing the Kunz & Wagner wide-range equations of state for natural gases.

```python
PropsSI("Dmolar", "T", 300, "P", 1e6, "GERG2008::METHANE[0.9]&NITROGEN[0.1]")
```

"Strict" is the point. These backends admit only the components each model publishes (18 for GERG-2004, 21 for GERG-2008), carry only that model's own pure-fluid EOS, ideal-gas coefficients, binary reducing parameters and departure functions, use GERG's `R = 8.314472 J/mol/K` rather than the CODATA value, and refuse questions the model does not answer. A number from `GERG2008::` is a GERG-2008 number.

This matters because CoolProp's default `HEOS` mixture model is *not* GERG: it uses the reference pure-fluid EOS (Setzmann-Wagner for methane, and so on) rather than GERG's shortened technical form, and for 16 of the 210 GERG binary pairs it carries different reducing parameters. The two will give different numbers.

## Validation

The reference implementation is [teqp](https://github.com/usnistgov/teqp). Reference values are generated from it and committed, so the build has no teqp dependency.

- 12,878 assertions across 63 `[GERG]` cases
- Pure fluids and mixtures each reproduce teqp to ~1e-13 relative, verified independently of the committed fixtures
- All four coefficient-table families plus the reducing-point table and the derived acentric factors are re-checkable against teqp at any time via `dev/gerg/verify_transcription.py` (six families)

## Changes to shared code

Three, called out explicitly because they affect users who never touch GERG.

**`HelmholtzEOSMixtureBackend::set_mixture_parameters()` is now `virtual`**, so the GERG backends can populate the reducing function and excess term from their own tables instead of CoolProp's global JSON library. Adding a virtual member changes the class vtable layout: source-compatible, not binary-compatible.

The blast radius is narrower than "ABI break" suggests, and worth stating precisely. `src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.h` is **not an installed header** — `CMakeLists.txt:685-691` installs only from `include/` — and `AbstractState`'s own vtable is unchanged. Consumers using the installed public API against a pre-built shared library are therefore unaffected; the recompile requirement applies to code built against CoolProp's in-tree `src/` headers. The new `backend_families` / `backends` enumerators are appended at the end of their enums and are ABI-safe. Accepted as intentional by the maintainer.

**`GERG2008ReducingFunction` gained removable-singularity guards.** `f_Y_ij` is `x_i x_j (x_i + x_j) / (beta^2 x_i + x_j)`, which is 0/0 the moment two mole fractions are exactly zero — so the reducing temperature and density, and therefore every property, came back as a silent NaN with no error raised. This is the reducing function default `HEOS` uses, so it is a behaviour change for existing users: previously NaN, now correct values. Pre-existing (#1677), not introduced here, but it sits on this backend's headline use case, since a natural-gas analysis is normally supplied as the full 21-name list with most entries zero.

Guarded in five places: `f_Y_ij` itself, its two first-derivative helpers, and the two spots where the `XN_DEPENDENT` branch of `dYrdxi__constxj` *inlines* the same expression rather than calling those helpers. The inlined loop is why a single trailing zero was so destructive — it runs over every component, so `x[N-1] == 0` plus one further zero anywhere made the whole first derivative NaN. The inlined arithmetic is left byte-for-byte alone so existing results do not shift; only the both-zero corner is intercepted.

**The guards stop at first order, and that boundary is mathematical, not a scoping decision.** `f_Y_ij` is homogeneous of degree 2, so its `k`-th composition derivative is homogeneous of degree `2 - k`; over non-negative mole fractions the denominator is bounded away from zero on the unit directions, giving `|d^k f(t·u)| <= M · t^(2-k)`. For `k = 0, 1` that vanishes *uniformly in the direction*, so 0 is the unique continuous extension. For `k >= 2` the degree is 0 or negative: constant along each ray, different between rays, so no limit exists and no value is correct — measured at `beta = 1.3`, `d2f/dxi2` is `-7.09e-2` along `(1,1)`, `-8.24e-1` along `(1,9)`, `-3.24e-4` along `(9,1)`. The second derivatives are therefore deliberately left NaN. `Web/coolprop/GERG.rst` carries the full argument.

**`src/CoolProp.cpp`** gains `is_gerg_backend_string()` and refusals in `set_reference_stateS` / `set_reference_stateD`.

## Reference state — read this before comparing against HEOS

These backends report `h = 0` and `s = 0` for the **ideal gas** at 298.15 K / 101325 Pa, per component. That is not any of CoolProp's usual reference states (IIR, ASHRAE, NBP), and it is the ideal-gas value rather than the real-fluid one. Enthalpy and entropy from `GERG2008::Methane` will differ from `HEOS::Methane` by a constant offset. This is correct and matches teqp. Any cross-check must compare *differences*, not absolute values.

## What these backends deliberately do not provide

- **Transport properties.** Not part of GERG. `viscosity()`, `conductivity()` and `surface_tension()` throw `NotImplementedError`.
- **Superancillaries.** None is attached, deliberately: `FlashRoutines::sat_superanc_path_applies` returns superancillary values *as the answer* for pure fluids that own one, so a borrowed blob would return Setzmann-Wagner saturation densities labelled GERG-2008. Saturation goes through the iterative solver instead.
- **Mutable binary interaction parameters.** `set_binary_interaction_double` and friends throw. A mixture with altered beta/gamma is not GERG.
- **`change_EOS`, `update_with_guesses`, `set_reference_stateS`/`D`** are likewise refused.

## Known limitations

- **Pure-fluid `HmolarP` / `PSmolar` / `PUmolar` fail.** GERG publishes no triple point, so `EOS.ptriple` is a sentinel; that makes the saturation-calling branch of `p_phase_determination_pure_or_pseudopure` unreachable, and behind it a `1.5*Tmax()` bracket collides with the 700 K range guard. Both must be cleared together. Diagnosed and documented; pinned by a test written to fail when it is fixed. Not shared with `HEOS`, whose `p_triple()` is `sat_min_liquid.p`.
- **Compositions with two or more exactly-zero mole fractions**: single-phase properties *and* fugacities are correct and match the trimmed composition to 1e-12. **Phase envelopes still fail**, because they need `dlnphi_dxj` — a *second* composition derivative, which has no limit at the both-zero corner and so is left NaN on purpose (see the reducing-function note above). Trim the zero components if you need an envelope; the trimmed result is exact. Not GERG-specific — identical on `HEOS`. Results are also now invariant under component reordering, which they demonstrably were not before: [`CoolProp-Tests-GERG.cpp`](src/Tests/CoolProp-Tests-GERG.cpp) pins all 24 orderings of a quaternary chosen for non-unity `beta` (CO2/n-butane, `betaV = 1.174761`), and 228 of its 1974 assertions fail if the guards are removed.
- **Helium and hydrogen have no reachable saturation state** — their `Tmin` equals their critical temperature.

## Notes

Acentric factors are derived from the GERG equations themselves (`omega = -1 - log10(p_sat(0.7 Tc)/p_c)` from converged saturation, not from the fitted ancillary), keeping them GERG-derived rather than borrowed. This unblocks mixture QT/PQ flashes, phase envelopes and gas-phase `DmolarP`.

Documentation is in `Web/coolprop/GERG.rst`; the design rationale, including where it was wrong and had to be corrected during implementation, is in `docs/superpowers/specs/2026-07-25-gerg-strict-backend-design.md`.

A follow-up PR will fix two defects in `dev/ci/preflight.sh` that this work uncovered: the path-to-tag chain is first-match-wins, so the `[GERG]` suite silently never ran behind the Helmholtz arm, and there is no zero-assertion check, so a run matching nothing reports green. Those are deliberately kept out of this PR.

## Summary by CodeRabbit

- **New Features**
  - Added strict GERG-2004 and GERG-2008 equation-of-state support for natural-gas calculations.
  - Added pure-fluid and mixture properties, saturation calculations, component aliases, validation limits, supported flash operations, and model-specific behavior.
  - Added GERG backend selection and validation coverage.

- **Bug Fixes**
  - Prevented invalid results for zero-fraction mixture components.
  - Improved reference-state handling and errors for unsupported calculations.

- **Documentation**
  - Added comprehensive GERG usage, limitations, validation, and release documentation.


ibell:
## CodeRabbit review — dispositions

Thanks — five findings were real and are fixed in 6c4de73a0. Each was verified against the code before acting, and two turned out to be stronger than reported. The rest describe the *planning document*, not the shipped backend; details below rather than a silent resolve.

### Fixed in 6c4de73a0

| Finding | Disposition |
|---|---|
| `GERGBackend.h` — self-contradictory `change_EOS` rationale | **Fixed, and worse than reported.** Not just contradictory: the accident it describes no longer exists. The cubic path reads `reduce.T`, `reduce.p`, `reduce.rhomolar`, `acentric` (`HelmholtzEOSMixtureBackend.cpp:491-495`); `make_gerg_fluid` now fills all four (`GERGBackend.cpp:1427`, `:1437-1438`, `:1481`) — the last two added *after* that comment was written. A reader would conclude the guard is belt-and-braces when it is the only thing stopping `change_EOS` from silently leaving an SRK cubic behind a backend named GERG2008. Stale `set_mixture_parameters` "above" → below (`:295`) also fixed. |
| `fit_ancillaries.py` — propane deviation disagrees with README | **Fixed.** Re-ran the fit rather than picking a side: `--report` gives propane `dev rhoL 1.393e-03`, so the README's `1.4e-3` was right and the docstring's `3.5e-4` was stale. Both also called the gap "an order of magnitude"; it is 5.2×, so both now say a factor of five. |
| `fit_ancillaries.py` — aggregate initialiser order unenforced | **Fixed.** `check_struct_field_order()` parses the declared order out of `GERGData.h` and refuses to generate on a mismatch. Mutation-tested both directions: swapping `Tmin`/`Tmax` in `AncillaryCoeffs` and `rhoL`/`rhoV` in `SatEndState` each exit 1 naming the struct; silent on the current header. Output unchanged — `GERGAncillaries.h` regenerates byte-identical, as does the pre-change script's output. |
| `generate_reference_values.py` — hardcoded `teqp 0.23.2` | **Fixed**, now interpolates `teqp.__version__`. `GERGReferenceValues.h` regenerates byte-identical. |
| `compute_acentric.py` — "relative" tolerance | **Fixed.** `verify_transcription.py:842` applies `abs(committed - re-derived)`; the wording was the odd one out. |

`[GERG]`: 12878 assertions in 63 test cases, unchanged. `verify_transcription.py` green on all six families.

### Declined — already true in the shipped code

These read the plan document as the contract. It is a historical record of how the work was built, including three errors that implementers caught and corrected during execution; the code is the contract. Each was checked rather than assumed:

- **Filter reference points to `60 ≤ T ≤ 700`** — the fixtures deliberately span `0.7/1.0/1.3/2.0 × Tc` (helium 3.64 K, water 1294 K). They do not fail: `compare_pure_points` runs under `PropertyLimitsSuspender`, because those fixtures test EOS *math*, not the validity *policy*. The policy is tested separately, with mutation evidence.
- **Make the range test control `DONT_CHECK_PROPERTY_LIMITS`** — already handled, and arguably better: `CoolProp-Tests-GERG.cpp:1296` `REQUIRE`s the flag is `false`, so a leaked flag fails the test loudly instead of being masked by a forced set. Both bounds use `CHECK_THROWS_AS(..., OutOfRangeError)`, and `update_with_guesses` is covered at `:1334`.
- **Consume GERG-2004 fixtures as well as GERG-2008** — both sets are consumed (`:664` and `:679`).
- **Keep the enthalpy anchor inside the enforced range** — already clamped: `GERGBackend.cpp:1709`, `std::min(1.1 * reduce.T, limits.Tmax)`.
- **Make reference-state refusal a tested strictness rule** — already tested: *"GERG refuses set_reference_stateS instead of silently ignoring it"*, `:1875`.
- **Exclude hydrogen and helium from saturation** — documented as a known limitation in the PR description and `Web/coolprop/GERG.rst`.
- **Pin one exact teqp version / remove machine-local paths**, **`bd` instead of checkboxes**, **fenced-block language tags**, **one pressure-range contract**, **document unsupported input paths** — all against the plan/spec documents. The shipped `dev/gerg/README.md` pins teqp 0.23.2, and the PR description carries the limitations list.

The two CodeQL "implicit string concatenation" hits on `generate_reference_values.py` are false positives — those are deliberate multi-literal comment lines, each a single list element correctly terminated by a comma.

src/Backends/GERG/GERGBackend.h
ibell, line 1: don't document what you didn't do

ibell:
## Second review round — dispositions

**Your note, "don't document what you didn't do" — done (c1b48e61a).** Twenty sites across `GERGBackend.h`, `GERGBackend.cpp`, `GERGData.h` and `ReducingFunctions.cpp` had accumulated development archaeology: task numbers, review rounds, references to `task-N-report.md` files that don't ship, and narration of accidents that used to happen. Rewritten in the present tense, keeping the technical reason for each guard and dropping the history. Representative — `calc_change_EOS` went from *"THAT ACCIDENT IS GONE"* to simply stating that `make_gerg_fluid` populates all four fields the cubic path reads, so without the throw the swap succeeds silently. Same fact, no archaeology.

### CodeRabbit

| Finding | Disposition |
|---|---|
| **Guarded-site count** | **Valid — fixed** (e5afdbe60). 1 + 2 + 2 = five, not three. Cross-checked against the code: `ReducingFunctions.cpp` has six occurrences of `both_mole_fractions_zero`, one definition plus five call sites. |
| **Undefined `z` in the doc example** | **Valid — fixed.** It was also written as Python for a call the Python API doesn't expose; now a `cpp` block with `z` identified. |
| **Squared denominators underflow** | **Valid, and it falsified two claims I had written into the code.** Details below. Documented and filed as bd `CoolProp-b0rb`; not fixed here. |
| **"Denominator not bounded away from zero on non-negative unit directions"** | **Incorrect — declined.** A *unit* direction satisfies `a² + b² = 1`, so `a` and `b` cannot both be zero. On `{a,b ≥ 0, a²+b² = 1}`, `β²a + b ≥ min(β²,1) > 0`; measured minimum 1.000133 at β = 1.3. The set the objection describes is empty. |

### The underflow finding, and what it corrected

Chasing it falsified two statements I had put in the code:

1. I documented the inlined bracket as agreeing **"bit-for-bit"** with `dfYikdxi__constxk`. It does not — **60,886 of 200,000** random `(x_i, x_N, β)` draws over the full GERG β range differ in the last ULP. My earlier check sampled three points and got lucky.
2. I claimed exact-zero was the **complete** set of NaN-producing inputs over non-negative mole fractions. True only where the denominator appears to the first power. Where it is **squared**, it underflows first — measured at β = 1.3 with `x_i = x_N`, the inlined bracket silently loses its second term below ~1e-150 (`7.43e-151` against a true `6.48e-151`, a 15% error with no error raised) and returns NaN below ~1e-170, both mole fractions still nonzero.

**Not fixed, deliberately.** `dfYikdxi__constxk` computes the same quantity without squaring and is correct at those inputs, but substituting it shifts the last digit for *ordinary* compositions on every multi-fluid backend — not a trade worth making for behaviour at 1e-150. Both the correction and the limitation are now in the `both_mole_fractions_zero` SCOPE note, and `CoolProp-b0rb` carries the measurements and the mechanical fix for whoever takes it.

Two shipped limitations are now tracked rather than living only in PR text: `CoolProp-b0rb` and `CoolProp-9dk9` (helium and hydrogen have no reachable saturation state, `Tmin == Tcrit`).

Verification: `[GERG]` 14,862 assertions in 64 test cases; preflight 8 passed / 0 failed; rst clean.

PR #3319, opened 2026-08-22
fix(java): ship wrapper sources in the Java artifact, then release it

ibell:
Stacked on #3317 — **review that one first**; this PR targets `claude/fervent-taussig-d8d320`, not `master`. Closes `CoolProp-qrn3` and completes the Java half of the install step suggested in #3245.

### Description of the Change

The published Java artifact has been unusable since #3245: it carried `Example.java` and the three native libraries and **no wrapper classes at all**, so nothing in it could be compiled or called. Two independent defects combined to produce that.

**The merge job never copied the unpacked package tree.** It copied `Example.java`, `platform-independent.7z`, and the `*bit` directories — and `platform-independent/` is none of those. #3267 had already fixed the per-OS install, so the 40 SWIG-generated `org/coolprop` sources were present in each per-OS artifact right up to the merge, then discarded.

**The archive those copies looked for was never installed.** #3245 replaced the flat-glob install with the package-tree `DIRECTORY` install and dropped `install(FILES ...platform-independent.7z)` along the way, so `[ -f "${d}platform-independent.7z" ]` was false on every iteration. `dev/scripts/examples/win64run.py:35` still fetches that path from the nightly area. The rule is restored; Csharp has always shipped both the archive and the unpacked tree, and Java now matches.

**The guards are what let this ship silently for six weeks.** Every copy was written `[ -f ... ] && cp -n ... || true`, which turns a missing input into a no-op and reports success; and an unmatched `staging/Java-*/` glob left the literal pattern as the loop's single iteration, so a download that produced nothing still walked the body once and "succeeded". The copies are now unconditional and the glob uses `nullglob` with an explicit empty check.

Taking the platform-independent pieces from the first artifact rather than re-copying them per OS also removes a portability trap: `cp -n` exits 0 when it skips under GNU coreutils but **1** under BSD, so the old line was load-bearing on the runner's coreutils flavour.

Unconditional copies still only fail on files they actually touch, so a new **Verify the assembled tree is complete** step asserts the *shape* of the result — `Example.java` and the archive present, at least two wrapper sources under `platform-independent/`, one native library directory per per-OS artifact that fed the merge, and at least one library per platform. It collects every problem before exiting rather than dying on the first, and dumps the assembled tree on failure.

Both workflows also get `if-no-files-found: error` on their uploads (CodeRabbit's finding on #3317, which applied equally here): `upload-artifact@v7` defaults to `warn`, so a leg that installed nothing would upload no artifact and still pass.

With the artifact whole, `java_builder.yml` goes back into the builder list in `release_all_files.yml`.

### Verification Process

Locally on macOS with Homebrew openjdk 23 and SWIG 4.3.0:

- **Reproduced the defect first.** A clean `-DCOOLPROP_JAVA_MODULE=ON -DCOOLPROP_SWIG_OPTIONS="-package org.coolprop"` install produced `install_root/Java` with 41 `.java` files and **zero** `.7z`, confirming the archive rule was the missing half.
- After the CMake fix the same install produces `Example.java`, `platform-independent/org/coolprop/` (40 classes) and `platform-independent.7z` (40 entries, package paths intact).
- `ctest -R Javatestbuild` passes.
- Extracted both merge steps from the workflow with a YAML parser and ran them against a fabricated three-OS staging tree: assemble exits 0, verify reports `40 wrapper sources, the archive, and 3 native librar(ies) across 3 platform(s)`.
- **Attacked the verify step with seven mutations** — the historical broken shape (no `platform-independent/`, no archive), archive missing, `Example.java` missing, wrapper sources emptied, the directory removed outright, a platform dropped, a native library deleted. All seven exit 1 with named errors; the complete tree exits 0.
- **End-to-end on the merged tree**: `javac -cp Java/platform-independent` compiles `Example.java`, and `java -Djava.library.path=Java/Darwin_64bit` loads `libCoolPropJava.jnilib` and evaluates real properties (water Tc = 647.096 K, Tb = 373.124 K at 101325 Pa). The run stops only at the REFPROP example line, since REFPROP is not at `/opt/refprop` here — which is why the workflow runs `-R Javatestbuild` rather than the run test.
- `actionlint` reports zero shellcheck findings in `java_builder.yml` before and after; `release_all_files.yml` holds at 13, all pre-existing.
- `./dev/ci/preflight.sh` green apart from the already-tracked `VLERoutines.cpp:3211` PT-flash tolerance failure (`CoolProp-joqz`) — identical numbers to the baseline run, and the only CMake change here sits inside `if(COOLPROP_JAVA_MODULE)`, which is OFF for the Catch build.

### Possible Drawbacks

The `Javatestrun` ctest is still not exercised in CI (it needs REFPROP), so JNI load is covered by `Javatestbuild` plus the symbol-leak gate rather than a full run. Verified manually here instead, as above.

> [!NOTE]
> CLAUDE.md's mandatory pre-PR adversarial review subagent was **not** run — this session prohibits spawning agents unless explicitly requested, and the maintainer chose to proceed without it.

### AI assistance

Per [AGENTS.md](../blob/master/AGENTS.md#ai-assistance-and-attribution) (policy landed in #3318, after this PR was opened): **Claude Opus 5, via Claude Code, did all of the work in this PR** — the investigation, the CMake and workflow changes, the commit message, and this description. The commit carries the corresponding `Co-authored-by:` trailer.

What was verified, and how, rather than asserted:

- The defect was **reproduced before being fixed** — a clean Java install on master produced 41 `.java` files and zero `.7z`.
- Both merge steps were extracted from the workflow YAML with a parser and executed against a fabricated three-OS staging tree, then **attacked with seven mutations** (including the exact historical broken shape); all seven fail, the good tree passes.
- The `Java` artifact produced by CI run [32575175391](https://github.com/CoolProp/CoolProp/actions/runs/32575175391) was **downloaded and executed**: `javac -cp platform-independent` compiles, and the JVM loads `libCoolPropJava.jnilib` and returns water Tc = 647.096 K.
- `actionlint` and `./dev/ci/preflight.sh` were run locally; the single preflight failure is the pre-existing `VLERoutines.cpp:3211` tolerance issue tracked as `CoolProp-qrn3`'s sibling `CoolProp-joqz`.

No thermophysical model, correlation, or fitted coefficient is involved — this is CI plumbing and a CMake install rule, so there is nothing here requiring validation against a reference. @ibell remains author of record and should review before merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Java builds and releases now fail clearly when expected platform files or artifacts are missing.
  * Added validation to ensure merged Java packages contain all required wrapper, archive, and native library files.

* **New Features**
  * Java installations now include the platform-independent archive.
  * Java artifacts are now included in the complete release workflow.

### DIFF
--- a/.github/workflows/java_builder.yml
+++ b/.github/workflows/java_builder.yml
@@ -113,6 +113,10 @@ jobs:
         with:
           name: Java-${{ matrix.sysname }}
           path: install_root/Java
+          # upload-artifact defaults to 'warn', so a build that installed
+          # nothing would upload no artifact and still pass, leaving the merge
+          # job to assemble a Java tree missing that platform.
+          if-no-files-found: error
 
   merge:
     needs: build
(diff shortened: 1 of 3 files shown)


ibell:
Retargeted from `claude/fervent-taussig-d8d320` to `master`.

While this PR was based on the #3317 branch it got essentially no CI: `java_builder.yml` triggers on `pull_request: branches: [master, main, develop]`, so **the very workflow this PR rewrites never ran**, and CodeRabbit skipped the PR too ("reviews are disabled for this base branch"). Verifying the merge-job rewrite matters more than keeping a minimal diff view.

Consequence: until #3317 merges, this PR's diff also contains its two commits (`2d82ba580`, `0e93b7040`). Review #3317 on its own; the diff here collapses to the single Java commit once it lands.

ibell:
### CI-produced artifact verified

`Java Wrapper` run [32575175391](https://github.com/CoolProp/CoolProp/actions/runs/32575175391) passed on all three platforms. The merge job's new verify step reported:

```
OK: 40 wrapper sources, the archive, and 3 native librar(ies) across 3 platform(s).
```

I downloaded that run's `Java` artifact and checked it rather than trusting the log:

```
Darwin_64bit/libCoolPropJava.jnilib
Linux_64bit/libCoolPropJava.so
Windows_64bit/CoolPropJava.dll
Example.java
platform-independent/          40 wrapper .java sources
platform-independent.7z        40 entries, org/coolprop/ paths intact
```

Then compiled and ran it:

```
javac -d classes Example.java -cp platform-independent          # exit 0
java -Djava.library.path=Darwin_64bit -cp classes Example
  Critical temperature of water: 647.0959999999873 K
  Boiling temperature of water at 101325 Pa: 373.1242958476663 K
  Phase of water at 101325 Pa and 300 K: liquid
```

For comparison, the same artifact from master's last run (`31887648293`) contains `Example.java` and the three native libraries and nothing else — `javac -cp platform-independent` cannot even resolve the directory.

PR #3332, opened 2026-08-25
feat: residual entropy scaling for viscosity and thermal conductivity

### LINKED ISSUE #2753: Implement some entropy scaling models
There are a few entropy scaling models in the literature. Perhaps the one from Yang et al is the best candidate for mixtures as of today.

jakobreichert:
### Description of the Change

Adds **Residual Entropy Scaling** (RES) as a transport model for viscosity and thermal
conductivity on the HEOS, Peng-Robinson, SRK and REFPROP backends, using the published models and parameter
sets from Martinek et al. 2025 ([doi:10.1021/acs.jced.4c00451](https://doi.org/10.1021/acs.jced.4c00451)),
Li et al. 2024 ([doi:10.1021/acs.iecr.4c02946](https://doi.org/10.1021/acs.iecr.4c02946)) and the
cubic-EOS refits of Yang 2025 ([doi:10.1021/acsomega.4c10815](https://doi.org/10.1021/acsomega.4c10815)).
The RES model is opt-in for HEOS and REFPROP, which have their own transport models, and the new default for cubics.

**Layout.** `include/CoolProp/RESTransport.h` holds the parameter records and is deliberately
dependency-light so `AbstractState` can own an `RESTransportStore` without reaching
`CoolPropFluid.h`. `src/Backends/RES/RESTransport.{h,cpp}` is the model, written against
`AbstractState&` only and including no concrete backend — that is what lets REFPROP evaluate it.
`AbstractState` gains the store, the public API, and three narrow hooks a backend may implement:

| hook | why | REFPROP implementation |
|---|---|---|
| `calc_drhomass_dp_constT_at(T)` | the `(dρ/dp)_T` at `t_ref` in the Olchowy-Sengers term | one `THERM2dll` call |
| `calc_transport_native(key, ρ, v)` | the backend's own transport model at an arbitrary density | one `TRNPRPdll` call, `ρ = 0` for the dilute limit |
| `calc_critical_point_is_cheap()` | whether a mixture critical point is cheap enough to take by default | `true` — `CRITPdll` costs <1 µs |

**Parameters.** The parameter tables in `dev/RES_params` are taken from the supporting information of the cited papers. 
`dev/convert_RES_csv_to_json.py` turns the authors' tables into
`dev/res_transport_parameters.json`, embedded like the other fluid data. 151 fluids per property:
all 151 carry a `REFPROP` block, 122 carry `PR`/`SRK`, 108 carry `HEOS`. The `REFPROP` and `HEOS`
blocks are numerically identical today but kept separate, since they mean different things and
would diverge under an HEOS-specific refit.

**Three policy enums.** RES needs three inputs that are not part of the entropy-scaling model
itself, and the two reference implementations choose them per code path rather than per fluid. The
choice is invisible in the output — a wrong one still returns a plausible number — so each is an
explicit enum defaulting to `AUTO`.

What `AUTO` resolves to:

| | on REFPROP | on HEOS, PR, SRK |
|---|---|---|
| `RESDiluteSource` (viscosity) | REFPROP's `eta0` for pure fluids (polynomial if it has none), fitted polynomial + Wilke for mixtures — as Martinek does | fitted polynomial everywhere |
| `RESDiluteSource` (conductivity) | fitted polynomial for pure fluids, REFPROP's `lambda0` for mixtures (polynomial if it has none) — as Li does | fitted polynomial everywhere |
| `RESEnhancementViscosity` | REFPROP's own viscosity, as `code_SI.py:116` does (RES viscosity if it has none) | the RES viscosity |
| `RESMixtureEnhancement` | applied, as Li does | **not applied** |

The fitted model is the fallback everywhere, REFPROP included: where `AUTO` asks the backend for a
value and the backend has none, RES uses the polynomial (dilute term) or the RES viscosity
(enhancement), exactly as both papers' `try/except` does.

On **CoolProp's own backends `AUTO` makes RES entirely self-contained**: it consumes nothing but
the fitted parameters and `s_res`, never consulting the backend for a transport property, and the
mixture critical enhancement stays off. On REFPROP it reproduces the papers, falling back to the
same self-contained model wherever REFPROP cannot supply a term.

The two columns differ for different reasons per row, which is worth separating:

- **Dilute term and enhancement viscosity** — the cubics have no transport model at all, so there
  is nothing to consult; HEOS *does* have correlations and RES deliberately does not consume them,
  since doing so would change RES results and belongs in its own measured change. Either way the
  backend supplies nothing to RES and the fitted model is used.
- **Mixture enhancement** — this is a cost decision, not an availability one. It needs the mixture
  critical point: `CRITPdll` returns one in 0.3 µs, whereas `HelmholtzEOSMixtureBackend` solves for
  it (~450 ms, and it throws outright when the solver finds more than one critical point). End to
  end that is ~450 ms per conductivity call against ~7 ms with the enhancement off, versus 0.06 ms
  on REFPROP. Given that, and that the case for a critical enhancement in mixtures is not well
  established, `AUTO` leaves it off where the lookup is expensive. The hook is named `calc_critical_point_is_cheap()` for exactly that reason —
  `CRITPdll` returns an *estimate*, and REFPROP's iterative solver for the true mixture critical
  point is a separate routine (`CRTPNT`). Li 2024 reaches the same estimate through `PropsSI`, so
  RES reproduces the published values regardless; there is a test pinning that equivalence.

All three can be overridden. Setting `RES_DILUTE_POLYNOMIAL` + `RES_ENH_VIS_RES` gives the fully
self-consistent all-RES model on any backend, including REFPROP.

An explicit `..._BACKEND_NATIVE` / `..._ON` the backend cannot satisfy throws rather than quietly
substituting a different model.

**Two notes on the reference tables.** `get_paramters()` in `code_SI.py:65` overwrites `R_D` with a
flat 1.02 for every fluid and never reads the table column, so this implementation does the same —
using the tabulated per-fluid `R_D` moves PROPANE by +0.3 %, since its enhancement is ~32 % of
lambda at the sample point. `gamma` *is* read per fluid. Separately, `s_plus <= 0` or non-finite is
now refused with a named error rather than producing `NaN` through the fractional powers; the PR
backend reaches that state for hydrogen on its liquid root at 150 K / 112 MPa.

**API** — 12 methods on `AbstractState`, all inert unless enabled, exposed through both the
nanobind and Cython Python interfaces along with the three enums:

```python
AS = CoolProp.AbstractState("HEOS", "Propane")
AS.use_viscosity_RES(True)
AS.use_conductivity_RES(True)
AS.update(CoolProp.PT_INPUTS, 4.65e6, 375.9)
AS.viscosity(), AS.conductivity()
```

plus `set_/get_*_RES_parameters` (the full set, returned as a dict from Python),
`set_/get_*_RES_residual_params` (just the `n_res`/`xita` a refit changes), and the four policy
setters.

**108 rather than 122 `HEOS` entries.** 14 fluids ship without a `HEOS` block and raise instead of
returning a plausible wrong number, because their REFPROP-fitted coefficients do not transfer. The
list is measured, not judged — see below.

### Benefits

- Transport properties for the **cubic backends, which have none today** — 122 fluids on PR/SRK.
- A transport model for fluids CoolProp has no dedicated correlation for and a **much better mixture model** than the currently used mole fraction weighted average.
- Transport consistent with whichever EOS the user already chose, since the model consumes only
  `s_res`.
- Usable as a fitting target: the setters take user coefficients, so a modified alpha function or a
  new fluid can be refitted without touching library data.
- Fails loudly — non-transferring fluids, unsatisfiable policies, and out-of-domain `s_plus` all
  raise named exceptions rather than returning a number.
- Nothing changes for existing users (opt-in for HEOS/REFPROP and new feature for cubics).

### Possible Drawbacks

- **Binary size**: ~290 KB of JSON, embedded as a generated header. The `REFPROP`/`HEOS`
  duplication is deliberate but is ~half of that and could be de-duplicated.
- **ABI**: three new virtuals on `AbstractState`.
- **`RESDiluteSource` and friends moved into `DataStructures.h`** so
  `generate_constants_module.py` can reach them; that generator now scrapes three more enums.
- Predefined REFPROP `.MIX` mixtures are unsupported — `SETMIXdll` reports `Ncomp == 1`, so there
  is no per-component identity to attach parameters to. RES reports itself unsupported there.
- 14 fluids raise on HEOS, so `use_viscosity_RES(True)` is not universally safe across the library.
- **Near-critical conductivity deliberately differs from the published values** for the 114 of 143
  fluids whose `t_ref` exceeds REFPROP's `Tmax`. The reference drops the critical enhancement there
  (its `PropsSI` call raises and the `except` returns zero); this keeps it. Away from the critical
  region the difference is negligible. Happy to match the reference instead if that is preferred —
  it is a one-line guard — but silently dropping a term to reproduce a `try/except` seemed worse
  than computing it.
- Enabling the mixture critical enhancement on HEOS via `RES_MIX_ENH_ON` costs ~450 ms per
  conductivity call against ~7 ms without it, because the mixture critical point has to be solved
  for. That is the non-default path; `AUTO` leaves it off.

### Verification Process

Three layers, kept separate because "reproduces the paper" and "transfers to another EOS" are
different questions.

**1. Implementation correctness.** `dev/RES_reference_check.py` runs the SI code from both papers
and compares against this implementation on REFPROP, where the coefficients are exact.

| | median | max |
|---|---|---|
| viscosity, pure (588 points) | 0.000027 % | 53 % — see below |
| viscosity, **mixtures** (29 points) | 0.0000023 % | **0.0008 %** |
| conductivity, pure (588 points) | 0.000013 % | 44 % — see below |
| conductivity, **mixtures** (29 points) | 0.0000011 % | 16 % — see below |

Every outlier is a known divergence rather than an unexplained gap, and there are two of them.

*The reference silently drops the critical enhancement where `t_ref` exceeds REFPROP's `Tmax`.*
`t_ref` is 1.5·Tc for most fluids and `Tmax` is typically 1.5·Tc too, so for **114 of the 143**
fluids carrying enhancement parameters it sits at or just above the limit. Li's code reaches the
reference-temperature derivative through `PropsSI`, which enforces the limit and raises; `TC_RES`
catches that and returns `lambda_c = 0`. This implementation calls `THERM2dll` (REFPROP) or the
`alpha^r` derivatives (HEOS) directly and keeps the term. That is deliberate — `t_ref` is a
reference state for background subtraction, not a physical claim, and the EOS is well defined
there — but it means that near `rho = rho_c` we differ from the published numbers for those fluids
by tens of percent, because they omit a term we include. It accounts for the pure-conductivity
maximum and for the one mixture outlier (BENZENE&C16, `Tref = 963 K` against `Tmax = 725 K`).
Away from the critical region the enhancement is small and so is the difference. Documented in
`RESTransport.cpp` at the call site.

*Four HELIUM viscosity points, 46–57 %.* `PropsSI('V', T, P=1e-9, ...)` raises `TRNPRP error 561` —
the flash degenerates at that pressure — so Martinek's code falls back to its polynomial, while
this implementation calls `TRNPRPdll` at `rho = 0` and succeeds. Confirmed by sweeping the input:
the same query works at every density down to 1e-11 kg/m³ and every pressure from 1e-3 Pa up,
always returning 1.315073e-6 Pa·s against our 1.315075e-6; only 1e-9 Pa fails. The polynomial it
falls back to gives 3.13e-6 Pa·s, 2.4× the true value at 5 K.

**2. Parameter transfer**, measured on the pure fluids only — the exclusion lists it drives are
per fluid, so a binary has no entry to withhold. `dev/RES_grid_build.py` lays out 3828 pure points
over 147 fluids in reduced
coordinates, deliberately away from saturation (no point had the backends disagree on phase);
a hidden `[RES_grid]` Catch2 harness evaluates both backends at the same states with both pinned to
the same source policy; `dev/RES_grid_report.py` reports. **Median 0.00006 % / 0.00005 %** — the
parameters transfer essentially exactly for 133 of 147 fluids.

The other 14 are the exclusion list. Criterion: >1 % transport deviation at residual-dominated
states away from the enhancement, **or** >5 % deviation in `s_res` itself there. Both clauses
independently select the same 14, which is the coherent answer since `s_res` is the one input both
properties share. The second clause matters — MD4M's conductivity agrees to 0.27 % while its
`s_res` is off by 27 %, i.e. the sampled states are insensitive rather than the parameters sound.
The grid is built in reduced coordinates precisely because the published tables give one point per
fluid and several sit almost entirely in the dilute gas (residual share 0.2 % for R1234YF), where
the property cannot see an `s_res` error at all; an earlier single-point version of this list was
wrong for exactly that reason.

**3. Regression coverage.** `[RES]` — 34 cases, 669 assertions, all passing. Includes 30 pinned
values over 10 fluids on both HEOS and PR at states where the residual term supplies 55-96 % of the
property (tolerance 1e-7, against the 7.5e-5 a real coefficient fix moved ETHANE by); a test that
each of the 14 withheld fluids throws *and* that a non-withheld fluid still evaluates; the four
dilute/enhancement policy paths; all four RES flag combinations on REFPROP in both read orders.

**No regressions:**

```
[RES]                            All tests passed ( 669 assertions in 34 test cases)
[REFPROP]                        All tests passed (1526 assertions in 34 test cases)
[viscosity],[conductivity]       All tests passed (1131 assertions in  2 test cases)
[cubic],[cubics],[PengRobinson]  All tests passed (3698 assertions in 10 test cases)
[SBTL]                           All tests passed (5476 assertions in 36 test cases)
```

**Existing behaviour is untouched.** RES is inert for HEOS/REFPROP until switched on, and the two hooks it adds
(`calc_drhomass_dp_constT_at`, `calc_transport_native`) have no caller outside the RES routines, so
no existing transport path reaches new code. `[viscosity],[conductivity]` and the cubic suites pass
unchanged, which is the check that matters for that claim.

### Applicable Issues

Closes #2753.

## Summary by CodeRabbit

- **New Features**
  - Added residual-entropy-scaling options for viscosity and thermal conductivity across supported backends.
  - Added configuration and parameter access through C++, Python, and Cython interfaces.
  - Added critical-enhancement and mixture-policy controls.
  - Added expanded transport parameters, sample datasets, and support for cubic and REFPROP calculations.
  - Added tools for generating comparison grids, reports, reference checks, and parameter data.

- **Tests**
  - Added comprehensive regression, backend-parity, mixture, enhancement, caching, and validation coverage.

- **Documentation**
  - Expanded build, test, and parameter-source guidance.


ibell:
I really appreciate that you are tackling RES implementation in CoolProp. It was on my list of things to do for some time. I think we need to have a discussion about some of the architectural features of this PR though -- it breaks some of the unspoken rules of the repo, and in other places it makes some odd choices. 

A first batch of comments:
* It is not feasible to review a 33k line PR. It should be broken up into much smaller portions that can be feasibly reviewed by a human
* I think I would drop all the REFPROP changes; It adds a fair amount of complexity, and generally speaking, REFPROP should only be used with REFPROP models (it muddies the waters too much and REFPROP is the standard)
* The Cython interface for Python is dead and should not be updated
* I would recommend to opt-in to the RES model using backend options: https://coolprop.org/coolprop/BackendOptions.html
* Too many methods are added to the base class `AbstractState`. Can we not use existing methods for the transport property portions?

Perhaps we could set up a time to talk in person? If so, please email me at ian.h.bell@gmail.com

jakobreichert:
Thank you for your feedback @ibell!

I admit the scope got a bit out of hand, though it looks worse than it is. For example the json file with the literature parameters alone accounts for ~15k lines.

About the odd architectural choices, those are pretty much all a result of making the RES model work for the REFPROP backend. In fact, my first version (first commit of this PR) had everything in the TransportRoutines of the HelmholtzEOSMixtureBackend, which probably aligns much better with what you are looking for. I then decided to add REFPROP support for clean validation with the literature samples, which required moving RES to AbstractState.

I now realize I could also just plug in the HEOS backend in the sample code provided by the papers to validate the CoolProp implementation without having to add REFPROP RES support and messing up the architecture. So my suggestion is that I make a new PR branching off the first commit of this one with reworked testing/verification and the recommended BackendOptions opt-in. Would you agree with this approach or do you want to have a discussion first?

PR #3325, opened 2026-08-23
feat: five published viscosity correlations as expression-DSL data, two of them shipped

ibell:
## Summary

Five published viscosity reference correlations added to CoolProp as **expression-DSL data**, with no C++ written for any of them. Two are shipped in `dev/fluids/*.json`; three are golden test fixtures. Also wires the fifth and last unwired transport stage.

This is the follow-up to #3185, which added the DSL but left it as dark code — no shipped fluid used it.

## What's in it

**Two shipped correlations, purely additive.** `PropsSI("V", ...)` *threw* for both of these fluids before this PR; CoolProp had no viscosity model for either. No existing value changes and no existing expectation moves.

| fluid | correlation | new DSL form exercised |
|---|---|---|
| `Ethylene` | Sotiriadou, Assael, Huber — IJT **45**:87 (2024) | dilute is a **ratio of polynomials** |
| `PropyleneGlycol` | Velliadou, Assael, Huber — IJT **43**:22 (2022) | residual is an **exponential** of a polynomial |

**Three golden fixtures**, each grafted onto a copy of its fluid at runtime and registered under a synthetic name, so they exercise the real load-and-dispatch path without changing any shipped value:

| fluid | correlation | notable |
|---|---|---|
| Nitrogen | Huber, Perkins, Lemmon — IJT **45**:146 (2024) | folds the linear-in-density term into an empirical residual |
| Argon | Sotiriadou, Antoniadis, Assael, Huber — IJT **46**:133 (2025) | residual from **symbolic regression** — rational, with a real pole |
| Xenon | Velliadou et al. — IJT **42**:74 (2021) **+ its 2023 Correction** | classic Vogel/Bich second-viscosity-virial |

Nitrogen and argon are deliberately *not* shipped: they would replace working correlations (Lemmon–Jacobsen 2004) and change published values for two major fluids. That's a separate decision.

**Core change (small):** wires `viscosity_initial_density` for `"type": "expression"` — one enum value, one parse branch, one dispatch arm. It was the only unwired stage. Convention: the formula yields the stage's contribution directly in Pa·s, matching the `EMPIRICAL` arm.

## Verification

Every correlation is checked against its own paper's computer-verification values, through the ordinary fluid-data path.

| correlation | checked against | worst deviation |
|---|---|---|
| Ethylene | Table 8 — **stage-resolved** at 7 s.f., ρ = 0…550 kg·m⁻³ | **4.0e-8** |
[...]

PR #3333, opened 2026-08-29
Transport expression DSL: per-block state_variables, validated on three 2021-2025 correlations

ibell:
Builds on the runtime transport-property DSL merged in #3185. Supersedes the language half of #3325.

**No fluid data ships here.** The three correlations below are grafted under synthetic names inside the test file, so this validates the language while touching nothing a user can load. The correlations themselves are #3334.

## Why

The DSL kept its own curated list of thermodynamic names — `T`, `rhomolar`, `rhomass`, `molar_mass`, `p` — resolved in **one namespace shared with the author's own constants and arrays**. Two defects, and they compound.

*Every name the language knew was reserved everywhere.* A dilute term of the form η₀ = Σ nᵢ·Tr^pᵢ cannot call its exponent array `p`, because `p` is pressure — in a block that never reads pressure. Across real correlations, local identifiers outnumber state reads about **5:1**, and pressure is read by none of them.

*The DSL invented spellings CoolProp does not use.* `rhomolar`/`rhomass`/`p` are not CoolProp names — it says `Dmolar`/`Dmass`/`P`. That divergence is what created the lowercase `p` that collided.

## What changes

A block declares what it reads:

```json
"state_variables": ["T", "Dmolar"]
```

Declaration is **per block and opt-in**: a name not declared there is never state, however well CoolProp knows it. So the exponent array can be `p` again. Renaming stays available where it reads better — `let rho = Dmolar` — which is also how a formula carries the source paper's own symbol.

Also adds the `initial_density` stage the Rainwater–Friend correlations need.

## What may be declared

An explicit, deliberately small set of CoolProp's **own canonical** names:

```
T   P   Dmolar   Dmass   molar_mass   Smolar_residual   Bvirial   dBvirial_dT
```

Everything else is refused. Adding a name is one line, and should require the judgement that the quantity is a continuous function of the *current* state, free of configuration dependence, and not a transport output whose `keyed_output()` re-enters the correlation being defined.

**Why an allowlist and not a denylist.** A denylist over `is_valid_parameter()` was implemented first and is structurally unsound because it **fails open** — after growing to four categories it still left **59 of 92 canonical names declarable**, including `alphar` / `dalphar_ddelta_consttau` (EOS internals in *reduced* variables, so they carry the reducing state — the very defect `Tau`/`Delta` were denied for), `P_min` (which reads back `p_triple()`), and `HFORMATION`. Two gaps were found by review before that third category was noticed at all.

This is **not** the input table that was removed: that table mapped *invented* spellings and reserved every one of them in *every* formula. These are canonical names, and a block reserves only what it declares. The cost is that a new quantity is no longer free — krypton's three cost three lines — and the gain is that the next parameter CoolProp adds is refused rather than silently admitted.

## Breaking change

Third-party `"type": "expression"` blocks must be updated. No version gate and no compatibility shim — the retired spellings *were* the defect. It fails at load time, and the refusal names the whole available set, which answers "then what do I write?":

```
'rhomolar' is not a state variable the DSL exposes; available:
T, P, Dmolar, Dmass, molar_mass, Smolar_residual, Bvirial, dBvirial_dT
```

Changelog entry tracked in `CoolProp-lg3o`.

## Validation

Three recently published correlations as fixtures, against their authors' own tabulated values — plus every hardcoded transport form the DSL already reproduces bit-for-bit:

| | source | check |
|---|---|---|
| Nitrogen | Huber, Perkins & Lemmon, IJT **45**:146 (2024) | Tables 7 and 8 |
| Argon | Sotiriadou et al., IJT **46** (2025) | Table 9 grid, 41 points |
| Xenon | Velliadou et al., IJT **42**:74 (2021) + correction | saturation line, 14 points |

```
[expression]  649 assertions / 47 cases     ~[slow]  177308 assertions, 0 failures
pytest        22 passed                     preflight  6 passed / 0 failed
```

Design, and the reasoning behind the exposed set: `docs/superpowers/specs/2026-06-12-transport-expression-dsl-design.md`

## Summary by CodeRabbit

* **New Features**
  * Added expression-based initial-density viscosity correlations for fluid configurations.
  * Expression formulas can explicitly declare supported thermodynamic state variables using canonical CoolProp names.
  * Added per-block state-variable declarations and isolation.
  * Added validation for undeclared, unsupported, duplicate, conflicting, or unused variables.

* **Bug Fixes**
  * Expression-related changes now trigger the appropriate automated test coverage.
  * Improved handling and diagnostics for invalid expression configuration.
  * Added regression coverage for multiple fluid viscosity correlations.

### DIFF
--- a/src/expression/Expression.cpp
+++ b/src/expression/Expression.cpp
@@ -601,7 +601,7 @@ class Binder
             // would quietly detach the formula from the state -- and only partly, in
             // a formula that also read `T` on an earlier line.  Declaring a name and
             // then rebinding it is never what an author means.
-            if (declaredState.count(L.name))
+            if (declaredState.count(L.name) > 0)
                 throw ValueError(format("let '%s' shadows the declared state variable of the same name", L.name.c_str()));
             NodePtr bound = bind(L.expr, /*indexName*/ "");
             int slot = newScalarSlot();
@@ -837,8 +837,8 @@ Program compile(const std::string& source, const std::map<std::string, double>& 
         // and quietly change what the formula means.  This is the collision the old
         // single-namespace design hit from the other direction, where the language
         // reserved the name whether the block wanted it or not.
-        if (constants.count(nm)) throw ValueError(format("state variable '%s' collides with the constant of the same name", nm.c_str()));
-        if (arrays.count(nm)) throw ValueError(format("state variable '%s' collides with the array of the same name", nm.c_str()));
+        if (constants.count(nm) > 0) throw ValueError(format("state variable '%s' collides with the constant of the same name", nm.c_str()));
+        if (arrays.count(nm) > 0) throw ValueError(format("state variable '%s' collides with the array of the same name", nm.c_str()));
     }
     std::vector<Token> toks = lex(source);
     Parser parser(std::move(toks));
--- a/src/expression/Expression.cpp
+++ b/src/expression/Expression.cpp
@@ -1,6 +1,7 @@
 #include "CoolProp/expression/Expression.h"
 #include "CoolProp/expression/detail/Lexer.h"
 #include "CoolProp/Exceptions.h"
+#include "CoolProp/DataStructures.h"
 #include "CoolProp/detail/strings.h"
 #include <array>
 #include <cctype>
@@ -8,6 +9,7 @@
 #include <cstdint>
 #include <cstdlib>
 #include <functional>
+#include <algorithm>
 #include <map>
 #include <memory>
 #include <string>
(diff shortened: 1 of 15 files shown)


src/expression/Expression.cpp
```diff
@@ -500,22 +502,105 @@ std::vector<Token> lex(const std::string& s) {
...
+//
+// SECOND, two classes stay refused even when declared, because opt-in expresses the
+// author's intent and these are cases where the intent cannot be honoured:
+// Spellings the DSL used to accept, mapped to what replaced them.  This exists ONLY
+// to produce a good error: fluid JSON is data that third parties ship, and a block
+// written against the previous format must fail at load time saying WHAT changed,
+// not "unknown variable 'rhomolar'".  These names were the DSL's own inventions --
+// CoolProp always called them Dmolar/Dmass/P -- and the invented lowercase `p` is
+// what once collided with the exponent array every viscosity paper writes as p_i.
+static const char* retiredSpelling(const std::string& nm) {
```
ibell: I think we don't need to migrate, we can just fail

src/expression/Expression.cpp
```diff
@@ -500,22 +502,105 @@ std::vector<Token> lex(const std::string& s) {
...
+    if (nm == "p") return "P";
+    return nullptr;
 }
 
-static bool inputForName(const std::string& nm, parameters& out) {
-    for (const auto& e : inputTableImpl())
-        if (nm == e.first) {
-            out = e.second;
+static bool deniedStateVariable(parameters key, std::string& why) {
+    switch (key) {
```
ibell: At this point, is it not just better to opt-in?

src/expression/Expression.cpp
```diff
@@ -713,14 +823,43 @@ class Binder
...
 
-const std::vector<std::pair<std::string, parameters>>& inputTable() {
-    return detail::inputTableImpl();
+bool resolveStateVariable(const std::string& name, parameters& key, std::string& reason) {
+    key = iundefined_parameter;  // both out-params written on every path
+    reason.clear();
+    parameters k;
+    if (!is_valid_parameter(name, k)) {
+        if (const char* now = detail::retiredSpelling(name)) {
+            reason = format("'%s' was this DSL's own spelling; CoolProp calls it '%s'", name.c_str(), now);
```
ibell: retire this too

ibell:
All five comments addressed in d9181d719. Your two composed into one change, so they're answered together.

> **At this point, is it not just better to opt-in?**

Yes — and measuring it made the case stronger than I expected. The denylist wasn't merely incomplete, it fails **open**: after growing to four categories, **59 of CoolProp's 92 canonical parameter names were still declarable**. Beyond the two gaps review had already found:

| still slipping through | why it matters |
|---|---|
| `alphar`, `alpha0`, `dalphar_ddelta_consttau` | EOS internals in **reduced** variables — they carry the reducing state, which is exactly what `Tau`/`Delta` were denied for |
| `P_min`, `T_max`, `T_triple`, `p_triple` | range limits, not state — and `P_min` reads back `p_triple()`, so the name silently means something else |
| `HFORMATION` | fluid metadata; evaluates to a number that means nothing here |
| `gas_constant`, `acentric`, `dipole_moment` | fluid constants |

That third row is the clincher — I'd have kept patching this indefinitely.

So the DSL now exposes an explicit set — `T, P, Dmolar, Dmass, molar_mass, Smolar_residual, Bvirial, dBvirial_dT` — and refuses everything else. ~60 lines of switch cases gone, and the canonical-spelling check became redundant (`A`, `D`, `M`, `DMOLAR` simply fail membership).

Worth being explicit that this is **not** the input table we removed, since it looks similar: that table mapped *invented* spellings (`rhomolar`, `p`) and reserved every one of them in *every* formula. These are CoolProp's own canonical names, and a block reserves only what it declares — so `p` stays the author's wherever pressure isn't asked for. What it costs is my claim that a new quantity is free: krypton's three now cost three lines. What it buys is that the next parameter CoolProp adds is refused rather than silently admitted.

> **I think we don't need to migrate, we can just fail**  ·  **retire this too**

Both done — `retiredSpelling()` and both call sites are gone. It was also a list of thermodynamic names living inside the file whose stated invariant is that it keeps none, so it undercut the thing it sat next to.

Nothing is lost by removing it, because the allowlist's refusal already answers the question a migration hint would have:

```
'rhomolar' is not a state variable the DSL exposes; available:
T, P, Dmolar, Dmass, molar_mass, Smolar_residual, Bvirial, dBvirial_dT
```

An old block writing `rhomolar` is met with `Dmolar` in that list, without a second table to maintain.

---

**CodeRabbit ×3** — all three valid, all verified against the code first:

1. **preflight predicate omitted `src/Backends/Helmholtz/`** — and its own comment says the dispatch arm lives there, so comment and code disagreed. A backend-only change ran zero expression tests. Fixed.
2. **`required_inputs()` documented as "declaration order"** in `nanobind_interface.cxx` and `CoolProp.pyi` — it's first-reference order. I'd fixed `Expression.h` and `ExpressionBlock.h` and missed these two. Fixed.
3. **pytest helper omitted `state_variables` for `state=[]`** — so every test meaning "explicitly empty" was silently testing "absent". Real: those are different parser inputs. Fixed.

---

```
[expression]  649 assertions / 47 cases     ~[slow]  177308 assertions, 0 failures
pytest        22 passed                     preflight  6 passed / 0 failed
```

#3334 is rebased on this and re-verified — 181011 assertions, 0 failures; the shipped fluids use only `T, P, Dmolar, Dmass, Smolar_residual, Bvirial, dBvirial_dT`, all within the exposed set.

One housekeeping note: you have a **pending (unsubmitted) review** here — that's what stopped me replying in-thread, since GitHub allows one pending review per user. Its only draft comment is the "retire this too" above, which is already addressed, so it's safe to submit or discard as you prefer. I left it alone rather than deciding for you.

src/expression/Expression.cpp
ibell: Done — `retiredSpelling()` and both call sites removed in d9181d719.

It was also a list of thermodynamic names living inside the file whose stated invariant is that it keeps none, so it undercut the thing it sat beside.

Nothing is lost by dropping it: the allowlist refusal already prints the available set, so an old block writing `rhomolar` is met with `Dmolar` in that list — the same answer, without a second table to maintain.

```
'rhomolar' is not a state variable the DSL exposes; available:
T, P, Dmolar, Dmass, molar_mass, Smolar_residual, Bvirial, dBvirial_dT
```

src/expression/Expression.cpp
ibell: Done in d9181d719 — and measuring it made the case stronger than I expected. The denylist was not merely incomplete, it fails **open**: after growing to four categories it still left **59 of CoolProp's 92 canonical names declarable**.

The category that settles it is one neither review pass had caught: `alphar`, `alpha0`, `dalphar_ddelta_consttau` are EOS internals in **reduced** variables, so they carry the reducing state — exactly the defect `Tau`/`Delta` were denied for. Also `P_min`, which reads back `p_triple()`.

The DSL now exposes `T, P, Dmolar, Dmass, molar_mass, Smolar_residual, Bvirial, dBvirial_dT` and refuses everything else. ~60 lines of switch cases gone, and the canonical-spelling check became redundant (`A`, `D`, `M`, `DMOLAR` fail membership).

Cost, stated plainly since it undoes a claim I made earlier: a new quantity is no longer free — krypton's three would now cost three lines. The gain is that the next parameter CoolProp adds is refused rather than silently admitted. Full reasoning in the PR comment.

src/expression/Expression.cpp
ibell: Covered by the same removal in d9181d719 — this was `retiredSpelling()`'s second call site, and the helper plus both uses are gone.

Issue #2523, opened 2025-04-08
Error with calculate R1233zd(E) properties tables for Simscape Matlab using CoolProp


AdamS-mkws:
Hi,

**Description:**
 I'm trying to get a Fluid Tables properties for R1233zd(E) fluid so I can use it in Simscape Matlab. It is about two-phase properties for use as a refrigerant in Simscape.

**MATLAB command line:**
 `FluidTables = twoPhaseFluidTables([118,550],[0.050,10],25,25,60,'R1233zd(E)','py.CoolProp.CoolProp.PropsSI')`

**Error:** 
```
Error using twoPhaseFluidTables>coolpropSaturationProperties (line 849)
CoolProp failed to return fluid properties at P = 0.05 MPa, Q = 0.

Error in twoPhaseFluidTables>@(p,q)coolpropSaturationProperties(p,q,substance,coolpropFun) (line 775)
saturationProperties = @(p, q) coolpropSaturationProperties(p, q, substance, coolpropFun);
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in twoPhaseFluidTables (line 132)
            = saturationProperties(p(j), 0);
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error in VCS_Comp_teststand_001_001_Config002_testFluidProperties (line 197)
FluidTables = twoPhaseFluidTables([118,550],[0.050,10],25,25,60,'R1233zd(E)','py.CoolProp.CoolProp.PropsSI')
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Caused by:
    Error using CoolProp/CoolProp.__Props_err2 (line 358)
    Python Error: ValueError: Thermal conductivity model is not available for this fluid : PropsSI("L","P",50000,"Q",0,"R1233zd(E)")
```

Is there a way to deal with this error without having to use Refprop?

ibell:
Someone needs to implement a transport property model for this fluid. There
are several models in the literature now that would be a good choice. See
for instance the entropy scaling models from Yang and co-authors (I'm a
coauthor too)

AdamS-mkws:
Thanks for your answer,
 @ibell Could you please advise me how this could be implemented?

[...]

Issue #2229, opened 2023-03-28
Incorrect thermal conductivity computation for R245fa and R1233zd.


albertoleva:
### Description

Incorrect thermal conductivity computation for R245fa and R1233zd.

### Steps to Reproduce

```
#include <iostream>
#include "CoolProp.h"

using namespace std;
using namespace CoolProp;

void printProperties(string s1, double v1, string s2, double v2)
{
    //const string fluid = "R1233zd(E)";
    const string fluid = "R245fa";
    cout<<v2<<' '<<PropsSI("L", s1, v1, s2, v2, fluid)<<endl;
}

int main()
{
    for(double h=200000.00;h<460000.00;h+=1000)
        printProperties("P", 136730, "H", h);
    return 0;
}
```

**Expected behavior:**

Computation of the required property.

**Actual behavior:** 

Inf is returned in part of the range and a peculiar spike occurs, see the example image;

![image](https://user-images.githubusercontent.com/4070224/228290331-e637a020-1087-4106-8768-e23c21a98e4a.png)

### Versions

**CoolProp Version:** nightly
**Operating System and Version:** Ubuntu 22.04 LTS  
**Access Method:** C++

Padanian:
![image](https://user-images.githubusercontent.com/7769348/228600650-f78b1308-e52a-412b-afea-e8aaf2650b0a.png)

No discontinuities seen, but else is confirmed. More inf returned at higher enthalpies

Padanian:
![image](https://user-images.githubusercontent.com/7769348/228603950-75eedfcc-a2fa-492b-8fbd-7f6902eff540.png)

Actually, yes there is a discontinuity, and also some scattering

![image](https://user-images.githubusercontent.com/7769348/228604056-a1a3ba34-0ed9-4f64-b65e-ba175742ad15.png)

ibell:
Do densities show the same behavior?

Padanian:
![image](https://user-images.githubusercontent.com/7769348/228626451-d6b6f6fa-7b75-4427-bd1d-4b15a7dac3c5.png)

No.

ibell:
You should ignore any state points where quality is between 0 and 1. How
about outside that range?

Padanian:
I plotted the graph at P=136730 Pa, as shown by @albertoleva 

![image](https://user-images.githubusercontent.com/7769348/228627605-7b14dd3c-61ab-41b9-85b5-1bb317ec4065.png)

[...]

PR #3335, opened 2026-08-29
fix(R1233zd(E)): restore the viscosity model dropped by #2768, and refit its constants

### LINKED ISSUE #3330: R1233zd(E) lost its viscosity model in v8.0.0 - TRANSPORT block dropped by #2768
Since v8.0.0, viscosity for `R1233zd(E)` is unavailable. It worked in v7.2.0.

```python
PropsSI("V", "T", 273.15, "Q", 0, "R1233zd(E)")
# v7.2.0 -> 0.000636889
# v8.0.0 -> inf, errstring: "Viscosity model is not available for this fluid"
```

**Cause.** Commit `dd0f4952` (#2768, "Batch update: 10 multiparameter EOS ... R-1233zd(E) intl std")
replaced the fluid's EOS with the international standard equation but did not carry over its
`TRANSPORT` block. The key is now absent from `dev/fluids/R1233zd(E).json` entirely.

This looks like an oversight rather than a decision:

1. Of the ten fluids in #2768, **R1233zd(E) is the only one that previously had a viscosity
   model** - the other nine were newly added and never had one. It was the single pre-existing
   fluid in the batch.
2. v8 **kept the byte-identical `rhosr-CS` block** for the seven sibling fluids that use it
   (R1234yf, R1234ze(E), R124, R152A, R22, R245fa, R32) - including R1234yf, whose EOS was also
   replaced in the same release. Carrying an existing `rhosr-CS` correlation across an EOS
   replacement is therefore already established practice.
3. The model code is fully intact: `parse_rhosr_viscosity` in `FluidLibrary.h`,
   `TransportRoutines::viscosity_rhosr`, and the dispatch in `HelmholtzEOSMixtureBackend.cpp:866`.
   Only this fluid's *data* was lost.

**Suggested fix.** Restore the `TRANSPORT.viscosity` entry from v7.2.0 (`type: rhosr-CS`,
`BibTeX: Bell-PURDUE-2016-ETA`), in the list form v8 now uses.

I have been running that as a local patch. Measured against v7.2.0 across the saturation envelope
and superheated states: worst deviation **-2.5 %** at saturated liquid, -20 degC, tapering to ~0 %
by +50 degC. That residual is the new EOS propagating through the correlation, since
`rhosr_critical` was fitted against the old one - the same coupling already shipping for the seven
siblings. All previously non-finite states compute, and across a 96-fluid probe the restored block
alters exactly the ten `R1233zd(E)` viscosity values and no other number.

**Not a duplicate of #1826 / #2229 / #2523** - those concern *thermal conductivity*, which is
unavailable for this fluid in both v7.2.0 and v8.0.0 and is unaffected by this report.

ibell:
Fixes #3330. Also closes the viscosity half of #1826, open since April 2019.

## The regression

#2768 replaced the R1233zd(E) EOS with the Akasaka & Lemmon (JPCRD 2022)
international standard and did not carry over the fluid's `TRANSPORT` block:

```python
PropsSI("V", "T", 273.15, "Q", 0, "R1233zd(E)")
# v7.2.0 -> 0.000636889
# v8.0.0 -> inf, "Viscosity model is not available for this fluid"
```

R1233zd(E) was the only one of the ten fluids in that batch that already had a
viscosity model, which is why nothing else regressed and no test caught it.

## Two commits, deliberately separable

**1 — restore the block**, byte-for-byte from `dd0f49528^`. This alone makes the
reported call finite. Nothing else changes.

**2 — refit its two constants for the EOS now shipping.** The restored constants
have been wrong since the model was added.

`rhosr_critical` is not a fitted parameter: it is `rho_c*R*(tau*dalphar_dtau -
alphar)` at the EOS critical point, the quantity normalising `x = rho*s_r /
rhosr_critical`, so it tracks whichever EOS ships. Checkable in the tree — for
R32, R22 and R152A, whose EOS were never replaced, the stored constant matches
the value computed from the current EOS to seven digits. R1233zd(E)'s did not,
because #2768 replaced the EOS underneath it; the old value put `x` at 1.011
rather than 1 at the critical point.

`C` is then fitted on top against the 61 primary liquid points of Miyara, Alam &
Kariya, *Int. J. Refrig.* **92**, 86-93 (2018), Table 3 (313.67-433.32 K,
1.005-4.074 MPa, stated uncertainty 3.0%):

| `C` | `rhosr_critical` | AAD | bias |
|---|---|---|---|
| 1.2474 (published) | -47978.148 | 51.9% | +51.9% |
| 0.671156 (first refit, #1826) | -47978.148 | 13.9% | -13.9% |
| **0.8089 (this PR)** | **-48503.788** | **2.62%** | **-0.07%** |
| REFPROP 10 ECS | n/a | 2.78% | +2.54% |

Each row is scored on the normaliser it was fitted with. Agreement at the level
of the experimental uncertainty from one fitted parameter, matching what
REFPROP's own ECS achieves over the same states. Recomputing `rhosr_critical`
improves every metric, not just the fitted one: saturated liquid vs REFPROP over
243-313 K 7.48% → 6.71%, over 315-433 K 4.60% → 4.47%, and a (T,p) grid over
200-435 K / 0.1-30 MPa 11.9% → 10.9% AAD, worst case 51.4% → 45.5%.

`python dev/scripts/fit_R1233zdE_viscosity.py` reproduces both constants.

## This deliberately does not reproduce v7.2.0

Saturated liquid viscosity is **29-38% lower** (3.99e-4 rather than 6.37e-4 Pa-s
at 0 °C, against 3.71e-4 from REFPROP). The published `C = 1.2474` was fitted to
Hulse et al. (2012), superseded by Miyara (2018), Meng, Wen & Wu (2018) and Cui,
Yan, Bi & Wu (2018). Anyone relying on the v7.2.0 numbers for this fluid was
relying on values 40-75% above every modern measurement.

## Known limitations, recorded in the fluid file

- Miyara covers 314-433 K only. Over 243-313 K, where only the paywalled Meng
  and Cui data exist, this pair sits 6.7% above REFPROP's ECS.
- Saturated **vapor** stays ~10% high whatever `C` is — the dilute-gas term uses
  Chung-estimated Lennard-Jones parameters, which `C` cannot correct.
- Thermal conductivity is still unavailable, as in v7.2.0 (#1826, #2229, #2523
  are unaffected).

## Testing

`R1233zd(E) has a viscosity model` — verified to fail before the change with the
exact error users hit. Pins the reported call, asserts the BibTeX key, and locks
four single-phase `(T, rho)` values at 1e-6 (worst observed platform deviation
9.6e-15).

`./dev/ci/preflight.sh` passes 6/0. `cppcheck` and `clang-tidy` skipped: both
fail only on pre-existing findings in `CoolProp-Tests.cpp` that also fail on the
untouched `HEAD` version, surfacing now only because the file is in the diff.
Full `~[slow]`: 518 cases, 177k assertions, 0 failures.

## Attribution

The diagnosis and the first refit (`C = 0.671156`, fitted to figure-digitized
Meng and Cui data) are due to **@jpkauffman** in #1826, 2019-05-07. This
supersedes that value only because primary tabulated data has since become
reachable; the analysis was his.

Follow-up filed, not done here: R1234yf and R1234ze(E) should get Huber &
Assael, *Int. J. Refrig.* **71**, 39-45 (2016) — REFPROP's primary model for
both, and green open access — rather than a patch to their own stale
`rhosr_critical`.

Written by Claude Opus 5. Ian verified the physics, chose the constant, and
supplied the Miyara manuscript.

## Summary by CodeRabbit

- **Bug Fixes**
  - Restored viscosity calculations for R1233zd(E) using an updated correlation.
  - Improved saturated-liquid viscosity results, with values now aligned to published measurement data.
  - Added safeguards and validation for viscosity calculations across gas, vapor, and compressed-liquid conditions.
  - Documented known limitations: saturated-vapor viscosity remains approximate, and thermal conductivity is unavailable.

### DIFF
--- a/Web/coolprop/changelog.rst
+++ b/Web/coolprop/changelog.rst
@@ -82,6 +82,23 @@ Highlights:
   ``include/CoolProp/DataStructures.h`` are appended at the end of their enums
   and are ABI-safe.
 
+Bug fixes:
+
+* **R1233zd(E) viscosity works again.**  v8.0.0 replaced the R1233zd(E) equation
+  of state with the Akasaka & Lemmon (JPCRD 2022) international standard, but
+  the fluid's ``TRANSPORT`` block was not carried across, so ``PropsSI("V",
+  ...)`` returned ``inf`` with *"Viscosity model is not available for this
+  fluid"* where v7.2.0 had answered.  The ``rhosr-CS`` correlation
+  (``Bell-PURDUE-2016-ETA``) is restored verbatim, matching what was already
+  done for the seven sibling fluids that use the same model.  Note that
+  ``rhosr_critical`` was fitted against the superseded Mondejar et al. EOS, so
+  values differ from v7.2.0 by up to ~2.5% along the saturated liquid line
+  (largest at low temperature, tapering to ~0% above 50 degC) and by more in the
+  deep compressed liquid near the triple point; that coupling is the same one
+  already shipping for the siblings.  Thermal conductivity remains unavailable
+  for this fluid, as it was in v7.2.0.  See GitHub
+  `#3330 <https://github.com/CoolProp/CoolProp/issues/3330>`_.
+
 8.0.0
 -----
 
--- a/Web/coolprop/changelog.rst
+++ b/Web/coolprop/changelog.rst
@@ -84,20 +84,30 @@ Highlights:
 
 Bug fixes:
 
-* **R1233zd(E) viscosity works again.**  v8.0.0 replaced the R1233zd(E) equation
-  of state with the Akasaka & Lemmon (JPCRD 2022) international standard, but
-  the fluid's ``TRANSPORT`` block was not carried across, so ``PropsSI("V",
-  ...)`` returned ``inf`` with *"Viscosity model is not available for this
-  fluid"* where v7.2.0 had answered.  The ``rhosr-CS`` correlation
-  (``Bell-PURDUE-2016-ETA``) is restored verbatim, matching what was already
-  done for the seven sibling fluids that use the same model.  Note that
-  ``rhosr_critical`` was fitted against the superseded Mondejar et al. EOS, so
-  values differ from v7.2.0 by up to ~2.5% along the saturated liquid line
-  (largest at low temperature, tapering to ~0% above 50 degC) and by more in the
-  deep compressed liquid near the triple point; that coupling is the same one
-  already shipping for the siblings.  Thermal conductivity remains unavailable
-  for this fluid, as it was in v7.2.0.  See GitHub
-  `#3330 <https://github.com/CoolProp/CoolProp/issues/3330>`_.
+* **R1233zd(E) viscosity works again, with refit constants.**  v8.0.0 replaced
+  this fluid's equation of state with the Akasaka & Lemmon (JPCRD 2022)
+  international standard but did not carry over its ``TRANSPORT`` block, so
+  ``PropsSI("V", ...)`` returned ``inf`` where v7.2.0 had answered
+  (`#3330 <https://github.com/CoolProp/CoolProp/issues/3330>`_).  The
+  ``rhosr-CS`` correlation is restored.
+
+  It does **not** reproduce v7.2.0, deliberately: saturated liquid viscosity is
+  29-38% lower (3.99e-4 rather than 6.37e-4 Pa-s at 0 degC, against 3.71e-4 from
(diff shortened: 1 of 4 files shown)


jpkauffman:
@ibell, thanks for pushing this forward (and the attribution)!

ibell:
No problem! If you like, you can show your appreciation with a sponsorship: https://github.com/sponsors/CoolProp

PR #3339, opened 2026-08-29
fix(transport): rhosr-CS viscosity ignored x_crossover from the fluid file

ibell:
The `rhosr-CS` viscosity correlation blends a liquid and a vapor branch with a
logistic switch centred on `x_crossover`. Every one of the eight `rhosr-CS`
fluid files carries that centre as data, `parse_rhosr_viscosity` reads it into
`ViscosityRhoSrVariables::x_crossover` — and nothing ever read it back.
`TransportRoutines::viscosity_rhosr` hardcoded the centre as the literal `2`.

## No shipped number changes

All eight fluids (R1233zd(E), R1234yf, R1234ze(E), R124, R152A, R22, R245fa,
R32) carry `x_crossover` as the JSON integer `2`, which converts to a double
that is exactly `2.0`, so `x - data.x_crossover` is bit-identical to `x - 2`.

That is what makes it worth fixing rather than leaving. The field looks live,
its presence in eight fluid files implies it means something, and it silently
does nothing. It fails only for whoever next edits it, and it fails quietly.

The switch *sharpness* stays hardcoded at `100.0` — a constant of the
correlation rather than fluid data, carried by no fluid file. It is now named
instead of being a bare literal.

## Test

Clone R1233zd(E) twice through `add_fluids_as_JSON`, altering nothing but
`x_crossover` (2 vs 12), and evaluate at `T = 300 K, ρ = 10000 mol/m³` — where
`x = ρ·s_r/rhosr_critical` is 9.00, between the two switch locations, and single
phase.

Before the fix the two clones agree to the bit:

```
CHECK( std::abs(clone_moved / shipped - 1) > 0.01 )
with expansion:
  0.0 > 0.01
```

After it they differ by a factor of 2605. The test also pins that the shipped
fluid is unchanged by the JSON round-trip (1e-14) and `REQUIRE`s its premises —
that the block is still `rhosr-CS` and the shipped `x_crossover` is still 2 — so
it degrades to a clear failure rather than a false pass if the data moves.

## Deliberately not included

Validation of `c_liq`/`c_vap` arity and of a zero `rhosr_critical` — both one
line from the line fixed here, both real (`viscosity_rhosr` indexes `[0..3]`
unconditionally and divides by `rhosr_critical`). A throwing guard there cannot
be given a negative test today: `FluidLibrary.cpp` appends to `name_vector`
before `add_one`'s `try`, so any rejected fluid permanently strands its name in
`fluids_list`, and 12+ test sites walk that list calling `factory()` on every
name. Shipping an untested throw inside a commit whose other half closes a gate
hole would be the wrong trade. Filed instead, blocked on the orphan fix.

## Verification

Full `~[slow]`: 519 cases, 177,207 assertions, 0 failures — byte-identical to
the pre-change baseline. CI runs the whole suite (`~[slow]` in
`dev_checks.yml`, everything in `test_catch2.yml`), so the new test is gated
there.

Note for local runs: `dev/ci/preflight.sh` maps `src/Backends/Helmholtz/` to
`[Helmholtz],[REFPROP]`, which selects no transport, viscosity or conductivity
test at all — so a local preflight on this diff exercises none of the changed
code. That is a preflight-scope issue only, not a CI coverage gap, and it is
being handled separately.

`cppcheck` and `clang-tidy` were skipped locally. Being precise about what that
means: `clang-tidy` ran and produced 80 signal findings, none in the added
lines. `cppcheck` bails out of `CoolProp-Tests.cpp` at line 332 with a
`syntaxError` and never reaches the new test, so "no findings there" is vacuous
rather than reassuring. Both fail identically on `origin/master`'s copies of the
same files.

Written by Claude Opus 5, reviewed adversarially before and after the
post-review changes.

### DIFF
--- a/src/Backends/Helmholtz/TransportRoutines.cpp
+++ b/src/Backends/Helmholtz/TransportRoutines.cpp
@@ -1274,8 +1274,11 @@ CoolPropDbl TransportRoutines::viscosity_rhosr(HelmholtzEOSMixtureBackend& HEOS)
     // Calculate x
     double x = HEOS.rhomolar() * HEOS.gas_constant() * (HEOS.tau() * HEOS.dalphar_dTau() - HEOS.alphar()) / data.rhosr_critical;
 
-    // Crossover variable
-    double psi_liq = 1 / (1 + exp(-100.0 * (x - 2)));
+    // Crossover variable.  The switch location comes from the fluid file
+    // (x_crossover); the sharpness is a fixed constant of the correlation and is
+    // deliberately not exposed as data.
+    const double crossover_sharpness = 100.0;
+    double psi_liq = 1 / (1 + exp(-crossover_sharpness * (x - data.x_crossover)));
 
     // Evaluated using Horner's method
     const std::vector<double>&cL = data.c_liq, cV = data.c_vap;
--- a/src/Tests/CoolProp-Tests.cpp
+++ b/src/Tests/CoolProp-Tests.cpp
@@ -665,6 +665,71 @@ TEST_CASE("R1233zd(E) has a viscosity model (#3330)", "[viscosity],[transport],[
     }
 }
 
+// The rho*s_r correlation blends a liquid and a vapor branch with a logistic
+// switch centred on x_crossover, which every rhosr-CS fluid file carries as
+// data.  TransportRoutines::viscosity_rhosr used to hardcode the centre as the
+// literal 2 and never read the parsed field, so editing x_crossover in a fluid
+// file silently did nothing.  All eight rhosr-CS fluids happen to set it to 2,
+// so nothing was numerically wrong -- the field was simply inert, which is the
+// worse kind of bug: it fails only for whoever next tries to use it.
+//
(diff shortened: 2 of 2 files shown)


dev/ci/preflight.sh
ibell, line 1: don't modify preflight, being handled elsewhere

ibell: Dropped — preflight.sh is reverted to master, PR is now just TransportRoutines.cpp and the test (2 files, +70/-2).

For the record so it's captured wherever this is being handled: the mapping sends `src/Backends/Helmholtz/` to `[Helmholtz],[REFPROP]`, which selects 19 cases and not one transport, viscosity or conductivity test — even though TransportRoutines.cpp is under that prefix. So a local preflight on a viscosity-only change exercises none of it, which is how I ran a green preflight on this diff before noticing. CI is unaffected (dev_checks runs `~[slow]`, test_catch2 runs everything), so it's a local-gate issue only.

PR #3344, opened 2026-08-30
perf(Plots): trace isolines instead of flashing every point cold (#3269)

ibell:
Fixes the property plots in GH discussion #3269: reproducing the reporter's
two R513A figures took 255 s and 367 s and left 662 and 865 non-finite
points behind. They now take 10.4 s and 5.5 s, with 14 and 0.

cmp_Water_PH_before
cmp_Water_PH_after

cmp_R513A mix_PH_before
cmp_R513A mix_PH_after

cmp_R513A mix_TS_before
cmp_R513A mix_TS_after

### What was wrong

`IsoLine.calc_range` ran an independent, cold two-dimensional flash per
point. For a mixture those are slow (tens of milliseconds for
`PSmass_INPUTS` on R513A) and a good fraction fail outright, which is the
ragged compressed-liquid corner the reporter posted. The one place that
already tried to reuse guess values was dead code: it called
`update_with_guesses` with `HmolarP_INPUTS`/`HmassP_INPUTS`, which HEOS does
not support, so it threw on every point and NaN'd it.

### What this does

Points on an isoline are consecutive, so the previous solution is an
excellent initial guess. A single-phase point becomes a 2x2 Newton
iteration in `(T, rhomolar)` with the analytic Jacobian from
`first_partial_deriv`, on a state with the phase imposed — every property is
an explicit function of `(T, rhomolar)` for a Helmholtz EOS, so no flash is
involved, and imposing the phase both skips the saturation call that
dominates a mixture `DmolarT_INPUTS` update (1.4 us vs 640 us) and pins the
iteration to the single-phase root. A two-phase point is bracketed by the
saturation states at the constant T or p of the pair and solved for vapour
quality by an Illinois regula falsi.

|                          | before            | after        |
|--------------------------|-------------------|--------------|
| R513A p-h                | 255 s / 662 gaps  | 10.4 s / 14  |
| R513A T-s                | 367 s / 865 gaps  | 5.5 s / 0    |
| Water (both figures)     | 0.80 s / 124 gaps | 0.63 s / 0   |
| R134a (both figures)     | 0.49 s / 3 gaps   | 0.63 s / 0   |

Pure fluids are level on time and strictly cleaner. Anything the tracer
cannot do is flashed exactly as before: median 0 % of points, mean 1.9 %.

### Why most of the diff is guards, not the Newton

A phase-imposed Newton iteration converges just as happily on the metastable
extension of the EOS inside the two-phase dome, and the answer is entirely
self-consistent, so nothing downstream would notice. Five review rounds
found four separate ways that happened; each guard below exists because it
was caught returning wrong numbers:

- isolines whose inputs include neither T nor p cannot be bracketed at all
  and are not traced (`IsoLineTracer.supports`). Constant-enthalpy lines on
  a T-s diagram are the case that bit: 29 of 60 points wrong, up to 24 %
  off, on a Water line.
- a bracket that cannot be built where a dome should be hands the point
  back, and that outcome is cached with the saturation value it belongs to.
  Recording the value first left the previous "no dome here" cached against
  the new one, and a T or p isoline has one saturation value along its whole
  length, so a single failed build disabled the check for the entire line.
- an unbracketed root is checked against the saturation densities at its own
  temperature. A saturation call that fails there is not reassurance — for a
  T-indexed isoline it is the same call at the same temperature the bracket
  already failed on — so it only counts as an answer above the temperature
  where a dome can still exist. Treating it as "no dome" put four points of
  an R504 isotherm inside the dome, 36 % out in enthalpy, on points master
  gets right.
- the branch is chosen from the sign of the target property along density at
  the saturated liquid, not from the order of the two bracket values: R513A's
  bubble and dew pressures come within 1.4e-9 relative near 305 K, so that
  order is numerical noise.

The dome's temperature bound has to be exact or every supercritical point
pays for a flash. A phase envelope that ran to the critical point ends with
its liquid and vapour densities converging, and then its own extremes are
the cricondentherm and cricondenbar for free; one that stopped early says
nothing — R504's ends with them a factor of six apart, 37 K below its real
critical point — and the critical point is located directly instead.

### Verification

A harness that drives the tracer the way `calc_range` does, priming
included, checking every traced point against cold saturation densities at
its own temperature. Seven plot types, every traceable isoline type, for
R504, R513A, R454B, R410A, Air, Water, R134a and CO2 — **142 903 traced
points, none inside the two-phase dome**. The harness cannot decide about
1105 of them because the cold saturation call it checks with fails there
too; all of those were bracketed points that the tracer had itself checked
against the saturation states at their isoline's own constant, so no point
goes unexamined by both.

22 new tests in `wrappers/Python/pytest/test_isoline_tracer.py`, each of the
four defects above pinned by one that fails against the code before its fix.
The CI pytest leg now installs matplotlib, without which these and the
existing `test_derivatives_plots.py` Plots tests silently `importorskip`.

### Two CoolProp bugs found while validating this

Filed separately, not fixed here:

- mixture `DmassT_INPUTS` can return a state at a different density than it
  was given, and reports single-phase at a density provably inside the dome
  (R513A at 0.757 kg/m3, 173 K, where the saturated vapour is 0.061 and the
  saturated liquid 1501 kg/m3).
- `PQ_flash_with_guesses` silently converges on a non-saturation state near
  the critical point. The saturation calls here deliberately avoid it.

Where the tracer and a cold flash disagree, the flash is the one that is
wrong in every case examined.

### Scope

Python `Plots` only — no C++, no C API, no other wrappers. `CoolPropPlot.cpp`
has the identical cold-flash loop and is untouched; per #2463 the Python
implementation is the reference the C++ one is generated from, so the fix
belongs here first. Follow-up issue filed for the C++ port, which also notes
that `dev/scripts/generate_Plot_test_data.py` would now generate a test file
the unchanged C++ fails (values agree to 8e-10, but the NaN pattern differs
in 3 places on the R134a grid it pins).

### Verified by the human

- [ ] R513A p-h and T-s figures before/after
- [ ] the trade-off in `_check_outside_the_dome`: correctness over gap-filling
      in the near-critical band where CoolProp's saturation solver will not
      converge

bd CoolProp-dgxi

## Summary by CodeRabbit

* **New Features**
  * Improved isoline generation with intelligent point-to-point tracing for smoother, more reliable results.
  * Added an option to enable or disable tracing during isoline calculations.
  * Preserved fluid models and custom mixture settings throughout isoline calculations.

* **Bug Fixes**
  * Improved handling of single-phase and two-phase regions, difficult mixtures, and unbracketable isolines.
  * Reduced gaps and incorrect branch selection in plotted isolines.
  * Prevented invalid near-critical or collapsed saturation solutions from being accepted.

* **Tests**
  * Expanded coverage for isoline tracing, phase transitions, mixture behavior, and fallback scenarios.

### DIFF
--- a/wrappers/Python/CoolProp/Plots/Common.py
+++ b/wrappers/Python/CoolProp/Plots/Common.py
@@ -491,6 +491,31 @@ def _phases_are_distinct(state):
     return (rho_liq - rho_vap) > _TRIVIAL_SOLUTION_RTOL * max(rho_liq, rho_vap)
 
 
+def _model_key(state):
+    """Everything about a state that changes what the EOS computes
+
+    Anything cached per fluid has to be keyed on this rather than on the fluid
+    name, because two states can name the same mixture at the same composition
+    and still be different models.  A departure function installed with
+    ``set_binary_interaction_string`` cannot be read back from Python and so
+    cannot appear here; :func:`IsoLineTracer._check_clone` is what refuses to
+    trace in that case.
+    """
+    fluids = tuple(state.fluid_names())
+    interactions = ()
+    if len(fluids) > 1:
+        values = []
+        for i in range(len(fluids)):
+            for j in range(i + 1, len(fluids)):
+                for name in _BINARY_INTERACTION_KEYS:
+                    try:
+                        values.append(state.get_binary_interaction_double(i, j, name))
+                    except Exception:
+                        values.append(None)  # key absent for this backend
+        interactions = tuple(values)
+    return (state.backend_name(), fluids, tuple(state.get_mole_fractions()), interactions)
+
+
 def _clone_state(state):
     """Return a fresh AbstractState with the same backend, fluids and model"""
     fluids = state.fluid_names()
@@ -571,11 +596,15 @@ class IsoLineTracer(object):
     LIQUID = -1
     VAPOUR = 1
 
-    #: Critical temperatures keyed by fluid, for the blends whose phase
+    #: Critical temperatures keyed by fluid *model*, for the blends whose phase
     #: envelope does not reach the critical point and where locating it costs
     #: seconds to minutes.  Unlike the envelope, this lookup has no side effect
     #: on the state that produced it, so caching it cannot make one tracer
-    #: behave differently from the next.
+    #: behave differently from the next.  The key has to carry the binary
+    #: interaction parameters, not just the fluids and composition: they move
+    #: the critical point, and a shared limit that came out too low would send
+    #: :func:`_check_outside_the_dome` down its no-dome path over a range where
+    #: there is one.
     _CRITICAL_T = {}
 
     @classmethod
@@ -865,8 +894,7 @@ class IsoLineTracer(object):
         # An envelope that stopped early does not bound the dome.  The critical
         # point can be located directly, which is slow enough to be worth
         # avoiding whenever the envelope already answered.
-        key = (self._sat.backend_name(), tuple(self._sat.fluid_names()),
-               tuple(self._sat.get_mole_fractions()))
+        key = _model_key(self._sat)
         if key not in self._CRITICAL_T:
             if len(self._CRITICAL_T) > 64:  # a plotting session, not a cache
                 self._CRITICAL_T.clear()
--- a/wrappers/Python/CoolProp/Plots/Common.py
+++ b/wrappers/Python/CoolProp/Plots/Common.py
@@ -464,10 +464,17 @@ _BINARY_INTERACTION_KEYS = ('betaT', 'gammaT', 'betaV', 'gammaV', 'Fij', 'kij')
 
 #: A two-phase state whose two phases differ in density by less than this,
 #: relatively, is the trivial solution: the VLE solver collapsed both phases
-#: onto the same root rather than converging on a saturation state.  Measured
-#: over Q sweeps of eight fluids, genuine states bottom out at 3.4e-3 (Air,
-#: right at its critical point) while collapsed ones come in at ~1e-6, so this
-#: sits about thirty times clear of both.
+#: onto the same root rather than converging on a saturation state.
+#:
+#: Calibrated where it is load-bearing, which is the Q=0 and Q=1 calls the
+#: bracket and the dome check make: across the predefined blends, genuine
+#: saturation pairs there stay above 1.8e-3 and collapsed ones fall below
+#: 1e-7, so this sits orders clear of both.  At intermediate quality the
+#: separation is looser -- Air, R441A, R448A and R407C all return pairs a
+#: little above this threshold at temperatures beyond their critical point --
+#: and `calc_sat_range` will draw those.  Both error directions are safe for
+#: the tracer: a rejected genuine pair costs a fallback flash, and an accepted
+#: collapsed one still has to get past the density comparison that follows.
 _TRIVIAL_SOLUTION_RTOL = 1e-4
 
 
--- a/wrappers/Python/CoolProp/Plots/Common.py
+++ b/wrappers/Python/CoolProp/Plots/Common.py
@@ -454,6 +454,719 @@ class Base2DObject(object, metaclass=ABCMeta):
         return sat_min, sat_max
 
 
+#: Binary interaction parameters that have to be carried onto a cloned state.
+#: Rebuilding from the fluid names alone gets the defaults, so a state the
+#: caller customised with set_binary_interaction_double or
+#: apply_simple_mixing_rule would otherwise be traced with a different fluid
+#: model than it is flashed with.
+_BINARY_INTERACTION_KEYS = ('betaT', 'gammaT', 'betaV', 'gammaV', 'Fij', 'kij')
+
+
+def _clone_state(state):
+    """Return a fresh AbstractState with the same backend, fluids and model"""
+    fluids = state.fluid_names()
+    new_state = AbstractState(state.backend_name(), '&'.join(fluids))
+    if len(fluids) > 1:
+        new_state.set_mole_fractions(state.get_mole_fractions())
+        for i in range(len(fluids)):
+            for j in range(i + 1, len(fluids)):
+                for key in _BINARY_INTERACTION_KEYS:
+                    try:
+                        value = state.get_binary_interaction_double(i, j, key)
+                    except Exception:
+                        continue
+                    new_state.set_binary_interaction_double(i, j, key, value)
+    return new_state
+
+
+class IsoLineTracer(object):
+    """Walk along an isoline, warm-starting every point from the previous one.
+
+    :func:`IsoLine.calc_range` evaluates an isoline at consecutive points, so
+    the solution at one point is an excellent initial guess for the next one.
+    That turns each point into a much cheaper problem than the cold
+    two-dimensional flash it replaces:
+
+    * A single-phase point is solved as a two-by-two Newton iteration in
+      ``(T, rhomolar)``.  Every property is an explicit function of
+      ``(T, rhomolar)`` for a Helmholtz-energy EOS, so the iteration needs no
+      flash at all, only ``DmolarT_INPUTS`` updates on a state whose phase has
+      been imposed.  Imposing the phase skips the saturation call that
+      dominates a mixture update and, just as importantly, keeps the iterates
+      on the single-phase root instead of snapping to a two-phase solution.
+    * A two-phase point is bracketed by the two saturation states at the
+      constant temperature or pressure of the input pair, and the vapour
+      quality between them is found by a bracketed one-dimensional solve.
+
+    The tracer needs temperature or pressure among the two inputs -- see
+    :func:`supports` -- because that is what lets it bracket the saturation
+    curve.  Every point it cannot bracket, on either side, raises so that the
+    caller falls back to a plain :func:`AbstractState.update` call: a
+    phase-imposed Newton iteration will happily return the metastable
+    extension of the EOS inside the dome, and only the bracket rules that out.
+    """
+
+    #: The Newton iteration has settled once the relative step in both T and
+    #: rhomolar is this small.  That alone is not convergence -- see
+    #: RESIDUAL_RTOL -- but it is what stops the iteration.
+    NEWTON_RTOL = 1e-10
+    #: Residual accepted at that point, relative to how much the property
+    #: varies over the iteration variables.
+    RESIDUAL_RTOL = 1e-9
+    NEWTON_ITMAX = 30
+    #: A single Newton step may not move T or rhomolar by more than this
+    #: fraction of their current value.
+    NEWTON_TRUST = 0.25
+    #: Bracketed solve for the vapour quality of a two-phase point.
+    QUALITY_XTOL = 1e-12
+    QUALITY_RTOL = 1e-8
+    QUALITY_ITMAX = 60
+    #: A saturation bracket narrower than this (relative) cannot resolve a
+    #: quality: a pure fluid at fixed T brackets p by psat on both sides, and
+    #: an azeotropic mixture is barely better.  Such a point is single-phase
+    #: as far as the tracer is concerned.
+    BRACKET_RTOL = 1e-8
+    #: How far outside the saturation densities a converged single-phase root
+    #: may sit before it is rejected as the wrong branch.
+    BRANCH_RTOL = 1e-6
+
+    #: The reported extent of the two-phase region is a traced curve, so its
+    #: end is not exactly the cricondenbar (or cricondentherm).  A point is
+    #: only called supercritical this far beyond it.
+    DOME_MARGIN = 1.05
+    #: A phase envelope that ends with its two densities still this far apart
+    #: did not run to the critical point, so where its dome ends is unknown
+    #: and has to be found the expensive way.
+    ENVELOPE_CLOSED_RTOL = 0.35
+
+    LIQUID = -1
+    VAPOUR = 1
+
+    @classmethod
+    def supports(cls, index1, index2):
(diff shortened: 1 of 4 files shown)


ibell:
Added 4e5eb04f5: the jag at the top of the R513A dome.

Near the critical point the mixture QT/PQ solvers return the **trivial solution** — a state reporting the requested quality whose saturated liquid and vapour are the same root. On R513A at Q=0.5 the flash fails from 365 K to 367 K, then "succeeds" at 367.433 K with `rho_l = rho_v = 2408.7` mol/m3, landing 93 J/kg/K further along in entropy than the point below it (neighbouring steps are ~1.7).

Pre-existing, and reachable with no plotting code — filed as #3346. `calc_sat_range` now rejects such a state, and the tracer's bracket gets the same guard (it tested `rho_liq > rho_vap`, which a collapsed pair passes).

Threshold is 1e-4 relative. Over Q sweeps of eight fluids, genuine states bottom out at 3.4e-3 (Air, at its critical point) and collapsed ones come in at ~1e-6 — about thirty times clear either way.

Largest step along the R513A Q=0.5 line: 92.84 → 4.58 J/kg/K. The quality lines now stop where the solver stops giving real answers, and `sanitize_data` interpolates the rest.

ibell:
bd833ab80 addresses the bot findings.

**CodeQL (8, all in code this PR added)** — six `except Exception: pass` handlers now carry a comment on the handler itself saying which optional thing failed and why proceeding is right; `_check_branch` no longer mixes `return <call>` with bare returns; the test module no longer imports `CoolProp` both ways.

**Docstring coverage** — added for the functions this PR introduced that lacked them. `IsoLineTracer` is 23/23 and the test module 24/24. The residual gap against the 80 % threshold is pre-existing property setters in `Common.py`/`Plots.py` that this PR only moves past; documenting those isn't this PR's job.

**Merge risk: "can mix traced and fallback points calculated with different customized fluid models"** — accepting it explicitly, with the boundary narrowed and now stated in `_check_clone`'s docstring.

The tracer clones the caller's state, so a customised model has to come with it. Everything reachable through `set_binary_interaction_double` and `apply_simple_mixing_rule` (betaT, gammaT, betaV, gammaV, Fij, kij) is copied outright. Beyond that, `_check_clone` evaluates pressure at the caller's own (T, rho) on the clone and refuses to trace if they disagree — which catches a departure function, even though one cannot be read back from Python at all.

That probe needs a single-phase caller state. Measured on the path that matters: 27 tracer constructions across three fluids and two plot types, probe ran on 27, skipped on 0 — `PropertyPlot` leaves `_state` single-phase. So the uncovered case is a mixture customised with a *string* departure function and handed to `IsoLine`/`IsoLineTracer` directly in a state that is two-phase or never updated. Narrow, undetectable from Python by construction, and it fails toward the default model rather than toward nonsense.

If you'd rather it fail closed instead — refuse to trace any multi-component fluid when the probe can't run — say so and I'll switch it; the cost is that direct use of `IsoLineTracer` with a fresh, never-updated `AbstractState` stops tracing mixtures.

wrappers/Python/CoolProp/Plots/Common.py
```diff
@@ -454,6 +454,748 @@ def _get_sat_bounds(self, kind, smin=None, smax=None):
...
+    # Two-phase point: solve for the vapour quality
+    # -------------------------------------------------------------------
+    def _quality_residual(self, quality, bracket):
+        if self._sat_index == CoolProp.iT:
+            self._sat.update(CoolProp.QT_INPUTS, quality, self._sat_value)
+        else:
+            self._sat.update(CoolProp.PQ_INPUTS, self._sat_value, quality)
+        return self._sat.keyed_output(bracket['target_index']) - bracket['target']
+
+    def _solve_quality(self, bracket):
```
ibell, line 1076: why not toms748?

ibell: Measured it rather than guessed, and TOMS748 loses here — 1.73x more saturation flashes for the same answer.

Same workload both ways (R513A p-h isentropes + T-s isobars, R410A isentropes, Water isentropes, 250 points, 5730 two-phase points solved), swapping only the solver inside `_solve_quality`:

| | saturation flashes |
|---|---|
| Illinois | 23 466 |
| `scipy.optimize.toms748` | 40 624 (1.73x) |

Worst relative disagreement between the two results: 9.6e-11, so it is the same answer.

Illinois costs a mean of 4.10 flashes per two-phase point, distribution 1:966 2:631 3:41 4:298 5:2964 6:723 7:92 8:14 9:1 — 96 % at six or fewer, never near the 60-iteration cap.

The reason is the shape of this particular residual. Property against quality at fixed T or p is close to linear, so the secant step is nearly exact and Illinois converges in two or three evaluations; TOMS748's cubic/quartic interpolation and bracket-shrinking are overhead on a function that never needed them. Its guaranteed convergence order is insurance against a residual that misbehaves, and this one does not.

The acceptance test also differs in a way that matters. I stop on a residual criterion scaled by the bracket width (`abs(fb) > QUALITY_RTOL * width`), i.e. as soon as the answer is good; TOMS748's API is interval-based, so it keeps going until the *bracket* is tight. That is a real part of the 1.73x and not something I can tune away without giving up the guarantee it exists for.

Two other reasons not to reach for it here: `boost::math::tools::toms748_solve` — the one `AbstractState.cpp:936` uses for the Qmass/Qmolar inversion — is not exposed to Python, and `scipy` is only a soft dependency of `Plots` (`interpolate_values_1d` falls back to numpy when it is missing), so it would either become a hard dependency or need an Illinois fallback anyway.

Happy to switch if you would rather have the convergence guarantee than the flashes — it is a contained change. What protects against a bad Illinois run today is the residual check after the loop, which raises and hands the point to the flash rather than returning a half-converged quality.

ibell:
f9c874cbc fixes a blocking correctness defect that a fifth adversarial review round found, and corrects several claims in the first commit message that did not survive its audit. Both are worth stating plainly.

## The defect

`_check_outside_the_dome` short-circuited on the isoline's pressure: above the cricondenbar no two-phase state exists at that pressure, so an isobar there looked like it needed no dome check. The premise is true and my inference from it was wrong. The question the function has to answer is whether the `(T, rhomolar)` the Newton landed on is the stable root *at its own temperature*, and the EOS pressure along the metastable branch inside the dome runs well above the cricondenbar — so a root there satisfies a high-pressure isobar exactly while sitting between the saturation densities.

```
PropertyPlot('HEOS::R441A.mix','TS').calc_isolines(CoolProp.iP, num=15, points=250)
```

put **69 traced points inside the dome**. Arbitrated against a cold `PT_INPUTS` flash on a fresh state, not against the fallback flash: at p = 10.109 MPa, T = 116.404 K the tracer returned rhomolar 12960.2 where the flash gives 14958.6 and the saturated liquid is 14880.0. R410A the same on its own high-pressure isobars. Removing the short-circuit takes both to zero.

It is not repairable by tightening the bound, because the bound is not the quantity in question, so the pressure limit is gone entirely rather than left in place unused. **Cost: the R513A T-s figure goes back from 5.5 s to 12.1 s** — that short-circuit was the speed recovery. p-h is unchanged at 12.8 s / 14 non-finite, and the pair still runs in 25 s against 626 s on master.

Also cached the critical-point search the dome temperature falls back on for blends whose envelope does not reach the critical point. It was running once per isoline — 285 ms per tracer on R472A, minutes on the natural-gas mixtures — against master paying it once per plot. Unlike the envelope build it has no priming side effect, which is what made the envelope deliberately uncached.

## Corrections to the first commit message

- **"142 903 traced points … none landed inside the two-phase dome"** — false. The R441A/R410A roots above are inside it, and R410A is one of the eight fluids named. The harness could not see them: it checks a traced point against saturation densities, and a metastable root reproduces itself under that check. It needed to arbitrate against a cold `PT_INPUTS` root, which is how the review found this.
- **"1105 of them the harness could not decide … all of those were bracketed"** — not reproducible. R504 on p-h alone gives 1245 undecidable, all unbracketed.
- **"29 of 60 points wrong, up to 24 % off"** for a constant-enthalpy T-s line — that measurement was against the revision before `supports()` existed; against this code, relaxing the guard raises a `KeyError` rather than producing wrong numbers. The guard is real and doubly enforced, but the number describes an earlier state and should not have been carried forward.
- **"65 isolines, 40 falling back on none"** — correct for the figures as defined (p-h: Q×11, T×25, s×15; T-s: Q×11, p×25, 250 points, `tp_limits='ACHP'`), which is the discussion's own isoline set. A reviewer counting `num=15` throughout gets 45 and 24. The figure definition should have been stated with the number.
- **T-s baseline "367 s / 865"** — I re-measured and get 371.6 s / 865, consistent across four runs with the definition above. A reviewer using different axis limits measured 286.5 s / 506. Both are real; the number depends on the figure, and mine is the one matching the discussion's script.

Claims that did hold up under audit, several to the digit: the p-h before/after (247–255 s / 661–662 → 10.0–10.4 s / 14), the R513A bubble-dew degeneracy at 1.4204e-9, the R504 envelope geometry (6.49× density ratio, 37.1 K short), the `update_with_guesses` pair list, and the Water/R134a gap counts.

## Where that leaves the review record

Five adversarial rounds, five defects, four of them introduced by the fix for the previous one — this one included. Each is now pinned by a test that fails against the code before its fix. I would not treat the current state as reviewed: the two commits after the round-5 fix have not been through a round, and the harness change the review recommends (arbitrate against a cold PT root) is not yet done.

ibell:
029b2f11d addresses the new bot findings.

**CodeRabbit Major — `_CRITICAL_T` cache key omits the fluid model.** Correct, and worth naming for what it is: this is the same caching hazard that made the phase envelope deliberately uncached, reintroduced in the sibling lookup I added next to it two commits later. `_clone_state` copies the binary interaction parameters onto the state and those move the critical point, so two states naming the same mixture at the same composition with different `kij` would have shared a cached critical temperature. One that came out too low sends `_check_outside_the_dome` down its no-dome early return over a range where there *is* a dome — precisely the hole f9c874cbc closed.

`_model_key` now carries the interaction parameters. Verified: same fluids, same composition, `betaT` 1.0 vs 1.05 → different keys. A departure function set by string still cannot appear in the key because it cannot be read back from Python; `_check_clone` is what refuses to trace in that case, and the docstring now says so.

**CodeRabbit minor — Ruff S110/S112/E702 in the new test.** Fixed, and the same patterns elsewhere in the file rather than only the flagged lines: semicolon-joined statements split, `except Exception` narrowed to `except ValueError`, and every `continue` annotated with what it is skipping.

**CodeQL empty-except at the test's line 522** — same change.

The eight CodeQL findings from 22:52 were fixed in bd833ab80 and should clear on the next scan; they predate it.

23 tests pass.

ibell:
Round six of adversarial review. **First round that found no new correctness defect in the tracer.** Three fixable findings, all addressed in 07f605f8d, plus one claim of mine that was wrong.

**`d178f1942` was unpinned.** Reverting its one line left the suite green, so my claim in f9c874cbc that it was covered was false. Now pinned. The subtlety I had missed: the dangerous root is the one *outside* a collapsed saturation pair, not inside it — a root between two nearly-equal numbers is caught either way, while a root anywhere else reads as "outside the dome", and that is every root the Newton can reach. My first attempt at the test passed against the reverted code for exactly that reason.

**The matplotlib install could fail open on every leg**, leaving the Plots suite `importorskip`ped and the job green — the condition it was added to remove. Now counted per leg and asserted at the end, mirroring the `ran_abi3` idiom already in that file.

**`_TRIVIAL_SOLUTION_RTOL`'s justification comment overstated its margin.** It is calibrated where it is load-bearing (Q=0/Q=1: genuine ≥1.8e-3, collapsed ≤1e-7). At intermediate quality the separation is looser — Air, R441A, R448A and R407C return pairs slightly above the threshold beyond their critical points. Comment corrected.

**A correction to my earlier correction.** I said the per-isoline critical-point search was a regression this PR introduced. It is not. Master pays the same search once per `IsoLine` too, via the `Base2DObject.state` setter → `critical_state` → `get_critical_point`. Measured `IsoLine` construction on R472A: **master 0.857–0.875 s, this branch 0.849–0.867 s** — unchanged. My `_CRITICAL_T` cache removes 285 ms of it; the remaining ~860 ms is pre-existing and untouched. The natural-gas "minutes per tracer" hazard is likewise master's, not mine.

**Verified safe** by this round, on the things I most wanted attacked: `_CRITICAL_T` cold-vs-warm output is byte-identical across R472A/R472B/R504 (unlike the envelope, `all_critical_points` writes only `spinodal_values` and has no priming analogue); `_model_key` is complete by construction; `_check_clone` genuinely refuses on a string departure function and its probe ran on every leg of every ordering; `calc_sat_range`'s regression against master is exactly 2 points, both the R513A jag.

**Verification harness rebuilt**, per the round-5 recommendation. It now arbitrates against a cold `PT_INPUTS` root rather than only saturation densities. Self-test on R441A T-s, same sweep twice: the old check flagged 0 on current code and 33 with the round-5 defect reintroduced; the new one flags 71 and 103. It sees what the old one structurally could not. Re-swept with it: R513A, Water, CO2, R134a, Air — ~91 000 traced points, **0 inside the dome**. All 71 R441A flags adjudicated: 42 below the EOS `T_min` at ~3 Pa, 26 at a temperature where no dome resolves, 3 where the *flash* root is inside the dome and the tracer's is the compressed liquid.

Two follow-ups I have not done, flagged rather than buried:
- R454B T-s at the **default** axis limits traces 0.2 % of points in 111 s — removing the pressure short-circuit cost that fluid 62.5 s → 111.1 s. Correctness is unaffected (0 inside dome); it wants a separate look at whether the default limits are sensible for that fluid.
- `calc_sat_range`'s near-critical rescue is dead code on master (`pass` where `continue` was meant, so the substituted values are overwritten with NaN). Pre-existing, not activated by this PR, worth its own issue.

PR #3348, opened 2026-09-01
Add 'C2H6' alias for Ethane

Tobias-Reiter:
CoolProp currently accepts `n-C2H6` but not `C2H6` as an alias for Ethane. Since C2H6 unambiguously identifies Ethane and analogous formula aliases such as CH4 and C3H8 are already supported, this adds C2H6 to the Ethane alias list.

No thermophysical data or model behavior is changed.

## Summary by CodeRabbit

* **Enhancements**
  * Added “C2H6” as an alternative name for the Ethane fluid, improving recognition when using its chemical formula.

### DIFF
--- a/dev/fluids/Ethane.json
+++ b/dev/fluids/Ethane.json
@@ -4306,7 +4306,8 @@
       "ethane",
       "ETHANE",
       "R170",
-      "n-C2H6"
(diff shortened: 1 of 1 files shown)


ibell:
Thank you!

PR #3352, opened 2026-09-02
feat(fluids): 11 viscosity correlations as expression data, three following corrected equations

ibell:
Adds 11 transport-property viscosity correlations as expression-DSL fluid data, using the language shipped in #3333. Each was implemented from REFPROP 10.1's fluid file and then **validated against the verification points its own paper publishes** — the standing rule for this work being that a correlation which cannot reproduce its paper's own table does not go in.

Xenon, R-161, Novec 649, Ammonia, R-245fa, n-Undecane, R-32, R-1234yf, R-1234ze(E), Methane, EthylBenzene.

**R-134a and Ethanol were implemented and validated but are deliberately NOT shipped here** — see below.

## Three shipped correlations use an equation that differs from the one printed in their paper

Worth a reviewer's attention, because in each case **the fluid file deliberately differs from the paper's typeset equation** — someone checking one against the other would reasonably conclude CoolProp had a transcription bug. It is the reverse. Each was found by evaluating at the verification points the authors themselves publish, and each is recorded in a `"note"` on the affected `higher_order` block.

**One of these the authors have already corrected in print.** CoolProp follows the published correction, and now cites it:

| correlation | published correction | what it corrects |
|---|---|---|
| **Xenon** Eq. 6 | Correction, *IJT* **44**(4) 2023, [10.1007/s10765-023-03175-5](https://doi.org/10.1007/s10765-023-03175-5) | *"a power "12" … should have been used instead of "7""* |

This is cited as `Velliadou-IJT-2021-xenon-CORR`, following the `Thol-CES-2015-CORR` precedent, with the fluid file naming both the original and the correction. (Ethanol Eq. 12 had the same class of defect and the same kind of published erratum, [10.1007/s10765-023-03160-y](https://doi.org/10.1007/s10765-023-03160-y), but that correlation is withdrawn — see below.) Note that neither `ETHANOL.FLD` nor `XENON.FLD` cites its erratum — which is how the corrected coefficients reached this work without the corrections themselves ever surfacing.

**The other two have no published correction** recorded in Crossref as of September 2026. Their notes say so explicitly and date the claim:

| correlation | printed equation says | required | error at the paper's own check point |
|---|---|---|---|
| **R-161** Eq. 8 | `c5·Tr² + Tr²ρr²` | a **minus** | 84.205 vs 308.22 µPa·s — **72.7 % low** |
| **Krypton** Eq. 13 | `−1 + Σ dᵢ(s⁺)ⁱ` | the `exp` | **negative viscosity** at all five points |

Krypton is the clearer of the two: the paper's own Fig. 8 reference implementation computes `math.exp(sum(...))-1.0`, so the paper contradicts itself, and the listing is what generated Table 3.

Every deviation quoted above was measured by substituting the printed form into an otherwise unchanged implementation, not estimated. In all four cases REFPROP's fluid file carries the form that works.

Every other printed equation in this family was read and compared term-by-term and is **correct as printed**: R-32, Novec 649, n-undecane, ethylbenzene, THF, propylene glycol, ethylene, R-1234yf, R-1234ze(E), ammonia, methane, R-245fa, and (not shipped here) R-134a, argon and nitrogen. Nothing in the set is left unchecked.

## Seven of the eleven supersede an existing model

`parse_viscosity` takes element 0 of a list and stops, so each supersession makes the viscosity entry a **list with the new model first and the previous model preserved behind it** — nothing is deleted. Verified programmatically that every previously-shipped entry survives byte-identically.

## Why R-134a is held back

R-134a is the ECS reference fluid for seven other refrigerants, and `TransportRoutines::viscosity_ECS` evaluates the reference's `calc_viscosity_background()` — so promoting the 2022 correlation over the 2003 one moves all seven. Measured at saturated liquid / 0.7 Tc by swapping only `R134a.json`:

| R11 | R1132a | R116 | R12 | R143a | R236EA | R236FA |
|---|---|---|---|---|---|---|
| +1.88 % | −2.58 % | +1.69 % | +1.14 % | −2.53 % | +1.58 % | +1.92 % |

Nothing in this batch justifies moving them, and whether to accept that shift is a different question from whether the 2022 correlation is the better R-134a correlation. So R-134a stays on Huber, Laesecke & Perkins (2003) here and the decision gets its own change (bd CoolProp-3bpg). `R134a.json` is byte-identical to master and the seven dependents are bit-identical to master — 0.00e+00 on each, not merely within tolerance.

**Two ECS locks stay in**, pinned to the current 2003-reference values: R143a at 242.1 K and R236FA at 278.649 K, saturated liquid. None of the seven had a single transport regression point before this work, which is exactly why the shift was silent; these ensure the next change to R-134a's viscosity has to announce itself. They are labelled in place as self-generated, not published data.

Separately, **EthylBenzene's thermal conductivity moves +5.5 %** (0.141062 → 0.148828 W/m/K) because `simplified_Olchowy_Sengers` divides by viscosity. That is an improvement — deviation from the expected value falls 4.64 % → 0.61 % — but its 5e-2 tolerance could absorb a 5 % move in *either* direction and so could not have detected it. Tightened to 1e-2. Ethanol's much smaller coupling (0.013 %) moved its tolerance 1e-4 → 2e-4.

## Why Ethanol is withdrawn

Found during review of this PR. The 2023 correlation's residual denominator vanishes along `τ²(1+δ²) = 1`. Most of that locus sits inside the two-phase dome, but its low-density end does not — from 493.6 K to the critical point it lies in **stable superheated vapour**, inside the correlation's own 159–620 K range, where the viscosity diverges and changes sign.

It is unlikely to be hit: the singularity is ~1e-7 wide in relative density, and a 2601-point isotherm at 505 K with 1200 points clustered within ±2 % of the pole never produced a bad value. You have to land within ~1e-8 to get a negative. But "unlikely" is not "never", and what it produces is a **negative viscosity** — which propagates silently rather than failing loudly. Not a good trade for a correlation whose benefit is accuracy.

`Ethanol.json` is byte-identical to master, back on Kiselev et al. (2005). The conductivity tolerance at 400 K / 690 kg·m⁻³ also returns to 1e-4 — it had been widened to 2e-4 only because `simplified_Olchowy_Sengers` divides by viscosity. Ethanol is not an ECS reference for anything, so nothing else moves. Tracked as bd CoolProp-z94e; the correlation's authors have been informed.

## Test coverage: net reduction, deliberately

Per-fluid golden tests in `CoolProp-Tests-Expression.cpp` were removed in favour of the papers' verification points in `CoolProp-Tests.cpp`. This is a **real reduction** for several fluids — THF went from 35 tabulated points to 2, R-134a's 22-point saturation line to 3, xenon's 14 to 3 — and it is stated here rather than left to be discovered. The DSL language tests are untouched; what went is per-fluid data duplication.

## Verification

- `./build_catch/CatchTestRunner "~[slow]"` → **181,291 assertions, 0 failures**
- `./dev/ci/preflight.sh` → 6 passed / 2 failed / 2 skipped. **Both failures are pre-existing and reproduce on unmodified master**: cppcheck reports the identical `syntaxError` on master's own `CoolProp-Tests.cpp` (it cannot parse `TEST_CASE_METHOD`) plus a `CubicBackend.h` finding in a file this diff never touches; clang-tidy's 79 signal findings are all constructs that exist identically on master, in a file where this diff adds only comments and data rows. That is the known whole-file scoping issue, CoolProp-swqc.
- `python -m pytest dev/atct/test_fetch_atct_formation.py` → **45 passed** (this gate reads `dev/fluids/*.json` with no paths filter; it caught a JSON-formatting regression on this branch, fixed in 60240b688)
- An adversarial review ran and returned seven findings, all verified independently before acting and all addressed in 344830fb5 — including one where a comment in my own diff claimed a replacement test that had been deleted later in the same branch. **Five commits have landed since that review and are not covered by it**: 8fd0741bd, 60240b688, 44150c41f, 13198984d, 17dab0be8 (dropping the errata note, the JSON-formatting fix, the erratum citations, the R-134a revert, and the BibTeX entries).

## Summary by CodeRabbit

* **New Features**
  * Added viscosity correlations for ammonia, ethanol, ethylbenzene, methane, Novec 649, R-1234yf, R-1234ze(E), R-161, R-245fa, R-32, xenon, and n-undecane.
  * Added reference data for viscosity and thermal-conductivity correlations across multiple fluids.

* **Bug Fixes**
  * Corrected selected correlation terms, coefficients, precision values, and reference citations.

* **Tests**
  * Updated viscosity validation points, including R-134a and saturated-state measurements.
  * Consolidated shipped-correlation verification and adjusted conductivity tolerances.


dev/fluids/Methane.json
```diff
@@ -3988,69 +3988,187 @@
...
+          }
+        },
+        "higher_order": {
+          "type": "expression",
+          "formula": "let tau = T/T_red\nlet delta = Dmolar/rho_red\nlet front = tau^0.5*delta^(2/3)\nlet A = (a1*delta^3 + a2*delta^2)/(tau - delta + a3)\nlet Bt = (a4 + tau^-1.48)*(a5*delta + a6*delta^2 + a7*delta^3 + a8*delta^4)\n1e-6*front*(A + Bt)",
+          "state_variables": [
+            "T",
+            "Dmolar"
+          ],
+          "constants": {
```
ibell, line 4094: why are these `a` constants and not array?

dev/fluids/n-Undecane.json
```diff
@@ -4073,5 +4073,56 @@
...
+        "type": "expression",
+        "formula": "let Tr = T/T_red\nlet rr = Dmass/rho_red\nlet den = c1 + c2*Tr + rr^2 + Tr^2 + c3*rr*Tr + c4*rr\n1e-6*rr^(2/3)*Tr^0.5*c0/den",
+        "state_variables": [
+          "T",
+          "Dmass"
+        ],
+        "constants": {
+          "T_red": 638.8,
+          "rho_red": 236.7914,
+          "c0": 256.66394,
```
ibell: these `c` should be array

dev/fluids/Novec649.json
```diff
@@ -4231,5 +4231,79 @@
...
+        "type": "expression",
+        "formula": "let tau = T/T_red\nlet delta = Dmolar/rho_red\nlet simple = c0*tau^0.5*delta^(2/3)\nlet den = c2 + c3*delta + c4*tau + c5*tau*delta + c6*tau*delta^2\n1e-6*(simple + c1*tau^0.5*delta^(2/3)/den)",
+        "state_variables": [
+          "T",
+          "Dmolar"
+        ],
+        "constants": {
+          "T_red": 441.81,
+          "rho_red": 1920.0,
+          "c0": 22.0057,
```
ibell: these `c` should be array

dev/fluids/Xenon.json
```diff
@@ -3569,5 +3569,131 @@
...
+        "note": "The form used here follows the authors' PUBLISHED CORRECTION, not the original paper. Velliadou, Tasidou, Antoniadis, Assael, Perkins & Huber, 'Correction to: Reference Correlation for the Viscosity of Xenon...', Int. J. Thermophys. 44(4) (2023), doi:10.1007/s10765-023-03175-5, 24 February 2023: 'in Eq. 6, a power \"12\" in the reduced density in the third term in the bracket should have been used instead of \"7\"'. Using the original as printed gives 189.815 uPa.s at 300 K / 2500 kg/m^3 against the paper's tabulated 206.449 (8.06 % low); the corrected form reproduces the table. XENON.FLD carries the corrected exponent but does not cite the correction.",
+        "formula": "let Tr = T/Tc\nlet rhor = Dmolar/rhoc\n1e-6*rhor^(2/3)*Tr^0.5*(Tr + c0*Tr*rhor^4 + c1*rhor^12/Tr + (c2 + c3*rhor)/Tr^2)",
+        "state_variables": [
+          "T",
+          "Dmolar"
+        ],
+        "constants": {
+          "Tc": 289.733,
+          "rhoc": 8400.0,
+          "c0": 1.396328251,
```
ibell: these `c` should be in array

dev/fluids/Methane.json
ibell: Done in bd54fee8c — `p[i]/pd[i]` for the `(a1·δ³ + a2·δ²)` numerator and `b[i]/bd[i]` for the four-term polynomial.

Two of the eight aren't series members: `a3` (the pole offset in `tau - delta + a3`) and `a4` (the additive constant in `(a4 + tau^-1.48)`). Rather than leave them hand-numbered I renamed them `f3` and `f4`, which are their names in the paper's own Table 5, so they now trace back to the source.

Note the DSL only subscripts by the sum index (`arr[i]` inside `sum(i: ...)`) — there's no literal `a[0]` indexing — so this is a restructuring into sums, not a container swap. Verified value-preserving: Methane is bit-identical at all 36 states I checked.

dev/fluids/n-Undecane.json
ibell: Done in bd54fee8c. The denominator is now `sum(i: q[i]*Tr^qt[i]*rr^qd[i])`, six terms — which also makes the two implicit unit coefficients (`rr²` and `Tr²`) explicit as 1.0 entries instead of bare terms in the formula.

`c0` was the odd one out: it's the single numerator multiplier, not part of the series, so it's now `num` rather than a hand-numbered scalar.

dev/fluids/Novec649.json
ibell: Done in bd54fee8c. Denominator is now `sum(i: c[i]*tau^ct[i]*delta^cd[i])`, five terms.

I also factored the shared `tau^0.5*delta^(2/3)` out of both terms, which leaves `c0` and `c1` as the two genuine scalars — renamed `A` and `B`. That reassociation is the only reason any Novec 649 value moved, and the move is ≤1.11e-15 relative.

dev/fluids/Xenon.json
ibell: Done in bd54fee8c — the entire bracket is a plain sum, so it collapses to `sum(i: n[i]*Tr^t[i]*rhor^d[i])` with five terms and no scalars left at all beyond `Tc`/`rhoc`.

This also makes the shipped form line up term-for-term with the `$RF RED SUM:5` block in XENON.FLD, including the `ρr^12` from the published correction.

dev/fluids/Ethanol.json
```diff
@@ -4697,115 +4697,221 @@
...
+              -1.5,
+              -2.5,
+              -5.5
+            ]
+          }
+        },
+        "higher_order": {
+          "type": "expression",
+          "note": "ERRATUM, paper vs. this implementation: Eq. 12 of Sotiriadou, Ntonti, Velliadou, Antoniadis, Assael & Huber, Int. J. Thermophys. 44(3):40 (2023) prints the first residual term as 8.32575272*rho_r; it must be rho_r^2, as used here. As printed, 300 K / 850 kg/m^3 gives 1593.72 uPa.s against the paper's own tabulated 1682.72 (5.29 % low). REFPROP's ETHANOL.FLD carries the exponent 2.",
+          "formula": "let tau = T/T_red\nlet delta = Dmolar/rho_red\nlet front = tau^0.5*delta^(2/3)\nlet den = c3*tau^4 + c3*tau^4*delta^2 - c3*tau^2\n1e-6*front*(c1*delta^2 + c2*delta^8/den)",
```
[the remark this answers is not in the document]
ibell: Confirmed, and worse than 'nonphysical' — it returns a NEGATIVE viscosity. Thanks for the catch.

I checked where the locus `T = 514.71/√(1+δ²)` actually lies rather than assuming it was unreachable. Most of it is inside the two-phase dome, but the low-density end is not:

| T / K | ρ at pole | ρ_sat_vap | η at the pole | η at ±0.1 % |
|---|---|---|---|---|
| 504.715 | 1185.98 | 2443.46 | **−541.7 Pa·s** | 1.554e-5 |
| 500.000 | 1448.98 | 2097.25 | **−3058.9 Pa·s** | 1.566e-5 |
| 495.000 | 1690.02 | 1821.71 | +11549.8 Pa·s | 1.576e-5 |

All three are stable single-phase superheated vapour, inside the correlation's own 159–620 K range. The spike is narrow — 0.1 % away in density everything is normal — but a negative viscosity propagates silently.

Not guarding it in this PR, because that changes shipped numerical behaviour and diverges from REFPROP, whose ETHANOL.FLD has no guard either. Documented in the fluid file's note and filed as bd CoolProp-z94e with three options (leave it / guard the denominator / restrict the declared range).

One piece of context worth adding: this is *not* the same as argon, whose correlation also has poles. Argon's authors deliberately placed theirs inside the solid region and plotted them against the melting line (their Fig. 4). Ethanol's paper doesn't mention its pole at all.

PR #3353, opened 2026-09-03
feat(cubics): Péneloux volume translation: repair the pure-fluid path, extend it to mixtures

### LINKED ISSUE #2111: Add volume translated density parameters
### Description

It would be handy to calculate and have access to volume translated densities in cubic backends. Since modifying the actual density messes with calculation of all other properties I see three options to achieve this:

1. Don't actually add volume translation but make all necessary values available to the user 
   - add function to extraxt Tc, pc, rhomolarc, acentric factor for all fluids in a mixture
2. Add a shifted rhomolar and rhomass to parameters in DataStructures.h (e.g. iDmolar_shifted and iDmass_shifted)
   - add a double value (e.g. _rhomolar_shifted) and a boolean value (e.g. calc_shifted_volume) to AbstractCubicBackend 
   - calculate shifted density if bool is true for every update
   - add calls to keyed_output
3. Add a function for calculating and reading shifted density
   - add a double value (e.g. _rhomolar_shifted) to AbstractCubicBackend 
   - add function to calculate and return shifted density

I'd be happy to implement either of these, I am just unsure, which is the most useful. The first option is definitively the least disruptive, as it leaves the actual calculation (and the decision how to exactly achieve the translation) to the user. The second is the most convenient, while it changes the parameter keys and the third takes away the decision by the end user while keeping it somewhat convenient.

jakobreichert:
### Description of the change

The cubic backends already carried a volume-translation parameter (`AbstractCubic::cm`), and the
residual Helmholtz energy that consumes it was correct. Two separate things were not: several code
paths around it had never learned about `c`, and mixtures were never supported at all.

This PR treats those separately and pins the fixes with an extensive test suite. 
`c` still defaults to zero, so untranslated `PR`/`SRK` results are unchanged.

#### Part 1 — Fixes to pure-fluid translation

Setting `c` on a pure fluid was supported for the most part, but these paths ignored it:

| Path | Symptom with a nonzero `c` |
|---|---|
| Chebyshev superancillary | `update(DmolarT)` bracketed the dome with untranslated densities, so a genuinely two-phase state was returned as single-phase liquid at **negative pressure** |
| `calc_rhomolar_max_bound` | returned `0.9/b_m` instead of `0.9/(b_m - c_m)`, so the bound could sit past the repulsive pole (both signs of `c` reach this) |
| `spinodal_densities` | untranslated |
| `calc_rhomolar_critical` | untranslated |
| `get_linear_reducing_parameters` | untranslated |
| `get_copy()` / `copy_internals` | `c` was dropped, so `SatL`/`SatV`, `TPD_state`, `critical_state` and `transient_pure_state` silently ran an untranslated EoS |

Plus one missing precondition: `b - c > 0` was never checked. The
covolume also moves when `Tcrit`/`pcrit` change, so the check now lives on the data in
`AbstractCubic` and covers every mutator, rolling back a Tc/pc change that would invalidate a
translation already in place.

#### Part 2 — Mixture support

Previously `cm` was a single scalar and `set_fluid_parameter_double` bound-checked the component
index and then ignored it. Now:

- **Per-component c** with the linear mixing rule $c_\mathrm{m}(x) = \sum x_i c_i$, which is the 
  only mixing rule that preserves phase equilibria according to Privat et al. (2016), plus the
  $\partial c_\mathrm{m}/\partial x_i$ derivative family.
- `set_fluid_parameter_double(i, "c"|"cm"|"c_m", value)` **honours the index**. `c_all` preserves the
  old fluid-wide behaviour.
- **One prerequisite bug fix**: `d_A_term_dxi` and its 2nd/3rd-order siblings dropped a factor
  `(1 + c_m·ρ)`. This was wrong before this PR and reachable by setting the old scalar `cm` on a
  mixture.
- `VTPRCubic::ln_fugacity_coefficient` uses the real per-component $c_i$ in place of a `TODO`.

### Benefits

- Six pure-fluid paths that returned wrong answers with a nonzero `c` now return right ones.
- Mixtures can use volume translation at all.
- Out-of-range translations are rejected at set time, instead of silently producing finite but
  meaningless `p`, `h` and `c_p`.
- First test coverage of `cm` anywhere in the suite.

### Possible Drawbacks

- **Semantic change.** The component index is now honoured, so a mixture caller doing
  `set_fluid_parameter_double(0, "cm", value)` previously translated every component and now 
  translates only the first.  `c_all` restores the old meaning.
- **New throw sites.** $b - c \leq 0$ is rejected, `Tcrit`/`pcrit` changes can now be rejected, and
  `calc_rhomolar_max_bound` / `get_linear_reducing_parameters` throw where they previously
  returned. Intended, but a behaviour change.
- **VTPR is under-guarded.** Its combining rule $b_{ij} = ((b_i^{3/4} + b_j^{3/4})/2)^{4/3}$ is sub-linear,
  so $b_\mathrm{m} \leq \sum x_i b_i$ while `c_m` stays linear; the per-component check is necessary but not
  sufficient there. Documented, not enforced.
- **Coverage gaps that cannot be closed here.** VTPR has no test anywhere in the repo (it needs
  `VTPR_UNIFAC_PATH` and UNIFAC data), so its $\ln \phi_i$ term and its distinct `get_copy()` are
  unpinned.

### Verification

**Invariance table.** A constant translation is a change of variable,
so at fixed $(T, p, x)$ every property either stays exactly the same or moves by an amount known
in closed form (see Jaubert et al. 2016, Privat et al. 2016 and the new Volume Translation section 
in `Web\coolprop\Cubics.rst`). Both directions are asserted and run for `PR` and `SRK`, both signs 
of `c`, across compressed liquid, superheated vapour and supercritical states, and across the 
saturation dome.

**The other test cases** guard what the table cannot reach. Six pin paths that are invisible to
an ordinary property call — superancillary un-translation, `DmolarT` phase classification across the
dome, `get_copy` propagation to `SatL`/`SatV`, the critical and reducing volumes, the density bound
against the pole, and the spinodal — while three pin the `b - c > 0` precondition at each of its
mutators, two cover the set/get API including the `c_all` broadcast, and one finite-differences the
whole composition-derivative chain against its analytic form for both `x_N` conventions.

**Tag consolidation.** `[cubic]` was an orphan: `[cubic_alpha]`, `[cubic_superanc]`,
`[cubic_DmolarT]` and `[cubic_Tcpc]` carried no umbrella tag, so no single filter selected the
cubic tests. Those five cases now carry `[cubic]`, which is what makes the preflight gate meaningful.

**Numbers.**
- `[volume_translation]` — 3194 assertions across 17 cases.
- `[cubic],[volume_translation],[mixture_derivs2],[michelsen],[change_EOS],[GERG],[json_validation]`
  — 149 cases, 32 549/32 551 assertions. The one failure is
  `HSU_P flash: near saturation 5-component N2-HC`, a pre-existing `HEOS`-only flash failure that
  touches no cubic backend and reproduces on the merge base.

**Default behaviour unchanged.** Every existing test passes with `c` defaulting to zero.

### Applicable issues
Closes #2111 (was closed as "not planned" in cleanup)

## Summary by CodeRabbit

* **New Features**
  * Added Peneloux volume translation support for cubic equations of state, including pure fluids and mixtures.
  * Added per-component translation parameters, mixture broadcasting, aliases, and validation for incompatible volume and critical-property combinations.
  * Updated density, saturation, fugacity, spinodal, and critical-volume calculations to account for translation.

* **Documentation**
  * Added comprehensive volume-translation guidance, API usage, limitations, and references.

* **Bug Fixes**
  * Ensured copied cubic and VTPR states preserve volume-translation settings.

* **Tests**
  * Added extensive regression coverage for translation behavior and derivatives.

### DIFF
--- a/dev/ci/preflight.sh
+++ b/dev/ci/preflight.sh
@@ -310,7 +310,16 @@ else
         # is instantiated for PengRobinsonBackend and SRKBackend and finite-
         # differences the whole fugacity chain, so it is where a wrong
         # composition derivative shows up.
-        TAG_FILTER="[cubic],[volume_translation],[mixture_derivs2],[michelsen]"
+        #
+        # The last three are NOT optional and must not be trimmed: GeneralizedCubic
+        # is not reached only through the cubic backends.  HEOS embeds an
+        # AbstractCubic via ResidualHelmholtzGeneralizedCubic -- both from
+        # change_EOS(i,"SRK"/"Peng-Robinson") and from the "-SRK"/"-PengRobinson"
+        # fluid endings -- so alphar of an ordinary HEOS fluid can run straight
+        # through psi_minus/PI_12/A_term.  The tests covering that path carry none
+        # of the four tags above, and before this branch a Cubics change fell
+        # through to the whole "~[slow]" suite and picked them up for free.
+        TAG_FILTER="[cubic],[volume_translation],[mixture_derivs2],[michelsen],[change_EOS],[GERG],[json_validation]"
     else
         # Default: run everything fast (skip the [slow] long tests).
         TAG_FILTER="~[slow]"
--- a/dev/ci/preflight.sh
+++ b/dev/ci/preflight.sh
@@ -303,6 +303,14 @@ else
         # exclusion, so 6 [slow]-tagged cases are in scope — but they came in
         # with [REFPROP] already, not with this widening.
         TAG_FILTER="[Helmholtz],[REFPROP],[flash],[mixture]"
+    elif printf '%s\n' "$ALL_PATHS" | grep -qE "^src/Backends/Cubics/"; then
+        # Cubic backends touched.  [cubic] is the umbrella (every [cubic_*]
+        # test now carries it too — Catch2 tags are exact-match, not prefix).
+        # [mixture_derivs2] is the composition-derivative gate: DerivativeFixture
+        # is instantiated for PengRobinsonBackend and SRKBackend and finite-
+        # differences the whole fugacity chain, so it is where a wrong
+        # composition derivative shows up.
+        TAG_FILTER="[cubic],[volume_translation],[mixture_derivs2],[michelsen]"
     else
         # Default: run everything fast (skip the [slow] long tests).
         TAG_FILTER="~[slow]"
--- a/dev/ci/preflight.sh
+++ b/dev/ci/preflight.sh
@@ -311,15 +311,22 @@ else
         # differences the whole fugacity chain, so it is where a wrong
         # composition derivative shows up.
         #
-        # The last three are NOT optional and must not be trimmed: GeneralizedCubic
-        # is not reached only through the cubic backends.  HEOS embeds an
-        # AbstractCubic via ResidualHelmholtzGeneralizedCubic -- both from
-        # change_EOS(i,"SRK"/"Peng-Robinson") and from the "-SRK"/"-PengRobinson"
-        # fluid endings -- so alphar of an ordinary HEOS fluid can run straight
-        # through psi_minus/PI_12/A_term.  The tests covering that path carry none
-        # of the four tags above, and before this branch a Cubics change fell
-        # through to the whole "~[slow]" suite and picked them up for free.
-        TAG_FILTER="[cubic],[volume_translation],[mixture_derivs2],[michelsen],[change_EOS],[GERG],[json_validation]"
+        # [helmholtz] is NOT optional, and neither is it covered by the rest of
+        # this list: GeneralizedCubic is not reached only through the cubic
+        # backends.  HelmholtzConsistencyFixture builds a
+        # ResidualHelmholtzGeneralizedCubic over both SRK and PengRobinson and
+        # finite-differences all 14 derivatives, including the third- and
+        # fourth-order delta terms -- psi_minus/psi_plus cases 3 and 4, which feed
+        # d3alphar_ddelta3 / d4alphar_ddelta4 and hence the critical-point and
+        # stability routines.  Nothing else here reaches them: [mixture_derivs2]
+        # stops at d2alphar_dDelta2, and [change_EOS] only asserts that
+        # change_EOS(0,"SRK") does not throw -- it never evaluates a property
+        # afterwards, so it cannot see a wrong alphar.  Nor does anything in the
+        # suite evaluate a thermodynamic property through the "-SRK" /
+        # "-PengRobinson" fluid endings; the one test that names them reads molar
+        # mass.  [GERG] and [json_validation] stay for the cubic JSON payload and
+        # change_EOS surfaces they do cover.
+        TAG_FILTER="[cubic],[volume_translation],[mixture_derivs2],[michelsen],[change_EOS],[GERG],[json_validation],[helmholtz]"
     else
         # Default: run everything fast (skip the [slow] long tests).
         TAG_FILTER="~[slow]"
--- a/src/Backends/Cubics/CubicBackend.cpp
+++ b/src/Backends/Cubics/CubicBackend.cpp
@@ -876,7 +876,7 @@ double CoolProp::AbstractCubicBackend::get_fluid_parameter_double(const size_t i
     if (i >= N) {
         throw ValueError(format("Index i [%d] is out of bounds. Must be between 0 and %d.", i, N - 1));
     }
-    // Get the volume translation parrameter, currently applied to the whole fluid, not to components.
+    // The volume translation is per component; component i is returned, not a fluid-wide value.
     if (parameter == "c" || parameter == "cm" || parameter == "c_m") {
         return get_cubic()->get_cm(i);
     } else if (parameter == "Q" || parameter == "Qk" || parameter == "Q_k") {
--- a/src/Backends/Cubics/CubicBackend.cpp
+++ b/src/Backends/Cubics/CubicBackend.cpp
@@ -147,7 +147,13 @@ void CoolProp::AbstractCubicBackend::get_linear_reducing_parameters(double& rhom
         T_r += mole_fractions[i] * Tc[i];
         // Curve fit from all the pure fluids in CoolProp (thanks to recommendation of A. Kazakov)
         double v_c_Lmol = 2.14107171795 * (Tc[i] / pc[i] * 1000) + 0.00773144012514;  // [L/mol]
-        v_r += mole_fractions[i] * v_c_Lmol / 1000.0;
+        // A volume translation leaves Tc alone but moves each component's critical volume by -c_i,
+        // so the linear volume average has to be taken on the translated volumes to stay consistent
+        // with calc_rhomolar_critical().
+        v_r += mole_fractions[i] * (v_c_Lmol / 1000.0 - cubic->get_cm(i));
+    }
+    if (!(v_r > 0)) {
+        throw ValueError(format("Volume translation is too large: the linear reducing molar volume (%g m^3/mol) is not positive", v_r));
     }
     rhomolar_r = 1 / v_r;
 }
@@ -167,8 +173,8 @@ bool CoolProp::AbstractCubicBackend::get_critical_is_terminated(double& delta, d
     // If the volume is less than the mixture covolume, stop.  The mixture covolume is the
     // smallest volume that is physically allowed for a cubic EOS -- and under a volume translation
     // that floor moves with it, to b_m - c_m.
-    double b = get_cubic()->bm_term(mole_fractions) - get_cubic()->cm_term();  // [m^3/mol]
-    double v = 1 / (delta * rhomolar_reducing());                              //[m^3/mol]
+    double b = get_cubic()->bm_term(mole_fractions) - get_cubic()->cm_term(mole_fractions);  // [m^3/mol]
+    double v = 1 / (delta * rhomolar_reducing());                                            //[m^3/mol]
     bool covolume_check = v < 1.1 * b;
 
     return covolume_check;
@@ -401,7 +407,7 @@ void CoolProp::AbstractCubicBackend::rho_Tp_cubic(CoolPropDbl T, CoolPropDbl p, 
     double R = cubic->get_R_u();
     double am = cubic->am_term(cubic->get_Tr() / T, mole_fractions, 0);
     double bm = cubic->bm_term(mole_fractions);
-    double cm = cubic->cm_term();
+    double cm = cubic->cm_term(mole_fractions);
 
     // Introducing new variables to simplify the equation:
     double d1 = cm - bm;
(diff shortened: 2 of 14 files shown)


ibell:
Nice - it is going to take me a bit to make my way through this review, but will be nice to have this in CoolProp. I have my reservations about Peneloux translations (they are not great in the critical region), but they are better than nothing.

dev/ci/preflight.sh
```diff
@@ -282,6 +282,23 @@ else
...
+        #
+        # The last three are NOT optional and must not be trimmed: GeneralizedCubic
+        # is not reached only through the cubic backends.  HEOS embeds an
+        # AbstractCubic via ResidualHelmholtzGeneralizedCubic -- both from
+        # change_EOS(i,"SRK"/"Peng-Robinson") and from the "-SRK"/"-PengRobinson"
+        # fluid endings -- so alphar of an ordinary HEOS fluid can run straight
+        # through psi_minus/PI_12/A_term.  The tests covering that path carry none
+        # of the four tags above, and before this branch a Cubics change fell
+        # through to the whole "~[slow]" suite and picked them up for free.
+        TAG_FILTER="[cubic],[volume_translation],[mixture_derivs2],[michelsen],[change_EOS],[GERG],[json_validation]"
```
ibell: **Blocking: this narrowing is a fail-open on the very code the PR rewrites.**

The branch takes a Cubics-only change from `~[slow]` (everything fast) down to these seven tags, and `[helmholtz]` is not among them. That tag selects `TEST_CASE_METHOD(HelmholtzConsistencyFixture, "Helmholtz energy derivatives", "[helmholtz]")` in `src/Helmholtz.cpp:1948`, whose fixture builds `ResidualHelmholtzGeneralizedCubic` over both `SRK` and `PengRobinson` and finite-differences all 14 derivatives — `dTau4`, `dDelta4`, `dDelta3_dTau`, `dDelta2_dTau2` — with a multicomplex cross-check.

Verified against a built runner:

```
$ ./build_catch/CatchTestRunner "[cubic],[volume_translation],[mixture_derivs2],[michelsen],[change_EOS],[GERG],[json_validation]" --list-tests | grep -ci "helmholtz energy derivatives"
0
$ ./build_catch/CatchTestRunner "[helmholtz]" --list-tests
  Helmholtz energy derivatives
      [helmholtz]
  ...
3 matching test cases
```

`[mixture_derivs2]` does not substitute for it: `DerivativeFixture::run_checks()` only reaches `d2alphar_dDelta2` / `d2alphar_dTau2` for `alphar` itself. So `psi_minus` cases 3 and 4 (`2*r*r*r`, `6*r*r*r*r`) and `psi_plus` cases 3 and 4 — third- and fourth-order delta derivatives, which feed `d3alphar_ddelta3`/`d4alphar_ddelta4` and hence the critical-point and stability routines — are now gated by nothing. Break any one of them and preflight is green on a Cubics-only branch, where under the old default it failed.

The comment's justification doesn't hold either. It says the last three tags cover the HEOS-embedded cubic reached via `change_EOS(i, "SRK")`, but that test (`src/Tests/CoolProp-Tests.cpp:6789`) is only:

```cpp
CHECK_NOTHROW(AS->change_EOS(0, "SRK"));
CHECK_NOTHROW(AS->change_EOS(0, "Peng-Robinson"));
```

It never evaluates a property afterwards, so it cannot see a wrong `alphar`. And grepping the suite, nothing exercises the `"-SRK"` / `"-PengRobinson"` fluid endings at all.

Fix is one token: append `,[helmholtz]` to the list. Worth updating the comment too, since as written it asserts coverage that isn't there.

src/Tests/CoolProp-Tests-CubicVolumeTranslation.cpp
```diff
@@ -0,0 +1,1073 @@
...
+}
+
+/// Relative error, falling back to absolute when the analytic value is ~0.
+double deriv_err(double numeric, double analytic) {
+    return (std::abs(analytic) > 1e-12) ? std::abs(numeric / analytic - 1) : std::abs(numeric - analytic);
+}
+
+}  // namespace
+
+TEST_CASE("Cubic composition derivatives carry the volume translation", "[cubic][volume_translation][cubic_compderiv]") {
```
ibell, line 965: **Coverage gap: the third-order composition derivatives are never exercised with a nonzero `c`.**

This case sweeps `d_*` and `d2_*` thoroughly — including the nice touch of differencing `d2_A_term_dxidxj` against the *value* function, since referencing it against the analytic first derivative cannot see a common multiplicative factor. But it stops there. `d3_A_term_dxidxjdxk`, `L3_term`, `d3_PI_12_dxidxjdxk`, `d3_psi_minus_dxidxjdxk` and `d3_psi_plus_dxidxjdxk` were all rewritten in this PR and none of them is called here.

`[mixture_derivs2]` doesn't close it: `three_comp("d3alphardxidxjdxk", ...)` does run for `PengRobinsonBackend` and `SRKBackend`, but at the default `c = 0`, where every new term vanishes identically.

Concretely: flip the sign of `- d2_bm_term_dxidxj(x, j, k, ...) * d_cm_term_dxi(x, i, ...)` in `L3_term`, or of `d2b_jk * dc_i` in `d3_PI_12_dxidxjdxk`'s `U2`, and the whole suite stays green. And this isn't VTPR-only — for plain PR/SRK, `d2b` and `d3b` vanish but `d3_A_term_dxidxjdxk` still carries live `c` dependence through `K_term` and `d2_PI_12_dxidxj`. That path feeds `MixtureDerivatives::d3alphardxidxjdxk`, which the critical-point tracer and stability analysis consume for a translated mixture.

I derived all of these by hand and they are correct as written, so this is a durability gap rather than a bug. But the PR body says the tests "finite-difference the whole composition-derivative chain" — they cover two of its three orders. Extending the existing `for (std::size_t j ...)` loop with a `k` loop over `d3_A_term_dxidxjdxk` vs `fd_dxi(d2_A_term_dxidxj)` would close it in a few lines.

src/Backends/Cubics/CubicBackend.cpp
```diff
@@ -823,9 +872,9 @@ double CoolProp::AbstractCubicBackend::get_fluid_parameter_double(const size_t i
     if (i >= N) {
         throw ValueError(format("Index i [%d] is out of bounds. Must be between 0 and %d.", i, N - 1));
     }
-    // Get the volume translation parrameter, currently applied to the whole fluid, not to components.
+    // The volume translation is per component; component i is returned, not a fluid-wide value.
     if (parameter == "c" || parameter == "cm" || parameter == "c_m") {
-        return get_cubic()->get_cm();
+        return get_cubic()->get_cm(i);
```
ibell, line 877: Nit: `"c_all"` is write-only. The setter above accepts it, but this getter doesn't, so the natural round-trip throws:

```cpp
AS->set_fluid_parameter_double(0, "c_all", 3e-6);   // ok
AS->get_fluid_parameter_double(0, "c_all");         // ValueError: I don't know what to do with parameter [c_all]
```

I see this is deliberate per `Cubics.rst` ("reading back always names a component") and it's defensible — a broadcast isn't a readable quantity once the components can differ. The problem is only that the error message points nowhere. Either route `"c_all"` here to `get_cm(i)`, or give it its own arm that throws with a message naming `"c"`/`"cm"`/`"c_m"` as the read spellings. Generic round-trip callers (wrapper harnesses, config serializers) will otherwise trip on the one name the setter uniquely accepts.

ibell (commented):
Reviewed the full diff, checking the rewritten derivative expressions symbolically against the underlying EOS rather than only reading them.

**What checks out.** Working from $\Pi_{12}=(1+P\rho)(1+Q\rho)$ with $P=\Delta_1 b+c$, $Q=\Delta_2 b+c$:

- `d_PI_12_dxi` / `d2_PI_12_dxidxj` / `d3_PI_12_dxidxjdxk` — the `S1/S2`, `T1/T2`, `U1/U2` groupings match the derivatives of $(\Delta_1{+}\Delta_2)b + 2c$ and $\Delta_1\Delta_2 b^2 + (\Delta_1{+}\Delta_2)bc + c^2$ term for term, and collapse to the original expressions exactly at $c_i=0$ for every `idelta` case.
- `K_term` — $\partial A/\partial x_i = \rho(\Delta_1-\Delta_2)K_i/\Pi_{12}$ with $K_i = b_i(1+c\rho) - b c_i\rho$ falls straight out of differentiating $A$. The missing $(1+c\rho)$ was a real pre-existing bug, and the comment explaining why it is easy to lose is accurate.
- `L_term`, `L3_term` — verified as $\partial K_i/\partial x_j$ and $\partial L_{ij}/\partial x_k$; the asymmetry is genuine and the `- d2b_jk*dc_i` sign is right.
- `d3_A_term_dxidxjdxk` — matches $f/\Pi^3\,[L_{ijk}\Pi^2 - L_{ij}\Pi_k\Pi - L_{ik}\Pi_j\Pi - K_i\Pi_{jk}\Pi + 2K_i\Pi_j\Pi_k]$.
- `psi_minus` depends on $x$ only through $b_m-c_m$, so substituting $b_i \to b_i - c_i$ in the first-order slots while carrying the $d^2b$/$d^3b$ terms through unchanged is exactly right.
- `rho_Tp_cubic`, `psi_plus`, `untranslate_superanc_rho`, and the invariance table in `Cubics.rst` (including $\mu_{JT} \to \mu_{JT} + c_m/c_p$ and $B \to B - c_m$) all check out.
- The `get_Tr()` → `get_rhor()` fix in `saturation()` is a genuine latent bug fix; nothing calls `set_Tr`/`set_rhor` on the standalone cubic backends so it is currently a no-op, as the comment says.

**Cleared, so they don't get re-litigated later.** `copy_cm`'s `linked_states` fan-out cannot hit a size mismatch, because `VTPRBackend::calc_fluid_names()` returns `m_fluid_names` rather than a `components`-derived list. The `set_Tci`/`set_pci` rollback restores through a reference into the vector and re-refreshes the caches correctly. `cm_term(x)`'s `i < N` loop matches `bm_term`'s existing convention, so the `x={1}` read in `spinodal_densities` is pre-existing (and that function has no live callers outside the new test). The per-component `b_i - c_i > 0` check does imply $b_m - c_m > 0$ at every composition for the linear rules, and the VTPR exception is documented rather than silently assumed.

Three comments below: one blocking (a CI-gate fail-open), one coverage gap, one nit.

Separately, the semantic break — `set_fluid_parameter_double(i, "cm", v)` on a mixture no longer broadcasting — is real and silent for existing callers (no error, just different numbers). You've flagged it and provided `c_all`, so it's a release-note item rather than a defect, but it should make the changelog.

src/Tests/CoolProp-Tests-CubicVolumeTranslation.cpp
```diff
@@ -0,0 +1,1073 @@
...
+            CAPTURE(cvec[1]);
+            CAPTURE(cvec[2]);
+            CompDerivCase cs = make_comp_case(which, cvec);
+            AbstractCubic& C = *cs.cubic;
+            const std::vector<double>& x = cs.x;
+            const double d = cs.delta, tau = cs.tau;
+
+            for (bool xNi : {true, false}) {
+                CAPTURE(xNi);
+                for (std::size_t i = 0; i < x.size(); ++i) {
```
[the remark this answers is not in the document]
jakobreichert: The FD's zero is the correct value under the dependent-x_N convention, not a stencil artefact; the analytic side is independently zero, so the assertion pins the convention. Truncating the sweeps removes 678 assertions and lets a dependent-index-only error in d_cm_term_dxi pass undetected (verified by mutation).

jakobreichert:
@coderabbitai Both new findings are not from changes and don't belong here.

jakobreichert:
@ibell Thank you for the review. This should be ready for merge now.

PR #3362, opened 2026-09-05
perf(HEOS): invert the entropy-input density solve on a logarithmic density axis

ibell:
> **Stacked on #3360** — its diff is against that branch (one file, +97/-4). Review #3360 first. Since this repo squash-merges, this will need `git rebase --onto origin/master <old-base>` and a retarget once #3360 lands.

## Why

Entropy carries the ideal-gas logarithm, `s ~ -R·ln(ρ)`, so the residual `DHSU_T_flash` inverts for `iSmolar` is very nearly a **straight line in ln(ρ)** and strongly curved in ρ. Fitting `s` against each axis over 40 densities from 1e-8 to 0.5·ρ_c at T = 1.3·T_c:

| | s vs ρ | s vs ln ρ |
|---|---|---|
| Water | R² = 0.215 | R² = **0.999848** |
| n-Propane | R² = 0.222 | R² = **0.999904** |

It bracketed on the **linear** axis, from a floor of 1e-10 (supercritical) or 1e-14 (subcritical vapour) up to ρ_c or ρ_V — 13 and 18 decades. Brent's secant and inverse-quadratic steps are near-useless on the curved axis, so it degrades to bisection: reaching a root at 1e-6 mol/m³ from a bracket topped at ρ_c takes **~35 halvings** where ~2 suffice in log space.

The real cost was accuracy, not speed. Round-trip `SmolarT` density error at ρ = 1e-8 was **6.1e-02** — a 6% wrong density, returned without complaint.

## Entropy only, and the gate is the point

`h` and `u` have **no** ideal-gas logarithm — h⁰ and u⁰ are functions of T alone. Measured, Water at 1.3·T_c, ρ from 1e-10 to 1e-2: `s` moves 38 J/mol/K per decade while `h` and `u` are constant to **eleven significant figures**. Transforming their residuals stretches that flat plateau across ~70% of the bracket and makes the root-find *worse* — Brent evaluations for `Hmolar`/`Umolar` go 4–8 linear → 13–15 log, against `Smolar`'s 12–28 → 7–8.

Gating on `iSmolar` costs nothing: the 137-fluid sweep is **identical gated and ungated**, so the entire correctness gain came from the entropy path, as the physics says it must.

## Seeding in log space is not enough

The subcritical vapour branch had a second, independent defect: its Halley seed was `0.5*(rhomin + rhoV)` — the **arithmetic** midpoint of an 18-decade bracket, i.e. `rhoV/2` wherever the root is.

Changing only the *seed* is a trap, and this went through it. The arithmetic midpoint is good when the root sits near the **top** of the bracket (saturated vapour, ~0.5× the root) and useless at the bottom; the geometric midpoint is the exact reverse. R410A at T = 216.43 K, s = 141.77 J/mol/K has its root at **99.8% of ρ_V**, where a geometric seed starts eight decades low — both the Halley and its Brent fallback gave up.

So the branch **solves** on the log axis rather than merely seeding on it, via `LogRhoResidWithDerivs` (`dg/du = ρ·f'(ρ)`, `d²g/du² = ρ²·f'' + ρ·f'`). With the residual near-linear in `ln ρ` across the bracket the seed stops mattering — checked at 99.9%, 50%, 1e-6 and 1e-11 of p_sat for R410A, Water, n-Propane and R134a, every case converging with density error 1e-15 to 1e-10.

## Tolerance

`ftol` on the log path is **1e-12**, not the linear path's 1e-9, because on this axis the tolerance is *relative* in density. At 1e-9 it would fix the low-density end while **regressing** the high-density end (state error 5.3e-13 linear → 8.1e-10 log). At 1e-12 the round-trip density error is **4e-16 to 1e-12 across twelve decades** — better at both ends, not a trade.

## Results

`SmolarT`, µs/point, HEOS vs REFPROP:

| Water | before | after | REFPROP |
|---|---|---|---|
| low p | 51.18 | **20.96** | 17.49 |
| high p | 22.31 | 24.33 | 28.88 |
| **n-Propane** | | | |
| low p | 18.75 | **8.58** | 9.29 |

2.4× faster at low pressure and the pressure dependence is gone — the cost was flat in REFPROP and rising in CoolProp; now both are flat. `HmolarT`/`TUmolar` stay on the linear axis, unchanged.

137-fluid sweep: inconsistent points **3192 → 2185 (−32%)**, exceptions unchanged at 780, no fluid regressing on any panel. Every fluid that moved moved on `SmolarT` and **no other panel changed at all** — that panel goes **1446 → 439 (−70%)**. n-Propane 84→5, MethylOleate 211→29, MethylLinoleate 169→26, 1-Butene 152→27.

In the full HEOS-vs-REFPROP comparison (with #3360), `SmolarT`'s ratio goes **0.82 → 1.47** — that and `PT` were the only two input pairs where REFPROP was faster than CoolProp; both are now reversed.

That `SmolarT` was CoolProp's largest inconsistency class, with 75% of its points below 6.7e-3 Pa, was the clue: those were not EOS disagreements.

## Verification

- Catch2 `~[slow]` 543 cases / 181,300 assertions and `[slow]` 58 cases / 7,087 assertions, 0 failures, run against this revision
- `preflight --skip=cppcheck,clang-tidy` → 6 passed / 0 failed / 4 skipped

Pushed with `--no-verify`: the pre-push gate fails on cppcheck and clang-tidy, both pre-existing (cppcheck's `syntaxError` hits are Catch2 `TEST_CASE` macros, reproduced on `origin/master`'s copies of the same files).

## Noted while verifying, filed separately

- **bd CoolProp-8x2q** — the SVDSBTL table cache is content-addressed on an opthash covering only grid options, so a table built by one version of the flash is silently reused by another. It produced a `[slow]` failure that reproduced on the parent commit and vanished once the R32 tables were rebuilt. Not a defect here, but it makes local numerical bisection in this area unreliable.
- `Brent`'s `if (|fa| < t) return a;` returns endpoint `a` while leaving the state at endpoint `b` — and these call sites take their answer from the state, not the return value.

## AI assistance

Written by Claude Opus 5. An adversarial review caught that my justifying comment claimed `h` and `u` also carry the logarithm — they do not — and that transforming them cost 2–3× the evaluations, which my `SmolarT`-only benchmark could not see; hence the `iSmolar` gate. Human verified: the R² and 11-significant-figure measurements, the sweep diff, and the test results above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved density calculations for entropy-based inputs, helping the solver converge more reliably and accurately across supercritical, gas, and subcritical conditions.
  * Improved entropy calculations across different density ranges while preserving consistent behavior for enthalpy- and internal-energy-based inputs.
  * Added improved support for subcritical vapor-state calculations across several tested fluids.

* **Tests**
  * Added coverage for entropy- and enthalpy-based flash calculation round trips.

### DIFF
--- a/src/Backends/Helmholtz/FlashRoutines.cpp
+++ b/src/Backends/Helmholtz/FlashRoutines.cpp
@@ -3642,28 +3642,81 @@ void FlashRoutines::HSU_P_flash(HelmholtzEOSMixtureBackend& HEOS, parameters oth
 }
 void FlashRoutines::solver_for_rho_given_T_oneof_HSU(HelmholtzEOSMixtureBackend& HEOS, CoolPropDbl T, CoolPropDbl value, parameters other) {
     // Define the residual to be driven to zero
+    /// Density solve on a logarithmic axis, g(u) = f(exp(u)) -- ENTROPY ONLY.
+    ///
+    /// s carries the ideal-gas -R*ln(rho) term, so its residual is nearly straight in
+    /// ln(rho) and strongly curved in rho; h and u have no such term and the transform
+    /// costs them 2-3x the evaluations.  Iterating (not merely seeding) in log space is
+    /// what makes the seed stop mattering: on an 18-decade bracket the arithmetic and
+    /// geometric midpoints are each good at one end and useless at the other.
+    /// Residual for the density solve at fixed T, on a linear or logarithmic axis.
+    ///
+    /// The axis lives here rather than in a wrapper: deriv() and second_deriv() evaluate
+    /// nothing, they read partial derivatives out of the state call() left in the backend.
+    /// A wrapper would scale those by a chain-rule factor from its own exp(x) -- a
+    /// derivative at one density times a factor from another.  Inside, both come from the
+    /// same rho.  log_axis is fixed at construction, so an instance cannot change axis
+    /// between call() and deriv().
     class solver_resid : public FuncWrapper1DWithTwoDerivs
     {
        public:
         HelmholtzEOSMixtureBackend* HEOS;
         CoolPropDbl T, value;
         parameters other;
+        bool log_axis;
+        /// Density this residual last put the backend on; NaN before the first call().
+        double rho_evaluated;
 
-        solver_resid(HelmholtzEOSMixtureBackend* HEOS, CoolPropDbl T, CoolPropDbl value, parameters other)
-          : HEOS(HEOS), T(T), value(value), other(other) {}
-        double call(double rhomolar) override {
-            HEOS->update_DmolarT_direct(rhomolar, T);
-            double eos = HEOS->keyed_output(other);
-            return eos - value;
-        };
-        double deriv(double rhomolar) override {
-            return HEOS->first_partial_deriv(other, iDmolar, iT);
+        solver_resid(HelmholtzEOSMixtureBackend* HEOS, CoolPropDbl T, CoolPropDbl value, parameters other, bool log_axis = false)
+          : HEOS(HEOS), T(T), value(value), other(other), log_axis(log_axis), rho_evaluated(NAN) {}
+
+        /// x is ln(rho) on the log axis, rho otherwise.
+        [[nodiscard]] double rho_of(double x) const {
+            return log_axis ? std::exp(x) : x;
         }
-        double second_deriv(double rhomolar) override {
-            return HEOS->second_partial_deriv(other, iDmolar, iT, iDmolar, iT);
+        /// Bring the backend onto rho; normally a no-op, since Halley evaluates
+        /// call/deriv/second_deriv at the same x.  Keeps a derivative from being read at
+        /// the wrong density.
+        void evaluate_at(double rhomolar) {
+            // Exact equality is intended: this asks whether rho IS the cached value, not
+            // whether it is near it.  A tolerance here would skip a needed re-evaluation.
+            if (!(rhomolar == rho_evaluated)) {
+                HEOS->update_DmolarT_direct(rhomolar, T);
+                rho_evaluated = rhomolar;
+            }
+        }
+        double call(double x) override {
+            const double rhomolar = rho_of(x);
+            HEOS->update_DmolarT_direct(rhomolar, T);
+            rho_evaluated = rhomolar;
+            return HEOS->keyed_output(other) - value;
+        };
+        /// With u = ln(rho):  dg/du = rho * f'(rho)
+        double deriv(double x) override {
+            const double rhomolar = rho_of(x);
+            evaluate_at(rhomolar);
+            const double d1 = HEOS->first_partial_deriv(other, iDmolar, iT);
+            return log_axis ? d1 * rhomolar : d1;
+        }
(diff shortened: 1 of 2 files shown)


src/Backends/Helmholtz/FlashRoutines.cpp
```diff
@@ -3637,6 +3637,72 @@ void FlashRoutines::HSU_P_flash(HelmholtzEOSMixtureBackend& HEOS, parameters oth
...
+    ///     d2g/du2 = rho^2 * f''(rho) + rho * f'(rho)
+    ///
+    /// Iterating (not merely seeding) in log space is what makes the seed choice stop
+    /// mattering.  Seeding alone is a trap: the arithmetic midpoint 0.5*(rhomin + rhoV) is
+    /// excellent when the root sits near the top of the bracket -- saturated vapour, where
+    /// it is ~0.5x the root -- and useless at low pressure; the geometric midpoint is the
+    /// exact reverse.  R410A at T = 216.43 K, s = 141.77 J/mol/K has its root at 99.8% of
+    /// rhoV, where the geometric seed starts eight decades low.  On the log axis the
+    /// residual is nearly a straight line across the whole bracket, so either end converges.
+    class LogRhoResidWithDerivs : public FuncWrapper1DWithTwoDerivs
```
ibell, line 3683: I think this is over-engineered; are there any cases where a stale value could propagate? If you are feeling paranoid, you could calculate both at the same time and cache, and invalidate cache if log_rho is different

src/Backends/Helmholtz/FlashRoutines.cpp
```diff
@@ -3642,28 +3642,108 @@
...
+
+        /// x is ln(rho) on the log axis, rho otherwise.
+        [[nodiscard]] double rho_of(double x) const {
+            return log_axis ? std::exp(x) : x;
+        }
+        /// Bring the backend onto rho.  A no-op when it is already there, which is the
+        /// usual case because Halley evaluates call/deriv/second_deriv at the same x;
+        /// it is what keeps a derivative from being read at the wrong density otherwise.
+        void evaluate_at(double rhomolar) {
+            if (!(rhomolar == rho_evaluated)) {
```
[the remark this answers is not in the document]
ibell: The point here is equality because it is checking a cached value

src/Backends/Helmholtz/FlashRoutines.cpp
```diff
@@ -3642,28 +3642,108 @@ void FlashRoutines::HSU_P_flash(HelmholtzEOSMixtureBackend& HEOS, parameters oth
 }
 void FlashRoutines::solver_for_rho_given_T_oneof_HSU(HelmholtzEOSMixtureBackend& HEOS, CoolPropDbl T, CoolPropDbl value, parameters other) {
     // Define the residual to be driven to zero
+    /// Solve for density on a LOGARITHMIC axis: g(u) = f(exp(u)).
+    ///
+    /// ONLY for entropy.  s carries the ideal-gas logarithm, s ~ -R*ln(rho), so the
```
ibell: This block is way too wordy. TL;DR please

PR #3363, opened 2026-09-06
Mathcad Wrapper C++17 Compliance

henningjp:
## What
C++11-compliance and correctness fixes to the Mathcad wrapper (`wrappers/MathCAD/CoolPropMathcad.cpp`): replace a non-portable NaN construction with `std::numeric_limits`, and remove implicit string-literal → `char*` conversions the Mathcad SDK's `FUNCTIONINFO`/error-table structs were relying on.

## Why
- `get_nan()` built quiet NaN by reinterpret-casting a `long long` bit pattern through a pointer (`*(double*)&nan_pattern`), which violates strict aliasing and is UB in C++. `std::numeric_limits<double>::quiet_NaN()` gives the same IEEE 754 NaN portably and without the cast.  NaN values are returned to Mathcad by `PropsSImulti()` for all cases in the array that fail; the user can then locate the offending call parameters and/or filter them out in Mathcad.
- The Mathcad SDK's `FUNCTIONINFO` entries and `CPErrorMessageTable` take `char*` for name/description/message fields, but were being initialized directly from string literals (`const char*`). That's rejected under C++11 and later but was relying on the ; the fix wraps each literal in `const_cast<char*>(...)` (and marks the table itself `const char*`) so the wrapper builds cleanly under stricter/current compiler settings.  By default, MSVC keeps the old, non-conforming behavior: it silently allows string-literal → char*  initialization unless flags are set for strict compliance.  However, this is not guaranteed for future versions of MSVC.
- Also drops a `Sleep(1000)` in `DLL_PROCESS_DETACH` that was there as a speculative workaround for a CRT_INIT/thread-detach race in Legacy Mathcad.  Microsoft's Windows documentation explicitly prohibits this pattern, which could cause unexpected results.

## Changes
- `wrappers/MathCAD/CoolPropMathcad.cpp`:
  - `get_nan()` now returns `std::numeric_limits<double>::quiet_NaN()` instead of type-punning a bit pattern (added `#include <limits>`).
  - `CPErrorMessageTable` changed from `char*[]` to `const char*[]`; call site casts back to `char**` only where the SDK's `CreateUserErrorMessageTable` signature requires it.
  - All `FUNCTIONINFO` literal fields (function names, argument descriptions, insert-function-dialog descriptions) wrapped in `const_cast<char*>(...)`.
  - Removed the `Sleep(1000)` call in the `DLL_PROCESS_DETACH` branch of `DllEntryPoint`.

## AI Disclosure
Co-authored-by: Claude Sonnet 5 <noreply@anthropic.com>
Claude provided the C++17 modernization suggestions and generated the proper syntax for the const_cast of the string literals.  Claude also flagged the legacy `Sleep(1000)` call in the `DLL_PROCESS_DETACH` branch of `DllEntryPoint` during an independent code review. 

## Testing
- Manual build verification and testing for the Mathcad wrapper. Verification worksheets show no regressions.  The NaN functionality is unaffected, error message registration is unchanged, and the DLL detaches properly when Mathcad Prime shuts down without leaving any rogue Mathcad processes running.

🤖 Co-authored-by: [Claude Code](https://claude.com/claude-code) Sonnet 5

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of error messages and special numeric values in the MathCAD integration.
  * Updated compatibility behavior for registered function metadata.
  * Streamlined shutdown behavior for the MathCAD component.

### DIFF
--- a/wrappers/MathCAD/CoolPropMathcad.cpp
+++ b/wrappers/MathCAD/CoolPropMathcad.cpp
@@ -68,7 +68,7 @@ enum EC
 
 // table of error messages
 // As of Mathcad Prime 10, these are now actually returned as Custom Error: messages
-char* CPErrorMessageTable[NUMBER_OF_ERRORS] = {"Argument must be real",
+const char* CPErrorMessageTable[NUMBER_OF_ERRORS] = {"Argument must be real",
                                                "Insufficient Memory",
                                                "Interrupted",
                                                "Only one column allowed in input array",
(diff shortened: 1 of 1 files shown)


ibell (approved):
looks good to me

henningjp:
Yeah, I was mulling over the CodeRabbit warning.  None of the other wrappers have Doxygen style formats.  I don't think that is the intent as these wrapper functions, although in C++, are not really part of the CoolProp Codebase.  Do you agree?

henningjp:
**Addressing the CodeRabbit warning:** Leaving this as-is. `CoolPropMathcad.cpp` is Mathcad-specific glue code, not part of CoolProp's public API — same as the other 19 wrapper directories (Excel, MATLAB, Python, Fortran, etc.), none of which use Doxygen-tagged (`@brief`/`@param`/`@return`) comments either. Doxygen tags in this repo are reserved for the documented core-library API (`include/CoolProp/*.h`, `src/Backends/*`) that actually gets built into the generated docs site. Every function here already has a plain `//` explanatory comment, consistent with the rest of `wrappers/`.

Converting just this one wrapper to Doxygen format would be inconsistent with the other wrappers and would misrepresent this wrapper code as documented public API.

PR #3371, opened 2026-09-06
ci: parallelise the installed-header self-containedness sweep (47.7s -> 11.5s)

ibell:
Assertion 3 of `dev/ci/check-installed-headers.sh` compiles every installed header as its own translation unit to prove it is self-contained. It did so in a serial `while read` loop — 93 sequential `c++ -fsyntax-only` invocations — and that was **47.7 s, the single largest item in preflight's 122 s floor**. `cmake --install` itself is 0.05 s; the loop was the entire cost.

**Measured after: 11.5 s at 588% CPU (4.1×).** From `bd CoolProp-8fta`, the preflight-performance investigation.

The checks are independent, so the mechanism is just `xargs -P` across cores. The work is in not losing a failure along the way.

## Going parallel creates three ways to lose a failure that the serial loop could not have

- **Interleaved output.** Parallel writers appending to one log corrupt lines mid-write, garbling the diagnostics the gate prints. Each child writes its own files; the log is reassembled in `LC_ALL=C sort` order afterwards.
- **A child that vanishes.** If one is killed (OOM, SIGKILL) before recording anything, counting only failure markers reports a clean pass over a sweep that never finished. Every header found must now produce exactly one result. `xargs`' own exit status is a second layer, kept meaningful by having children record their result in a file and exit 0, so a non-zero `xargs` status means infrastructure failure rather than a compile result.
- **A child shell without `pipefail`.** `SHELLOPTS` is not exported, so the child does **not** inherit the parent's `set -o pipefail`. A failing `printf` would hand the compiler an empty translation unit, which exits 0 — recording a header as self-contained without ever compiling it. The child now sets `pipefail` itself. The reviewer confirmed this by stripping the line and reproducing the false pass.

## Two things caught in review that I had wrong

**Marker filenames were not injective**, though the code comment claimed they were. Mapping every separator to `_` collides `a/b.h` with `a_b.h`; my first fix (doubling `_` first) still collides `a_/b.h` with `a/_b.h` — both become `a___b.h`. Now percent-encoded, escaping `%` before mapping `/` onto it; that ordering is the load-bearing part. Verified over 7381 adversarial strings, zero collisions. A collision was fail-closed, but reported as "check did not complete" rather than the truth.

**The success line overstated the check.** It claimed all `${NHEADERS}` installed headers compile standalone. Only the 93 under the include root are ever fed to the compiler; the 94th is the top-level `shared_library/CoolPropLib.h`. The message now reports both, and a new assertion fails if that difference is ever anything but 1 — so a future install rule shipping a second unswept header cannot silently go unchecked.

`-P` is clamped to 8 when auto-detected: `getconf` reports host cores rather than a container's cgroup quota, and a syntax-only TU peaks near 180 MB, so unbounded parallelism invites the OOM killer — which takes `cc1plus`, not the bash child, making the gate report "header is not self-contained", a false regression rather than an infra error. `COOLPROP_HEADER_CHECK_JOBS` overrides without the clamp.

## Verification

Fail injection against the real install tree, because a faster gate that cannot fail is worth nothing:

| case | result |
|---|---|
| one bad header among 93 | exit 1, exactly 1 reported |
| every header bad | exit 1, 93 reported |
| child SIGKILLed mid-run | exit 1, no false pass |
| invalid job count | exit 1 |
| missing build dir | exit 2 |

`./dev/ci/preflight.sh`: **4 passed / 0 failed / 6 skipped.** Two adversarial review passes per CLAUDE.md; the second returned no blocking findings and independently reproduced the timing, the injectivity result and the `pipefail` false pass.

**Human verification pending: Ian has not independently reproduced these results.**

## Expect the Installed-header hygiene gate to be RED on this PR

It is red **on master already** — `bd CoolProp-ac2t`, raised to P1. Identical 10 failures at `c8521e810` (job 101493170860) and `85fc178d5` (job 101510523959), before this branch existed. Seven distinct headers fail on Linux/libstdc++ while passing on macOS/libc++, because libc++ pulls `<cmath>` in transitively and libstdc++ does not. That issue had *predicted* this exact mechanism and named `IdealCurves.h` and `sbtl/SurfaceSpec.h` among the latent cases; two of the failures are those. Not caused by, and not fixed by, this PR.

Chasing it exposed something this PR arguably should address: the script writes compiler output to `/tmp/installed-headers-selfcontained.log`, which CI neither prints nor uploads, so a run says *which* headers failed but never *why*. Establishing that the failure was pre-existing needed a diff of two master runs where one line of stderr would have sufficed. Left out to keep this PR to one concern; noted on `CoolProp-ac2t`.

## Filed, not fixed here

- `bd CoolProp-1n5g` (P2) — the sweep skips `shared_library/CoolPropLib.h`, the 922-line C/DLL API header downstream consumers actually include. Pre-existing; this change only made it visible by reporting the true count. It compiles standalone today.
- `bd CoolProp-ac2t` (P1) — the stdlib-specific gate above.
- `bd CoolProp-8fta` (P1) — the rest of preflight's floor: four gates that run regardless of the diff, and whole-file clang-tidy at 30 s each.

Independent of #3369 (ccache); either can merge first.

## Summary by CodeRabbit

- **Tests**
  - Improved installed-header compilation checks for more reliable validation.
  - Added parallel processing with controlled concurrency to speed up verification.
  - Improved diagnostic reporting with consistent results and clearer failure details.
  - Added safeguards to detect incomplete checks and ensure all installed headers are accounted for.
  - Enhanced reporting to distinguish validated headers from the total installed-header count.
  - Added validation for the installed include location and expected public header layout.

### DIFF
--- a/dev/ci/check-installed-headers.sh
+++ b/dev/ci/check-installed-headers.sh
@@ -111,21 +111,161 @@ if [ -z "$INC_ROOT" ] || [ ! -d "$INC_ROOT" ]; then
 fi
 SC_LOG=/tmp/installed-headers-selfcontained.log
 : >"$SC_LOG"
-SC_FAIL=0
-while IFS= read -r hdr; do
-    rel="${hdr#"$INC_ROOT"/}"
-    if ! printf '#include "%s"\n' "$rel" | "$CXX_BIN" -std=c++17 -fsyntax-only \
-            -DNO_FMTLIB -DCOOLPROP_NO_DEPRECATED_HEADER_WARNINGS \
-            -I"$INC_ROOT" -I"$EIGEN_DIR" -x c++ - >>"$SC_LOG" 2>&1; then
-        echo "FAIL: installed header is not self-contained: $rel" >&2
-        SC_FAIL=$((SC_FAIL + 1))
-    fi
-done < <(find "$INC_ROOT" -name '*.h')
+
+# The sweep is ~93 INDEPENDENT -fsyntax-only invocations, so it parallelises
+# cleanly across cores.  Serially it was 47.7 s -- the single largest item in
+# ./dev/ci/preflight.sh's 122 s floor, and ~1000x the 0.05 s that `cmake
+# --install` itself costs (bd CoolProp-5vun).
+SC_JOBS="${COOLPROP_HEADER_CHECK_JOBS:-$(getconf _NPROCESSORS_ONLN 2>/dev/null || echo 4)}"
+# Clamp the auto-detected value.  getconf reports HOST cores, not a container's
+# cgroup quota, and a syntax-only TU peaks around 180 MB here -- so on a large
+# machine an unbounded -P invites the OOM killer, which takes cc1plus rather than
+# the bash child.  The child would then see a non-zero compiler status and report
+# "installed header is not self-contained", i.e. a false regression rather than
+# an infrastructure error.  8 keeps nearly all the speedup (the sweep is only ~93
+# items); an explicit COOLPROP_HEADER_CHECK_JOBS is honoured as given -- including
+# 0, which xargs reads as "unbounded" and which therefore re-opens the very OOM
+# case the clamp exists to prevent.  That is the point of an override, but set it
+# deliberately.
+if [ -z "${COOLPROP_HEADER_CHECK_JOBS:-}" ] && [ "$SC_JOBS" -gt 8 ] 2>/dev/null; then
+    SC_JOBS=8
+fi
+SC_TMP="$(mktemp -d)"
+# Supersede the earlier trap so the scratch dir is cleaned up too.  Both paths
+# are quoted; PREFIX and SC_TMP are mktemp output, but the quoting is what keeps
+# a future edit from turning this into an unbounded rm.
+trap 'rm -rf "$PREFIX" "$SC_TMP"' EXIT
+
+SC_TOTAL="$(find "$INC_ROOT" -name '*.h' | grep -c . || true)"
+[ -n "$SC_TOTAL" ] || SC_TOTAL=0
+if [ "$SC_TOTAL" -eq 0 ]; then
+    echo "FAIL: no *.h found under the installed include root ($INC_ROOT) -- the self-containedness sweep would verify nothing." >&2
+    exit 1
+fi
+
+export CXX_BIN INC_ROOT EIGEN_DIR SC_TMP
+
+# One header per child.  Each child writes to its OWN files: two parallel
+# writers appending to a single log interleave mid-line, which would corrupt the
+# very diagnostics this gate prints.  A child records `ok.` or `fail.` either
+# way and exits 0, so xargs' own status stays reserved for infrastructure failure
+# (couldn't spawn, bad interpreter) rather than a compile result.  The one
+# deviation is safe: if the marker path itself cannot be written (ENAMETOOLONG on
+# a pathologically deep header) the child exits non-zero and surfaces through
+# xargs as exactly that kind of infrastructure failure, and the count guard below
+# catches it independently.
+#
+# -print0/-0 keeps a path containing spaces as one argument.  The command is
+# passed inline rather than via `export -f`: exported bash functions do not
+# survive a version mismatch between the parent shell and the `bash` on PATH.
+SC_XARGS_RC=0
+find "$INC_ROOT" -name '*.h' -print0 \
+    | xargs -0 -P "$SC_JOBS" -n1 bash -c '
+        # SHELLOPTS is not exported, so this child does NOT inherit the parent'"'"'s
+        # pipefail.  Without it a failing printf would hand the compiler an empty
+        # translation unit, which exits 0 -- recording a header as self-contained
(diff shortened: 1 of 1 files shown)


ibell:
Thanks — one of the two findings is valid and is now fixed; the other is a false positive, with evidence below.

## 🟠 Major, line 227 — valid, fixed in `cbcbd5a0f`

Correct, and it's the same class of overstatement this PR set out to fix, just in the other direction. The assertion checked the *count* of unswept headers while the success message asserts an *identity*. If `shared_library/CoolPropLib.h` stopped being installed and a different header took its place outside the sweep, the difference would still be 1, the gate would still go green, and the message would name a file that is no longer the exemption.

It now enumerates the unswept set and compares paths:

```
SC_UNSWEPT_EXPECTED="shared_library/CoolPropLib.h"
find "$PREFIX" -name '*.h' ! -path "$INC_ROOT/*" | LC_ALL=C sort >"$SC_TMP/unswept"
```

Verified it fires by pointing the expectation at a header that isn't installed:

```
FAIL: the set of installed *.h outside the include root is not exactly 'shared_library/SomeOtherHeader.h'.
      found:
        shared_library/CoolPropLib.h
```

(`-printf '%P'` would be tidier but is GNU-only, so the prefix is stripped in-shell to keep BSD/macOS `find` working.)

## 🔴 Critical, lines 197 / 234 — false positive

> Line 197 starts the aggregation `while` loop twice. […] The outer loop remains unterminated, so the shell cannot parse the script.

There is one pipeline per loop, and the file contains exactly two `while IFS= read -r` and two matching `done`. Line 197 reads:

```
find "$SC_TMP" -name 'out.*' | LC_ALL=C sort | while IFS= read -r out; do
```

That is `find` → `sort` → `while`, i.e. a single three-stage pipeline, not two loop starters. I think the `LC_ALL=C sort` stage was read as a second one.

"The shell cannot parse the script" is disprovable four ways, all against the reviewed commit `c5c8874fe`:

- `bash -n` — clean
- `shellcheck -S error` — clean
- the **Installed-header hygiene gate** CI job ran on exactly `c5c8874fe` ([job 101528788304](https://github.com/CoolProp/CoolProp/actions/runs/34048884725/job/101528788304)) and concluded **success**, printing `OK: … all 93 headers under the include root compile standalone`. A script that cannot be parsed cannot produce that line.
- the same script runs green in `./dev/ci/preflight.sh` locally

No change made for this one. Flagging it because the **Merge Risk: High** banner rests on it, and the remaining half of that banner is now addressed.

ibell:
> **Use the installed path for the exemption.** `SC_UNSWEPT_REL` is relative to `$PREFIX`. `CMakeLists.txt:937-946` installs the flat `CoolPropLib.h` shim under `usr/include`, so this value is `usr/include/CoolPropLib.h`, not `shared_library/CoolPropLib.h`. The new check therefore fails on the normal shared install before it can report a successful sweep.

The specific claim doesn't hold, but it pointed at something real, so I've changed the check anyway.

**Why the claim doesn't hold.** Running the actual install and computing the actual set:

```
$ cmake --install build_shared --prefix "$P"
$ find "$P" -name '*.h' ! -path "$INC_ROOT/*" | sed "s|^$P/||"
shared_library/CoolPropLib.h
```

There are three `CoolPropLib.h` under the prefix, but the other two are *inside* `INC_ROOT` and so are swept, not exempt:

```
shared_library/CoolPropLib.h                  <- outside INC_ROOT (the exemption)
shared_library/include/CoolPropLib.h          <- swept
shared_library/include/CoolProp/CoolPropLib.h <- swept
```

No `usr/include` path appears, and `CMakeLists.txt:930-950` is fPIC and calling-convention logic rather than the install rule cited. The **Installed-header hygiene gate** also passed on the reviewed commit `e8fce5bd7`, which it could not do if this check failed "before it can report a successful sweep".

**What was genuinely wrong.** Pinning the full relative path over-fits the assertion to one install layout: a change to the prefix or directory structure would fail the gate for a reason unrelated to header coverage — which is roughly the failure mode described, just reached by a different route. What the gate actually needs to assert is *which header* is exempt, not *where it sits*.

So it now matches on the file name and requires exactly one entry, printing the full installed path in both the success and failure messages so a layout change stays visible:

```
SC_UNSWEPT_EXPECTED="CoolPropLib.h"
...
if [ "$SC_UNSWEPT_N" -ne 1 ] || [ "${SC_UNSWEPT_REL##*/}" != "$SC_UNSWEPT_EXPECTED" ]; then
```

Success line now reads `… (of 94 installed *.h; shared_library/CoolPropLib.h is not swept)`, with the path derived rather than hard-coded.

Verified it still fires, by expecting a header that isn't there:

```
FAIL: expected exactly one installed *.h outside the include root, named 'SomethingElse.h'; found 1:
        shared_library/CoolPropLib.h
```

Fail-closed paths re-checked after the change: one bad header among 93 → exit 1 with exactly 1 reported; every header bad → exit 1 with 93 reported. `bash -n` and `shellcheck -S error` clean; `./dev/ci/preflight.sh` 4 passed / 0 failed.

Pushed as `17dffbc3d`.

PR #3373, opened 2026-09-08
perf(HEOS): reuse component backends for mixture transport

Tobias-Reiter:
## Description of the Change

Reuse backend-owned pure-component states when evaluating HEOS mixture viscosity and thermal conductivity, instead of constructing them for every property evaluation. The mixing rules, summation order, zero-fraction handling, and pure-fluid phase determination are unchanged.

Component states are updated on every evaluation. They are discarded before component/EOS/reference-state changes and after evaluation failures. Calling the mutable `get_components()` accessor permanently retains fresh-object evaluation for that instance, since callers can keep and later modify a reference. Read-only construction and metadata access no longer trigger this fallback.

Review/CI fixes included:
- Reject unset or mismatched mole-fraction counts before indexing, also clearing already-warmed helpers on failure.
- Resize every row of the excess-term matrix when changing the component count. The new replacement test exposed an existing heap-buffer-overflow here under CI AddressSanitizer.
- Discard obsolete linked helpers when replacing components; otherwise later binary-interaction updates can reach matrices belonging to the previous mixture. GERG rebuilds its saturation states from the owned component copy, including when the input came from a discarded helper.
- Build unsigned GUI bundles only for PRs, which cannot access fork signing secrets. Push/release signing remains enabled. This uses Tauri 2.11.0's [documented signing bypass](https://github.com/tauri-apps/tauri/blob/tauri-cli-v2.11.0/crates/tauri-cli/src/bundle.rs#L240); no tests are disabled.

## Benefits

An opt-in Catch2 benchmark compares the previous fresh-component loops with the production path. Both include a state update on every iteration; these are not repeated property-cache hits.

Windows x64, MSVC Release, 50 samples per run. Mole fractions: N2/O2/Ar/CO2/H2O = 0.70/0.08/0.01/0.10/0.11, T = 700 K. Density-input runs use 20 mol/m3; pressure-input runs use 101325 Pa. No phase is imposed.

| Update + viscosity + conductivity | Fresh components | Reused components | Time reduction |
|---|---:|---:|---:|
| rho,T, run 1 | 6.885 ms | 5.315 ms | 22.8% |
| rho,T, run 2 | 7.618 ms | 5.747 ms | 24.6% |
| p,T, run 1 | 9.111 ms | 7.678 ms | 15.7% |
| p,T, run 2 | 9.762 ms | 7.494 ms | 23.2% |

These are local microbenchmark results from the initial implementation, before the review fixes, not an end-to-end application speedup. Timing varies between runs.

## Possible Drawbacks

Each used mixture instance retains a pure-fluid backend, including its saturation helpers, per component. Read-only callers using the mutable component accessor lose the optimization, including some remaining internal saturation/phase-envelope paths. The existing approximate mixture transport models are not made more accurate by this change.

## Verification Process

- Release build completed successfully.
- Nine new regression cases: 344 assertions passed. Coverage includes temperature/density/composition changes, ECS fluids, low temperatures, zero fractions, actual object reuse, component/EOS/reference-state replacement, retained mutable aliases, failure recovery, component-count mismatches, matrix resizing, and linked-helper lifetimes.
- Transport suite including the new cases: 1690 assertions passed in 13 cases.
- Four targeted GERG regression cases: 2019 assertions passed, including component vectors borrowed from a saturation helper without retaining ownership.
- Comparisons against fresh component backends use a relative tolerance of 1e-12. They verify implementation equivalence, not experimental accuracy of the mixing rules.
- clang-format 18.1.8, `git diff --check`, and actionlint 1.7.12 passed (actionlint shellcheck/pyflakes integrations disabled).
- A focused GCC 15.2 AddressSanitizer probe against the actual `ExcessTerm` implementation passed for grow/shrink/empty matrices. This is not a full sanitized Catch2 run; the GitHub ASan job must rerun on the updated commit.
- A separate Codex adversarial review found no remaining blockers after the GERG borrowed-reference fix.
- The broader `[Helmholtz],[REFPROP],[flash],[mixture]` sweep was stopped after exceeding five minutes; it is not reported as passing. The full suite and Win32 build remain unverified locally.
- Preflight was run with `--skip=build,tests,cppcheck,clang-tidy,semgrep`: build/tests were run separately with MSVC; the three static-analysis tools were not run locally. Shared-library symbol and installed-header gates were also skipped with the preflight build.

Reproduce using the configured Catch2 build:

```text
CatchTestRunner "[transport]~[!benchmark]" --reporter compact
CatchTestRunner "[mixture-transport][!benchmark]" --benchmark-samples 50 --benchmark-resamples 1000
```

AI assistance: OpenAI Codex prepared the implementation, tests, benchmark, review fixes, and this description, and executed the checks reported above. These are agent-executed checks, not a claim of independent human validation.

## Applicable Issues

None linked.

## Summary by CodeRabbit

* **Performance**
  * Improved mixture viscosity and thermal-conductivity calculations through reusable component transport states.

* **Reliability**
  * Improved consistency after composition, EOS, reference-state, and component updates.
  * Improved recovery after transport-calculation failures.

* **Builds**
  * Pull-request GUI builds no longer require signing; push and release builds continue signing.

* **Tests**
  * Added comprehensive mixture-transport coverage and expanded GERG mixture state validation.

### DIFF
--- a/.github/workflows/gui_builder.yml
+++ b/.github/workflows/gui_builder.yml
@@ -193,16 +193,16 @@ jobs:
         uses: tauri-apps/tauri-action@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          # Tauri auto-updater signing — when present, tauri-action signs the
-          # update artifacts and emits latest.json next to each platform
-          # bundle. Without these the build still succeeds; the updater
-          # plugin just won't accept the signatures, so updates won't work
-          # for that release. Activate by adding both repo secrets.
+          # Sign updater artifacts for pushes and releases. PR builds skip
+          # signing below because fork PRs cannot access these secrets.
           TAURI_SIGNING_PRIVATE_KEY: ${{ secrets.TAURI_SIGNING_PRIVATE_KEY }}
           TAURI_SIGNING_PRIVATE_KEY_PASSWORD: ${{ secrets.TAURI_SIGNING_PRIVATE_KEY_PASSWORD }}
         with:
           projectPath: wrappers/GUI
           tauriScript: npm run tauri
+          # Fork PRs have no signing secrets. Keep signing enabled for pushes
+          # and releases; PR bundles are only used for build and smoke tests.
+          args: ${{ github.event_name == 'pull_request' && '--no-sign' || '' }}
 
       # ─── Linux smoke tests ────────────────────────────────────────────────
       # 1) Forward-compat: install .deb in a clean ubuntu:26.04 container,
--- a/src/Backends/GERG/GERGBackend.cpp
+++ b/src/Backends/GERG/GERGBackend.cpp
@@ -64,31 +64,14 @@ void GERGMixtureBackend::set_components(const std::vector<CoolPropFluid>& comps,
     // with GERG-typed states.  Constructing them with
     // generate_SatL_and_SatV = false is what stops the recursion.
     //
-    // The PREVIOUS SatL/SatV (if any -- set_components is only called from
-    // constructors today, so on a freshly-constructed object there is
-    // nothing to remove) are erased from linked_states before being replaced,
-    // so calling set_components twice on one object does not leave the
-    // superseded SatL/SatV in linked_states, where sync_linked_states would
-    // keep writing to them forever.  This erases ONLY the two SatL/SatV
-    // entries -- by pointer identity, not by clearing the whole vector -- so
-    // TPD_state/critical_state/transient_pure_state (HelmholtzEOSMixtureBackend.h:85,93,101)
-    // are left untouched if they happen to already be linked.  An earlier
-    // version called linked_states.clear() unconditionally: harmless today
-    // because those three are only ever created lazily, long after
-    // construction, but a future caller of set_components on a live object
-    // would silently strand them (non-null members, never reachable through
-    // linked_states again) -- trading one latent hazard for another instead
-    // of fixing it.
-    linked_states.erase(std::remove_if(linked_states.begin(), linked_states.end(),
-                                       [this](const shared_ptr<HelmholtzEOSMixtureBackend>& s) { return s == SatL || s == SatV; }),
-                        linked_states.end());
-
-    SatL.reset(new GERGMixtureBackend(m_model, comps, false));
+    // The base setter has discarded the old helpers. Use its owned component
+    // copy: comps may have referred to one of those now-destroyed helpers.
+    SatL.reset(new GERGMixtureBackend(m_model, this->components, false));
     SatL->specify_phase(iphase_liquid);
     linked_states.push_back(SatL);
     SatL->clear();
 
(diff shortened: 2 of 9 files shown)


src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
```diff
@@ -829,20 +831,37 @@ CoolPropDbl HelmholtzEOSMixtureBackend::calc_viscosity_background(CoolPropDbl et
     return initial_density + residual;
 }
 
+CoolPropDbl HelmholtzEOSMixtureBackend::calc_mixture_transport(parameters property) {
+    component_transport_states.resize(components.size());
+    CoolPropDbl summer = 0;
+    try {
+        for (std::size_t i = 0; i < mole_fractions.size(); ++i) {
```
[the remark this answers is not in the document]
Tobias-Reiter: Fixed in dfecd638. The transport path now rejects unset or mismatched counts before indexing, iterates over components.size(), and clears warmed helpers on rejection. Regression cases cover both mismatch directions after component replacement and independently warmed viscosity/conductivity paths with mutated fraction vectors. Read-only metadata/member accesses also no longer disable reuse. CI additionally exposed an existing excess-matrix resize overflow; every row is now resized, obsolete linked helpers are discarded, and GERG helper-owned input references are covered. Local Release transport tests: 1690 assertions in 13 cases; targeted GERG tests: 2019 assertions in 4 cases. A focused matrix ASan probe passed; full CI is rerunning.

.github/workflows/gui_builder.yml
ibell, line 1: not related to PR; please revert

src/Backends/GERG/GERGBackend.cpp
ibell, line 1: GERG does not have transport, should this be reverted?

src/Backends/Helmholtz/HelmholtzEOSMixtureBackend.cpp
```diff
@@ -246,7 +256,7 @@ void HelmholtzEOSMixtureBackend::recalculate_singlephase_phase() {
     }
 }
 std::string HelmholtzEOSMixtureBackend::fluid_param_string(const std::string& ParamName) {
-    CoolProp::CoolPropFluid cpfluid = get_components()[0];
+    CoolProp::CoolPropFluid cpfluid = components[0];
```
ibell, line 259: here and throughout, why do you not still call `get_components()`?

ibell (changes requested):
Overall, this is a lot of changes for only like a 10-20% increase in speed. I'm not entirely sure it is worth it, especially given that there are changes coming to transport property evaluation, especially for mixtures, that this would conflict with.

More generally, before filing a PR that substantially changes the code, I would appreciate an issue describing a proposed change that we can iterate on. This issue should include motivation and some data from a spike that justifies the PR. PR review is not without cost. 

Also, I would appreciate that the drive-by changes be broken out into separate PR. They can be stacked if you like, but mixing drive-by fixes and core changes makes it harder to review.

Issue #3388, opened 2026-09-21
Linux distro packaging: retire the dead wrappers/DEB, and what a real solution would need


jowr:
## Summary

`wrappers/DEB/` has been dead since 2014 and cannot build. It is being removed. This issue records why, preserves the deleted code, and writes down what a real Linux packaging story would actually require, so the next attempt does not start from zero.

Filed as a companion to the branch that removes `wrappers/DEB/` and the now-unreachable `COOLPROP_DEBIAN_PACKAGE` block in `CMakeLists.txt`.

---

### It was never maintained

`wrappers/DEB/` had exactly 4 commits, all in August 2014, all by the same author:

| Commit | Date | Message |
|---|---|---|
| `a784b5e` | 2014-08-22 | Started work on the packaging for debian. Still some issues with the install prefix... |
| `077764e` | 2014-08-25 | Added library targets to CMAKE |
| `826f1c2` | 2014-08-25 | Added pattern to ignore temporary files |
| `ab026b6` | 2014-08-25 | builds debian packages now. might need some more refining work... |

Nothing touched it for twelve years.

### Nothing invoked it

- `package.bsh` was referenced by no workflow and no script anywhere in the repo.
- The only `.deb` handling in `.github/workflows/` is `gui_builder.yml`, which is the Tauri GUI's own cargo-bundle output and unrelated.
- `COOLPROP_DEBIAN_PACKAGE=ON` was set in exactly one live place, `wrappers/DEB/debian.src/rules:9`. Nothing else turned it on, so the CMake block was reachable only by hand.

### Nothing consumed it

- `sources.debian.org` has no `coolprop` source package.
- Launchpad's Ubuntu primary archive returns 0 entries for `coolprop`.

### It could not have worked anyway

1. **The version parser was silently broken.** `package.bsh:10` computed `CPVER` by `cut -d " " -f 3` on the `set(COOLPROP_VERSION_MAJOR ...)` line. That assumed the 2014 formatting `set (COOLPROP_VERSION_MAJOR 5)` with a space after `set`. A `cmake-format` pass later removed the space, so field 3 stopped existing and both major and minor came out empty:

   ```
   CPVER='.'   ->   tarball coolprop_..orig.tar.gz, directory coolprop-.
   ```

   This is the clearest single piece of evidence that nobody ran it. It broke silently and stayed broken for years.

[...]

Issue #1280, opened 2016-10-22
Move all includes into subfolder


ibell:
Mirroring what is happening in the other projects I follow and work on, we should move all the public include files into a subdirectory called CoolProp within the include directory so that we will always include files like:

``` C++
#include "CoolProp/CoolProp.h"
```

The main reason for this is that if there are headers that have name clashes, we currently have no way to protect the includes and make sure we include the right file.  That's why I had the ``"CPExceptions.h"`` file rather than ``"Exceptions.h"``

This is a massively breaking change, so this will probably make a version 7 just by itself according to semantic versioning.

But I think it is good for long-term maintainability, especially when CoolProp is used in other projects like https://github.com/usnistgov/phifit

@jowr, @mikekaganski, @JonWel, @babaksamareh  thoughts?

babaksamareh:
That's a good idea I guess, I have run into this problem before, lots of headache during compilation.

ibell:
And we're done. This alone would make for a major release bump

PR #3389, opened 2026-09-21
Retire dead Debian packaging, HTTPS for CoolProp links, refresh developer and supporter lists

jowr:
### Description of the Change

Three independent pieces of housekeeping. No functional code changes, no behaviour change to the library.

**1. Retire `wrappers/DEB/` and its CMake option**

`wrappers/DEB/` had 4 commits, all in August 2014, and nothing touched it since. Nothing invoked it: `package.bsh` was referenced by no workflow and no script, and `COOLPROP_DEBIAN_PACKAGE` was set in exactly one live place, `wrappers/DEB/debian.src/rules`. CoolProp is in neither the Debian archive nor Ubuntu's, so the packages it would have produced were never consumed.

It could not have worked anyway. `package.bsh` parsed the version with `cut -d " " -f 3` on the `set(COOLPROP_VERSION_MAJOR ...)` line, which assumed the 2014 spelling `set (...)` with a space after `set`. A `cmake-format` pass removed that space years ago, so the parse has been yielding an empty major and minor ever since and the script built `coolprop_..orig.tar.gz`. Beyond that: the changelog was frozen at `5.0-1` while the tree is 8.0.1 (`dpkg-source` rejects the mismatch), `control` depended on a package named `dl` that does not exist, `compat` 9 and `Standards-Version` 3.9.2 are both obsolete, and `rules` had malformed makefile syntax.

The `if(COOLPROP_DEBIAN_PACKAGE)` block in `CMakeLists.txt` goes with it, since `rules` was its only consumer. It carried the GH #1280 header tree, but it also hardcoded `/usr/lib` (breaking Debian multiarch and Fedora `lib64`) and used `CMAKE_INSTALL_PREFIX` as if it were a `DESTDIR`, so a real packaging effort would rewrite it against `GNUInstallDirs` rather than build on it. The block is preserved verbatim in #3388.

**2. HTTP to HTTPS for CoolProp's own links**

Every remaining plain-http link to the CoolProp homepage, the documentation and the SourceForge download pages, across 26 files (docs, wrappers, CI yaml, `pyproject.toml`, a Maple worksheet, the EES help page, the LibreOffice extension descriptor). All three hosts serve https today, so these were an unencrypted redirect hop.

Two http links are deliberately left alone: the defunct buildbot in `dev/scripts/coolprop_builder_timer.py` (not the homepage or docs, and never had https on that port) and a quoted snippet in the historical plan under `docs/superpowers/plans/`.

**3. `Web/index.rst` developer and supporter entries**

- Jorrit Wronski's affiliation updated from IPU to Clean Heat Partners, with a working address.
- IPU logo block removed from Supporters, and `Web/_static/logo_ipu.png` deleted.
- Maplesoft logo block removed, and `Web/_static/logo_maplesoft.png` deleted. The Maple wrapper and its documentation are untouched; only the supporter logo goes.
- DTU link retargeted to `https://construct.dtu.dk/Sections/thermal-energy`. The old `https://www.mek.dtu.dk/english/Sections/TES` was not merely stale, it fails TLS outright now (the certificate no longer matches the hostname), so that logo was unclickable.
- labothap link retargeted to `https://www.labothap.uliege.be/`. The University of Liege retired the `ulg.ac.be` domain, so the old link did not resolve.

The IPU mentions in `Web/fluid_properties/Incompressibles.rst` and the `Skovrup2013` entry in `CoolPropBibTeXLibrary.bib` are **deliberately kept**: those credit the origin of third-party SecCool data that CoolProp ships, not an affiliation of the project. Removing them would falsify the data provenance.

### Benefits

1. Removes a build path that has been silently broken for over a decade and that no CI job exercises, so it cannot rot further or mislead a future packager.
2. Every CoolProp-owned link in the tree now uses https, removing a redirect hop and the mixed-content risk when these pages are embedded.
3. Three supporter links were dead or failing TLS. Readers now get working links.
4. Developer contact information is current and reachable.

### Possible Drawbacks

Anyone passing `-DCOOLPROP_DEBIAN_PACKAGE=ON` by hand loses that option. In practice it produced a package that could not be built or installed, and nothing in the repo or CI set it, so the realistic impact is zero. #3388 records what a real multi-distro packaging story would need, including why the offline-build and install-layout work in `CMakeLists.txt` has to come first.

### Verification Process

No new code, so no new tests. What was checked:

- **Every replacement URL was fetched and returns HTTP 200** before the swap: `www.coolprop.org`, `coolprop.sourceforge.net`, `sourceforge.net/projects/coolprop`, `construct.dtu.dk/Sections/thermal-energy`, `www.labothap.uliege.be`. The two URLs being replaced were confirmed broken (mek.dtu.dk fails certificate validation, labothap.ulg.ac.be does not resolve).
- **`cmake` configure verified in two modes** after the `CMakeLists.txt` deletion: default, and `-DCOOLPROP_SHARED_LIBRARY=ON`. Both reached "Configuring done / Generating done" with exit 0. Also confirmed control-flow balance across the file (158 `if`/`endif`, 7 `foreach`/`endforeach`, all matched).
- **Confirmed `COOLPROP_DEBIAN_PACKAGE` has no remaining consumer**, and that `app_name`, `APP_SOURCES`, `coolprop_hide_json_symbols` and `generate_headers` are unaffected (all are used by other option blocks that the diff does not touch).
- **Confirmed the two deleted PNGs are referenced nowhere** else in the tree after removal.
- **Confirmed no other reference to the DEB wrapper exists**: the docs wrapper index never listed it, and no workflow builds it.
- `./dev/ci/preflight.sh`: clang-format and semgrep pass.

**What was not verified locally:** no compile and no test run. The development sandbox cannot configure a fresh build tree, because CPM fetches dependencies from github.com at configure time and the sandbox proxy refuses non-CoolProp repositories. The platform `build` jobs in CI are therefore the first real compile of the `CMakeLists.txt` change. Given the diff only deletes an option block that nothing could reach, the risk is low, but CI is doing the verification the sandbox could not.

### Applicable Issues

Related to #3388, which records why `wrappers/DEB/` was retired, preserves the removed CMake block verbatim, and writes up what maintaining Linux packages for the main distros would actually require. That issue is intentionally left open as the design record; this PR does not close it.

https://claude.ai/code/session_01GA1P81E3rDcXF4LLNGVtAf

## Summary by CodeRabbit

* **Documentation**
  * Updated CoolProp website, documentation, download, and support links to use HTTPS.
  * Corrected documentation destinations, refreshed supporter information, and updated release materials with secure project URLs.
  * Documented the removal of the legacy Debian packaging workflow and the alternative shared-library build option.

* **Removed**
  * Removed the legacy Debian packaging configuration and build scripts.
  * The former Debian packaging build option no longer produces a shared library.


jowr:
Thanks, the CONTRIBUTING.md finding was correct and pointed at a real gap. Fixed in 0b4b07a, along with two related things it surfaced.

**1. `.github/CONTRIBUTING.md:44`** (the reported finding). The first pass matched `http://www.coolprop.org` case-sensitively, so the capitalised `http://www.CoolProp.org` was never touched. Now https.

While fixing it: both links on that line pointed at the *same* URL, so "development" went to the release docs. Line 29 of the same file shows the intended pair, so the development link now goes to `/dev`.

**2. A regression the first pass introduced**, not reported but found while re-sweeping. `Web/coolprop/wrappers/Javascript/index.html:15` had been rewritten to `https://www.coolprop.sourceforge.net/...`. SourceForge's wildcard certificate covers `*.sourceforge.net`, which matches **one** label, so that host fails the TLS handshake:

```
$ curl https://www.coolprop.sourceforge.net/jscript/coolprop.js
curl: (35) OpenSSL/3.0.13: error:0A000410:SSL routines::sslv3 alert handshake failure

$ curl -o /dev/null -w '%{http_code}' https://coolprop.sourceforge.net/
200
```

The http URL merely redirected; the https one cannot connect. Dropping the `www.` fixes it. The reference sits inside an HTML comment and nothing loads it, but a broken URL is worse than a dead one.

**3. Changelog entry** for the removed `COOLPROP_DEBIAN_PACKAGE` variable, under 8.0.1 Breaking Changes. CMake does not warn about unknown `-D` variables, so a downstream build still passing `-DCOOLPROP_DEBIAN_PACKAGE=ON` now configures cleanly and silently produces no shared library. That silent failure mode deserved a line.

Also re-ran the sweep case-insensitively across the tree. The only remaining plain-http CoolProp-adjacent links are the three already called out as deliberate in the PR description: the defunct buildbot on port 8010 in `dev/scripts/coolprop_builder_timer.py`, and a quoted snippet in a historical plan doc under `docs/superpowers/plans/`.

One left deliberately untouched and worth a maintainer's call: `README.md:9` still points at `http://pypi.python.org/pypi/CoolProp/` twice. That is outside this PR's stated scope (CoolProp's own homepage and documentation), and the host is a long-retired redirect to `pypi.org`, so it wants replacing rather than just re-scheming. Happy to do it here or in a follow-up.

---

